PR #62, opened 2015-01-04
Add 4K and 2.5K profiles.

kdeldycke:
A naive proposal adding default 4K and 2.5K profiles.


ddennedy:
Thanks, but remove the atsc_ prefix and use qhd_ and uhd_ instead.

kdeldycke:
@ddennedy : done ! :)

BTW, should we add the 120, 119.88 and 100 framerates for UHD profiles (as [per Rec. 2020](https://en.wikipedia.org/wiki/Rec._2020#Frame_rate)) ?

Can we push the vice further and add 8K UHD profiles ? :)

pez4brian:
I think this might be premature. For 4k and 8k, SMPT2036-1 and Rec.2020 both recommend that the colorspace is 2020, not 709. And the recommended bitdepth is 10bit or 12bit. MLT doesn't support either of those.

If we want to define a profile that uses HD transfer characteristics (8bit + 709) at 4k and 8k resolutions, then minimally, we shouldn't waste the proper names "UHD", "4K", "8K", "UHDTV1", "UHDTV2", etc.

I'm not sure about the standard transfer characteristics for QHD since I don't know what specification defines that. It kind of looks like a made up image format - so maybe BT.709 is OK for that?

ddennedy:
Those are good points, Brian, but is it true that 4K UHD and like require 2020 colorspace and > 8-bit? I think all 4K samples I have are 8-bit Rec. 709. They come from YouTube downloads, Elemental, and a Samsung Note 3 phone. I see a reference on this wikipedia page that the DigitalEurope standard for the UHD logo is 8-bit Rec. 709:
https://en.wikipedia.org/wiki/Ultra-high-definition_television#cite_ref-DigitalEuropeUHDlogoHDTVtest_136-0

Note that we are not using referring to it as ITU UHDTV or ATSC 3.0. In my experience, UHD and QHD are resolution designations.
https://en.wikipedia.org/wiki/Graphics_display_resolution#UHD_.284K.29_.283840.C3.972160.29

pez4brian:
I acknowledge that almost all 4k content in the wild is 709 8-bit. This is because everyone is using their existing tech and just bumping up the resolution.

Both 2020 and SMPTE2036 allow the use of 709. Here is a note from 2020:
"Conventional non-constant luminance Y'C' B C' R may be used when use of the same operational practices as those in SDTV and HDTV environments is of primary importance through a broadcasting chain (see Report ITU-R BT.2246)."

In the article you referenced, DigitalEurope allowed 8bit 709 as the bare minimum requirements to display the UHD logo on a TV set. But they are not saying that it "is the standard".

So 4k + 709 + 8bit isn't "illegal" or non-compliant. It is just "not the recommended way" to do it. My warning would be that if we use up the "good" profiles names for the bare minimum method, then we might regret it when the world finally catches up to using the recommended method - and we will wish that we could use the proper name for the proper profile.

On the other hand, what would be the consequence if we defined the profile as 709 8bit now, but in the future ended up changing it to 2020 10bit? Would everything break? Or would nobody notice?

ddennedy:
I understand your concern, but the mlt_profile does not specify a bit depth, and I do not anticipate adding that. We can change the uhd_ profiles to use 2020 when that becomes available, or if we anticipate that would cause some regression (e.g., Kdenlive builds against an older version of libswscale that does not yet do 2020), then we can add uhdtv_ or atsc3_ variants of these profiles. Or, for the sake of reducing profile clutter, we could add a bit of code that says if 2020 was requested but not available, then use 709.

pez4brian:
Ok. I'm cool with that. In that case, let's just create the comprehensive set of profiles for all 2.5k, 4k and 8k resolutions and frame rates.

QHD seems the right name for 2.5k. But what to call 4k and 8k?

For 4k, one could go with "4K UHDTV" which is a little ambiguous. Or, you could go with "UHDTV1" which is a term that was very precisely defined by SMPTE in SMPTE2036-1.

The same goes or 8k: "8K UHDTV" (colloquial) or "UHDTV2" (SMPTE).

kdeldycke:
Thanks @ddennedy for the detailed rationale (and for the merge).

PR #63, opened 2015-01-31
Fix image sequence not playing (issue #343326)

j-b-m:
Trying to play an image sequence pauses / freezes with SDL/xgl consumers. Reproducible with the images provided in this kdenlive bug:
https://bugs.kde.org/show_bug.cgi?id=343326 (see attachment "Image files used to create the slideshow").

The issue did not appear on the avformat consumer, so I guess it is related to frame dropping consumers. This patch fixes the issue - looks like the frame position must be set after loading the image.

### DIFF
--- a/src/modules/gtk2/producer_pixbuf.c
+++ b/src/modules/gtk2/producer_pixbuf.c
@@ -632,14 +632,14 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
 		// Set the producer on the frame properties
 		mlt_properties_set_data( properties, "producer_pixbuf", self, 0, NULL, NULL );
 
-		// Update timecode on the frame we're creating
-		mlt_frame_set_position( *frame, mlt_producer_position( producer ) );
-
 		// Refresh the pixbuf
 		self->pixbuf_cache = mlt_service_cache_get( MLT_PRODUCER_SERVICE( producer ), "pixbuf.pixbuf" );
 		self->pixbuf = mlt_cache_item_data( self->pixbuf_cache, NULL );
 		refresh_pixbuf( self, *frame );
 		mlt_cache_item_close( self->pixbuf_cache );
+
+                // Update timecode on the frame we're creating
+		mlt_frame_set_position( *frame, mlt_producer_position( producer ) );
 
 		// Set producer-specific frame properties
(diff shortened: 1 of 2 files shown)


ddennedy:
Running "melt -verbose test/.all.png" repeatedly reports "too many frames dropped - forcing next frame." However, a little debugging shows it is actually not forcing the next frame (not dropping it). The combination of slowness of libpng and design of the image producers pushed the mlt_consumer frame-dropping algorithm into a corner condition bug. As a result, frames were continually dropped. As a reminder, the frame-dropping algorithm has a "back off" design to at least occasionally show a frame to prevent things from appearing to be "stuck."

The following patch fixes that problem:
```C
--- src/framework/mlt_consumer.c
+++ src/framework/mlt_consumer.c
@@ -940,7 +940,7 @@ static void *consumer_read_ahead_thread( void *arg )
 		skip_next = 0;
 
 		// Only consider skipping if the buffer level is low (or really small)
-		if ( mlt_deque_count( priv->queue ) <= buffer / 5 + 1 )
+		if ( mlt_deque_count( priv->queue ) <= buffer / 5 + 1 && count > 1 )
 		{
 			// Skip next frame if average cost exceeds frame duration.
 			if ( time_process / count > frame_duration )
```

Next, what happens when your patch is applied? First of all, I notice that far fewer frames are dropped, which is a little bit of a pleasant surprise. Maybe there is something good in this patch. But what does your patch actually do and why does it have these benefits? Analysis reveals that it is defeating the pixbuf/qimage cache during the get_frame() calls. Basically, it makes it always evaluate the current image index as 0 and always keeps that pixbuf/qimage object warm in the cache. Furthermore, we no longer take the hit to read an image file for a frame that will simply be dropped. Ah, now that seems like a great benefit!

Why are we even bothering to refresh_pixbuf()/_qimage() in the get_frame() callbacks? What happens if we simply remove that call within the get_frame() callbacks? We get the same performance benefit as this contributed patch and further code reduction. However, taking a deeper look inside refresh_pixbuf()/_qimage() reveals some interesting calls to set the width and height properties on the producer and frame. Oh dear, run another test on a folder of images that are not uniform, and there is regression that distorts the aspect ratio.

Therefore, because the image producers are designed for the flexibility of handling both image sequences (typically uniform attributes) as well as automatic slideshows (non-uniform), both this patch and a more extensive version of it cause a functional regression.

PR #70, opened 2015-04-02
Use env var ${PKG_CONFIG} instead of hard coding pkg-config

heirecka:
Sorry, I also posted this to the mailing list because I read about Contributing Code on the homepage before realising that a pull request would probably be easier.

### DIFF
--- a/configure
+++ b/configure
@@ -147,6 +147,8 @@ build_config()
 		echo export prefix=$prefix
 		echo export libdir=$libdir
 		echo export bindir=$prefix/bin
+        [ -z ${PKG_CONFIG} ] && PKG_CONFIG="pkg-config"
+        echo export PKG_CONFIG="${PKG_CONFIG}"
 	) >> mlt-config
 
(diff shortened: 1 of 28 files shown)


ddennedy:
This was reverted and rejected. It is horribly broken when PKG_CONFIG env var is undefined, which is the default on most systems.

configure
```diff
@@ -147,6 +147,8 @@ build_config()
 		echo export prefix=$prefix
 		echo export libdir=$libdir
 		echo export bindir=$prefix/bin
+        [ -z ${PKG_CONFIG} ] && PKG_CONFIG="pkg-config"
```
pez4brian: I haven't tried this patch myself. But I wonder if the "&&" should be "||" to get the effect you are looking for.

ddennedy:
I had already fixed configure, but then I realized make also depends on env var PKG_CONFIG. There might be an easy fix for that too, but we have not needed this so far for 10 years. So, if someone wants to change it now, the onus is on them to provide a correct patch.

PR #71, opened 2015-04-17
added fps display to progress output

bjoern-r:
This commit adds fps display to the progress output. 
The algorithm is borrowed from ffmpeg.

### DIFF
--- a/src/melt/melt.c
+++ b/src/melt/melt.c
@@ -377,6 +377,14 @@ static void event_handling( mlt_producer producer, mlt_consumer consumer )
 
 #endif
 
+// borrowed from ffmpeg
+static int64_t av_gettime_relative(void)
+{
+	struct timespec ts;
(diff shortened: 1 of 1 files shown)


ddennedy:
Rejected because there are apps - namely Kdenlive and Shotcut plus surely some I do not know about including scripts - that run melt as a child process and parse the progress output. This will likely break those apps and impose superficial version dependencies. Also, clock_gettime() is not available on OS X and MinGW.

If you really want to get something like this in, then you must create a new option that outputs in some structured text format like CSV, YAML, or JSON such that an app is likely to use a parser that is more resilient to later additions. Such an output should preferably not add a new library dependency. And it must be a cross-platform (Linux, BSD, OS X, Windows MinGW) solution.

bjoern-r:
you are right, it was just a quick hack. I will rethink how to do it correctly.

PR #75, opened 2015-05-06
Process audio before video when multi-threading

pez4brian:
This request restores processing audio before video when real_time is < -1 or > 1.

The change in producer_avformat was reliable with my testing and is congruent with the fact that av_read_frame() is mutex protected everywhere else.

It is important to note that this change to mlt_consumer does NOT add parallel audio processing. The audio for each frame is processed sequentially in worker_get_frame() before the frame is passed to the thread pool for parallel image processing. So it is not necessary to make sure all audio producers/filters are thread save.

I propose that we do not merge these changes until after the next Shotcut release.

### DIFF
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -1377,8 +1377,10 @@ static mlt_frame worker_get_frame( mlt_consumer self, mlt_properties properties 
 	// Frame to return
 	mlt_frame frame = NULL;
 	consumer_private *priv = self->local;
-	double fps = mlt_properties_get_double( properties, "fps" );
 	int threads = abs( priv->real_time );
+	int audio_off = mlt_properties_get_int( properties, "audio_off" );
+	int samples = 0;
+	void *audio = NULL;
 	int buffer = mlt_properties_get_int( properties, "_buffer" );
 	buffer = buffer > 0 ? buffer : mlt_properties_get_int( properties, "buffer" );
 	// This is a heuristic to determine a suitable minimum buffer size for the number of threads.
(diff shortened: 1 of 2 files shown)


pez4brian:
I propose that we do not merge these changes until after the next Shotcut release.

ddennedy:
This still crashes for me. In the debugger, it crashes inside av_seek_frame() in producer_avformat.c, line 870. Meanwhile, another thread is in the av_seek_frame() of find_first_pts(). So, I add pthread_mutex_lock/unlock on the packets_mutex around the call on line 870, and I am back to the deadlock I mentioned previously. More work required to debug that deadlock.

pez4brian:
Thanks for testing. I'll get back on it.

pez4brian:
I'm having a difficult time reproducing the crash you report at line 870. I have been dropping a bunch of short clips on the timeline and then playing and seeking around. But I can't get a crash with the attached changes. I also increased the number of threads by forcing real_time=-10

What steps do you take to reproduce it?

ddennedy:
Hey, I primarily reproduce it by simply playing different clips - not adding them to a playlist or timeline. This is with video mode = automatic.

ddennedy:
I just noticed that I have difficulty to reproduce it until I play a MTS file - any MTS file, not just one from my camera. But then, I am also not using FFmpeg 2.3 here as per Shotcut at this time, but rather 2.4.2, and that might make a difference. Obviously, it needs to be stable outside of Shotcut controlled conditions as well.

pez4brian:
MTS files seem to be the key to failure. I'm back in business.

pez4brian:
@ddennedy This latest version seems to avoid crashes and deadlocks in my testing. Can you please give it a try in your environment?

ddennedy:
It is working for me so far with limited testing. I will test some more and expect to pull this soon.

PR #79, opened 2015-07-02
Add Animation::insert

vpinon:
Hello,

I'm mentoring Kdenlive's GSoC student for integration of property animation.
He is missing a Mlt++ keyframe insert function : here is a quick one that seems to work in unit tests...

Vincent.
PS: I'm very happy to start contributing to MLT ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.moc
 *.o
 *.so
 *.so.*
@@ -29,3 +30,5 @@ mlt-config
 packages.dat
 make.inc
 src/examples/play
+src/tests/*Makefile
+src/tests/*/[^.]*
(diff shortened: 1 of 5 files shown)


ddennedy:
Lack of Animation::insert() is intentional. You should be using Properties::anim_set(). This is explained on the wiki page: http://www.mltframework.org/bin/view/MLT/PropertyAnimation
I am inclined to reject this unless you can clearly explain why this is needed.

vpinon:
Sorry, we should have been reading the wiki and examples more carefully... drop it!

ddennedy:
Since no one yet has a lot of experience with the API, there might be some reason to justify this. Just re-open it if you run into a situation where you need this or it is much, much more convenient.

PR #83, opened 2015-08-03
fill support in composite transition

yuroller:
I would like to completely fill a composite-transition region by automatically cropping the B frame while compositing into the A frame. Using filter crop is an alternative but it is cumbersome as I would need to specify the exact pixel cropping sizes.
I use it to create a left-right split view:
melt -profile atsc_720p_30 \
colour:black length=1000 out=999 \
-track b_left.mkv out=999
-track b_right.mkv out=999
-transition composite out=999 a_track=0 b_track=1 geometry=0%/0%:50%x100% fill=1
-transition composite out=999 a_track=0 b_track=2 geometry=50%/0%:50%x100% fill=1

### DIFF
--- a/src/modules/core/transition_composite.c
+++ b/src/modules/core/transition_composite.c
@@ -811,12 +811,12 @@ static int get_b_frame_image( mlt_transition self, mlt_frame b_frame, uint8_t **
 		// ????: Shouln't this be the default behaviour?
 		if ( mlt_properties_get_int( properties, "fill" ) && scaled_width > 0 && scaled_height > 0 )
 		{
-			if ( scaled_height < normalised_height && scaled_width * normalised_height / scaled_height <= normalised_width )
+			if ( scaled_height < normalised_height && scaled_width * normalised_height / scaled_height >= normalised_width )
 			{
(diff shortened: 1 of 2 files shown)


ddennedy:
Sure. When fill=0 (default), composite will only downscale the source of b_track. IOW, if b_track is smaller than the geometry rectangle, then it is not upscaled unless fill=1. But, as you noticed, regardless of the fill setting, the scaling preserves the entire image by padding instead of completely filling the rectangle by cropping. Also, we cannot change the name of the existing "fill" property even if that sounds like not the best name.

yuroller:
Do you think that "crop_to_fill=1" can do in order to trigger the new behavior?
What is the precedence order between fill=1, distort=1 and crop_to_fill=1 as they look mutually exclusive?

ddennedy:
Yes, crop_to_fill works. distort=1 is highest priority making fill and
crop_to_fill meaningless. crop_to_fill=1 can imply fill=1 where it is like
an option to the fill behavior since I think you want it to invoke both
upscaling and downscaling as-needed.

PR #85, opened 2015-09-10
Fix automask filter starting with wrong profile info for geometry

j-b-m:
When starting an automask filter, it was using default pal_dv profile instead of actual profile. For example:
melt myclip.mpg -attach autotrack_rectangle obscure=1 geometry="50%/50%:50%x50%"

When using an HD video file, the initial rect was using 360x288 position and size instead of real half size of the clp


j-b-m:
Oh, sorry I did a copy / paste from the filter_shape.c effect which predates the mlt_filter_get_length2 stuff. I updated the patch with a fix for filter_shape.c too.

PR #87, opened 2015-09-19
Allow setting audio_device and audio_driver on sdl_audio consumer

j-b-m:
Since we now mostly use sdl_audio, we should be able to set SDL's audio_device and audio_driver like on other sdl consumer I think.


j-b-m:
Is it ok like this (updated pull request)?
if ( audio_driver && strcmp( audio_driver, "" ) )

ddennedy:
Squashed in a529095

PR #94, opened 2015-12-13
Fix affine transition losing last row/column of pixels

j-b-m:
Affine transition was losing last row and column of pixels. You can reproduce with this simple test:
melt color:red -track color:blue -transition affine

You can see a thin red line on the right and bottom corners. My patch fixes the problem


ddennedy:
Hi JB,
This issue seems related: https://github.com/mltframework/shotcut/issues/118
However, this does not resolve that. I wonder if there is a more general problem with offset or starting coordinates.

PR #95, opened 2016-01-25
Improve detection for low volume in audiolevel filter

j-b-m:
Currently, the audiolevel filter returns a 0 volume for all audio below -20dB. However, the -50dB to -20dB range is still audible, and it feels strange to have no apparent volume with this filter. My patch proposal increases sensitivity to return non zero values for low volume (allowing improved audio level meter in Kdenlive)

### DIFF
--- a/src/modules/normalize/filter_audiolevel.c
+++ b/src/modules/normalize/filter_audiolevel.c
@@ -75,7 +75,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 
 		for ( s = 0; s < num_samples; s++ )
 		{
-			double sample = cabs( pcm[c + s * num_channels] / 128.0 );
+			double sample = fabs( pcm[c + s * num_channels] / 128.0 );
 			val += sample;
 			if ( sample == 128 )
 				num_oversample++;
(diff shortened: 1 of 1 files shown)


ddennedy:
I think cabs() is incorrect, and it should be fabs(). cabs() gives me compiler warnings because complex.h is not included, and the value passed is not complex.

I am not entirely opposed to it, but I am curious why today I see audio levels less than -20 in Shotcut. I definitely see more definition at the low levels in Shotcut's audio waveforms after applying this (with fabs).

j-b-m:
Yes, fabs is definitely better. Looking at the code, it adds values from all samples in a frame to get the level. Currently, the value from each sample is treated as integer, so if we have 20 samples in a frame with a value of 0.9, the result will be 0 instead of 18.

We are adding rounding errors, so sometimes we get values when a few sample are greater than 1, but the result of current code is really not reliable for low volumes.

TheDiveO:
May a chime in? I originally reported this to Kdenlive because the new audio meter integrated in the clip/project monitors of Kdenlive seemed to be stuck at around -20dB. I'm feeding "voice over" audio into MLT at around -10dB. However, in Kdenlive I'm also mixing in sound effects that have to be unintrusive, so less than -20dB. Of course, the audio meter could be made start at -20dB earliest, but in some audio tools, such as Audacity I also see lower levels than -20dB. Though I have to admit that my behringer USB mixer has a -20dB LED as its lowest level mark...

ddennedy:
This was merged by amending the first commit in my local copy.

PR #97, opened 2016-01-30
Fix crash on Mlt::Animation::is_key

j-b-m:
While merging the animation framework in Kdenlive, I noticed a crash in Mlt::Animation::is_key.
It calls mlt_animation_get_item, but sets the item->property to NULL, which currently crashes.


ddennedy:
This was merged, but I cleaned up the indent chars by amending the commit.

PR #103, opened 2016-03-19
Qt5 titler

j-b-m:
Add support for text shadow in kdenlive titler, gradients in rectangles and fixes a small bug making rectangles display a border when not requested


bmatherly:
I see that you are calling this "Qt5" titler. Are you using new Qt5 features? Do we need to make sure that this module is excluded from the build if someone is building against Qt4?

j-b-m:
Good catch, I forgot about Qt4. Shouldn't be too hard to make it compile and run on both, working on a fix...

PR #107, opened 2016-04-01
Kdenlive  titler fixes

j-b-m:
Fixes the blurring of shadow when it has a color and some clipping in painting


ddennedy:
I am going to give you commit access. I think I should have done that a while ago.

PR #110, opened 2016-04-17
Fix crash introduced in recent affine fix

j-b-m:
My last commit to transition_affine.c introduced a possible crash when using bilinear rescale, because my commit tested values with lrintf when bilinear uses floorf to do the rounding. 
So a value of -0.2, would result in -1 and be outside of array.

Each type of interpolation uses a different rounding, so I could not find an easy way to correctly check the boundaries. My solution does lets -1 pass with bicubic, which is not correct but the bicubic interps do a safety check so no crash. The other solution would be to add a boundary check in the nearest/bilinear interp functions themselves but that would mean additional steps for each pixels which is why I came with this workaround. 
Comments welcome if someone has a better logical way to solve this.

### DIFF
--- a/src/modules/plus/transition_affine.c
+++ b/src/modules/plus/transition_affine.c
@@ -550,43 +550,34 @@ static int transition_get_image( mlt_frame a_frame, uint8_t **image, mlt_image_f
 			affine_scale( affine.matrix, scale_x, scale_y );
 		}
 
-		// Set the interpolation function
-		if ( interps == NULL || strcmp( interps, "nearest" ) == 0 || strcmp( interps, "neighbor" ) == 0 )
-			interp = interpNN_b32;
-		else if ( strcmp( interps, "tiles" ) == 0 || strcmp( interps, "fast_bilinear" ) == 0 )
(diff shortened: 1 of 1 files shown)


ddennedy:
I agree that it would be nice to avoid the overhead of range tests in the inner loop. I don't know enough about the various interp routines to understand why they have different roundings. So, I think this approach is fine for now.
Do you think we should factor this into the existing if else if statements that set interp function pointer?

j-b-m:
Oh yes, I now realize that in the above if else that set interp, we have 6 if () for 3 possible values.
So I guess we could mix my boundary settings with a simplified version of the above ifs, like:

if ( interps == NULL || strcmp( interps, "nearest" ) == 0 || strcmp( interps, "neighbor" ) == 0 || strcmp( interps, "tiles" ) == 0 || strcmp( interps, "fast_bilinear" ) == 0 )
{
    interp = interpNN_b32;
    minima =...
}

If it's ok for you I can make the commit.

ddennedy:
Yes, please and then you can merge it as well.

PR #111, opened 2016-04-20
crop_to_fill property to composite transition

yuroller:
The property crop_to_fill on composite transition fills the B frame completely by cropping edges of the A frame, in order to maintain the aspect ratio.
halign and valign properties can be used to select which edges to crop.


yuroller:
I think I have found a bug in this pull request: investigating tommorow

yuroller:
Now it should work properly even with smaller images than the destination rectangle: they are enlarged to fill the rectangle

ddennedy:
I made a squashed merge of this.

PR #114, opened 2016-06-23
Rename config.h to MltConfig.h

chouquette:
Hi,

config.h is usually the name of a per-project autoconf generated header. Shipping a header named as such seems like a potential issue for any project using autotools


ddennedy:
Good idea, thanks.

PR #121, opened 2016-08-12
producer_colour: fix crash and add support for mlt_image_yuv420p

gasparikm:


src/modules/core/producer_colour.c
```diff
@@ -155,7 +170,6 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 			memset(p, 0, size);
 			break;
 		default:
```
bmatherly: Suggestion: Explicitly set this case to mlt_image_rgb24a and either remove the default case or turn it into an error case.

src/modules/core/producer_colour.c
```diff
@@ -116,6 +120,17 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 
 		switch ( *format )
 		{
+		case mlt_image_yuv420p:
+		{
+			uint8_t y, u, v;
+
+			RGB2YUV_601_SCALED( color.r, color.g, color.b, y, u, v );
+			int plane_size =  *width * *height;
```
bmatherly: Suggestion: move this declaration above "unit8_t y, u, v;" above.

bmatherly:
+1 from me.

PR #127, opened 2016-09-27
pkg-config instead of sdl-config allows cross-builds

vpinon:
Hello,

In the process of building kdenlive & all deps through MXE, I applied this patch to MLT.
pkg-config is automatically configured to point to target libs, what I didn't manage to do with sdl-config (always pointing to host libs).
MLT now cross-builds properly with MXE, and of course still builds natively.
Would you be interested in building this way in any case?

Cheers,

Vincent


ddennedy:
I think it is OK, but I want to wait until Oct 1 because I do not want to interrupt shotcut builds for next release on Oct 1.

PR #132, opened 2016-10-12
added initial .travis.yml

krichter722:
Usage of this service pervents issues like [failure of make due to `factory.c:42:53: error: unknown type name ‘plugin_desc_t’`](https://travis-ci.org/krichter722/mlt/builds/167195137) (I assume you don't need an extra report for this).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: c
+before_script:
+- sudo apt-get -qq update
+- sudo apt-get build-dep melt
+script:
+- ./configure && make -j


ddennedy:
Thank you, but do not assume we have no CI because we are not using GitHub-Travis. :-) Our [nightly builds](http://build.mltframework.org/overview.html?guest=1) are not reporting any compile error because we use a different configure option (--enable-gpl). It seems we need to change up more options on some builds to get more coverage. Thanks for reporting the problem and your contribution.

PR #133, opened 2016-10-13
Sliced processing

max-verem:
Proposed patches implement sliced threading model of frame processing.

It makes possible to split processing of some large amount of data among several threads to improve performance.

Some sliced processing implemented in NDI consumer (for swab operation) and transition composite (picture compositing)

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -481,3 +481,11 @@ MLT_0.9.10 {
     mlt_factory_repository;
     mlt_properties_to_utf8;
 } MLT_0.9.8;
+
+MLT_0.9.11 {
+  global:
+    mlt_slices_init;
+    mlt_slices_close;
+    mlt_slices_run;
+} MLT_0.9.10;
+
--- a/src/framework/mlt_slices.c
+++ b/src/framework/mlt_slices.c
@@ -1,0 +1,224 @@
+/**
+ * \file mlt_slices.c
+ * \brief sliced threading processing helper
+ * \see mlt_slices_s
+ *
+ * Copyright (C) 2007-2016 Meltytech, LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include "mlt_slices.h"
+
+#include <stdlib.h>
+#include <unistd.h>
+#include <pthread.h>
+
+#define MAX_SLICES 32
+#define ENV_SLICES "MLT_SLICES_COUNT"
+
+typedef struct mlt_slices_s
+{
+	int f_exit;
+	int count;
+	int readys;
+	pthread_mutex_t cond_mutex;
+	pthread_cond_t cond_var_job;
+	pthread_cond_t cond_var_ready;
+	pthread_t threads[MAX_SLICES];
+	struct
+	{
+		int jobs, done;
+		mlt_slices_proc proc;
+		void* cookie;
+	} runtime;
+} *mlt_slices;
+
+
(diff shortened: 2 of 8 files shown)


src/framework/mlt_slices.c
```diff
@@ -0,0 +1,224 @@
+/**
+ * \file mlt_slices.c
+ * \brief sliced threading processing helper
+ * \see mlt_slices_s
+ *
+ * Copyright (C) 2007-2016 Meltytech, LLC
```
ddennedy: Set the copyright year for new files to only the current year (2016).

src/framework/mlt_slices.c
```diff
@@ -0,0 +1,224 @@
...
+ * \return the context pointer
+ */
+
+mlt_slices mlt_slices_init( int threads, int policy, int priority )
+{
+	pthread_attr_t tattr;
+	struct sched_param param;
+	mlt_slices ctx = (mlt_slices)calloc( 1, sizeof( struct mlt_slices_s ) );
+	char *env = getenv( ENV_SLICES );
+	int i, cpus = sysconf( _SC_NPROCESSORS_ONLN ), env_val = env ? atoi(env) : 0;
```
ddennedy: I do not think sysconf() is available in mingw toolchains. We need to ifndef(_WIN32) this out and initialize it to sane values.

src/framework/mlt_slices.c
```diff
@@ -0,0 +1,224 @@
...
+
+typedef struct mlt_slices_s
+{
+	int f_exit;
+	int count;
+	int readys;
+	pthread_mutex_t cond_mutex;
+	pthread_cond_t cond_var_job;
+	pthread_cond_t cond_var_ready;
+	pthread_t threads[MAX_SLICES];
```
ddennedy: It seems like we also need a thread pool. If there are multiple services using slice processing along with frame-threaded parallel processing in the consumer, then there can be many threads active simultaneously. We can say this initial version should not be used with consumer parallel processing, but have you thought about a thread pool yet?

src/modules/core/transition_composite.c
```diff
@@ -1279,6 +1351,20 @@ static int transition_get_image( mlt_frame a_frame, uint8_t **image, mlt_image_f
...
 			{
+				mlt_slices sliced = NULL;
+
+				// init or set slices obj
+				if ( mlt_properties_get( properties, "sliced_composite" ) )
+				{
+					sliced = mlt_properties_get_data( properties, "sliced_composite_obj", NULL );
+					if ( !sliced )
+					{
+						sliced = mlt_slices_init( 0, SCHED_OTHER, sched_get_priority_max( SCHED_OTHER ) );
```
ddennedy: FYI, on Linux, `sched_get_priority_max( SCHED_OTHER )` should be 0.

src/modules/ndi/ndi.c
```diff
@@ -305,6 +337,10 @@ static int consumer_ndi_start( mlt_consumer consumer )
 
 	if ( !self->f_running )
 	{
+		if ( !self->sliced_swab && mlt_properties_get( properties, "sliced_swab" )
```
ddennedy: Do you want to document this property?

src/framework/mlt.vers
```diff
@@ -481,3 +481,11 @@ MLT_0.9.10 {
     mlt_factory_repository;
     mlt_properties_to_utf8;
 } MLT_0.9.8;
+
+MLT_0.9.11 {
```
max-verem: done

src/framework/mlt_slices.c
max-verem: Actually this patches is attempt to utilize all CPU for processing, because of current frame-threaded processing in not optimal or gives not me performance boost. Actually frame-threaded processing require:

1) asynchronous avformat_consumer's read-forward thread
2) moving pixel format conversation to filter (out of that reading thread) and possible simplified with SSE2  procedures for cases not required luma-transfer.
3) dynamic changing priority of consumer's threads to provide highest priority to most wanted frame rendering thread

And yes, not all operation could be sliced, but only frame-threading, but even in this case slicing gives more performance boost.

src/modules/core/transition_composite.c
max-verem: That is just for example... not all remember this values...

src/framework/mlt_slices.c
max-verem: i will use:

```
#ifdef _WIN32
<------>int cpus = GetCurrentProcessorNumber( );
#else
<------>int cpus = sysconf( _SC_NPROCESSORS_ONLN );
#endif

```

src/modules/core/transition_composite.c
ddennedy: I am OK with leaving this as-is. I just wanted to let you know in case you expected something different. On Another OS, this might really give a higher priority if that is OK.

ddennedy:
Closing this for the other pull request.

PR #135, opened 2016-10-16
Sliced2

max-verem:
This is updated branch of recent pull request https://github.com/mltframework/mlt/pull/133 with changes, please review.

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -480,4 +480,7 @@ MLT_0.9.10 {
   global:
     mlt_factory_repository;
     mlt_properties_to_utf8;
+    mlt_slices_init;
+    mlt_slices_close;
+    mlt_slices_run;
 } MLT_0.9.8;
--- a/src/framework/mlt_slices.c
+++ b/src/framework/mlt_slices.c
@@ -1,0 +1,229 @@
+/**
+ * \file mlt_slices.c
+ * \brief sliced threading processing helper
+ * \see mlt_slices_s
+ *
+ * Copyright (C) 2016 Meltytech, LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include "mlt_slices.h"
+
+#include <stdlib.h>
+#include <unistd.h>
+#include <pthread.h>
+
+#define MAX_SLICES 32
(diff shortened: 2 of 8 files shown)


src/framework/mlt_slices.c
```diff
@@ -0,0 +1,229 @@
...
+{
+	pthread_attr_t tattr;
+	struct sched_param param;
+	mlt_slices ctx = (mlt_slices)calloc( 1, sizeof( struct mlt_slices_s ) );
+	char *env = getenv( ENV_SLICES );
+#ifdef _WIN32
+	int cpus = GetCurrentProcessorNumber( );
+#else
+	int cpus = sysconf( _SC_NPROCESSORS_ONLN );
+#endif
```
ddennedy: This does not adequately address Darwin and BSD targets. It needs to be `if defined(WIN32) ... elif defined (__linux__) ... else ....` Maybe the else should just be 1 or some macro definition.

max-verem: FreeBSD says:
https://www.freebsd.org/cgi/man.cgi?query=sysconf&sektion=3

and Darwin says the same:
https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/sysconf.3.html

call ```sysconf( _SC_NPROCESSORS_ONLN )``` present under non-windows system...

ddennedy: OK, great, thanks for tracking those down.

src/framework/mlt.vers
```diff
@@ -480,4 +480,7 @@ MLT_0.9.10 {
   global:
     mlt_factory_repository;
     mlt_properties_to_utf8;
+    mlt_slices_init;
+    mlt_slices_close;
+    mlt_slices_run;
 } MLT_0.9.8;
```
ddennedy: I am sorry there was some miscommunication in my previous pull request comment about the changes in this file. You were on the right track, but using the old versioning scheme, and I mis-communicated about MLT_0.9.10. This needs to look like this:
```
MLT_6.4.0 {
  global:
    mlt_slices_init;
    mlt_slices_close;
    mlt_slices_run;
} MLT_0.9.10;
```
The versioning changed recently to 6.x.y, but the symbols in MLT_0.9.10 were added between 0.9.8 and 6.0.0, and I did not want to change the symbol version block on people using git master. But, any new symbols now should go into a new block and use the new versioning scheme.

Feel free to fix this and force push your branch if needed to prevent the need to make a new branch or commit on top of this. Whatever works for you.

src/framework/mlt_slices.c
max-verem: in the same time http://www.shogun-toolbox.org/doc/en/0.10.0/Parallel_8h_source.html:
```
00042 #if defined(LINUX) && defined(_SC_NPROCESSORS_ONLN)
00043         return sysconf( _SC_NPROCESSORS_ONLN );
00044 #elif defined(DARWIN)
00045         int num; /* for calling external lib */
00046         size_t size=sizeof(num);
00047         if (!sysctlbyname("hw.ncpu", &num, &size, NULL, 0))
00048             return num;
00049 #endif
```

src/framework/mlt.vers
max-verem: is it correct:
```
index 14b73ca..e52a67d 100644
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -481,3 +481,10 @@ MLT_0.9.10 {
     mlt_factory_repository;
     mlt_properties_to_utf8;
 } MLT_0.9.8;
+
+MLT_6.4.0 {
+  global:
+    mlt_slices_init;
+    mlt_slices_close;
+    mlt_slices_run;
+} MLT_0.9.10;
```

max-verem:
i will provide updated pull request for this

PR #141, opened 2016-10-23
Pango3 wrap1

max-verem:
This is a pango3 branch that add wrapping support by reworking current width_fit and width_crop functionality.


ddennedy:
I am confused about why there are 2 pull "wrap" pull requests. This one breaks the properties interface; so, I suppose I shall reject this one and merge the other.

PR #142, opened 2016-10-23
Pango3 wrap2

max-verem:
This is a pango3 branch that add wrapping support by introducing two new additional parameters and does not break current module api


ddennedy:
I cherry-picked 573ed4a

PR #143, opened 2016-10-24
Implement shared memory producer/consumer

max-verem:
This patch implement shared memory interface that allow to use multiple processes for MLT.

Usage example:

```
/usr/local/mlt/bin/melt -profile atsc_1080i_50 jamala.mxf -consumer shmem:/tmp/demo8

/usr/local/mlt/bin/melt -profile atsc_1080i_50 -producer shmem:/tmp/demo8 -consumer ndi:demo01
```

File /tmp/demo8 should be existing...

I have also idea to associate **.shm** extension with **shmem** producer

### DIFF
--- a/src/modules/shmem/consumer_shmem.yml
+++ b/src/modules/shmem/consumer_shmem.yml
@@ -1,0 +1,34 @@
+schema_version: 0.2
+type: consumer
+identifier: shmem
+title: Shared memory output
+version: 3
+copyright: Copyright (C) 2016 Meltytech, LLC
+license: LGPL
+language: en
+creator: Maksym Veremeyenko
+tags:
+  - Audio
+  - Video
+description: Write frame to shared memory exchange interface
+
+notes: >
+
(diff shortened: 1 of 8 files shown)


src/modules/shmem/consumer_shmem.yml
```diff
@@ -0,0 +1,34 @@
+schema_version: 0.2
+type: consumer
+identifier: shmem
+title: Shared memory output
+version: 3
+copyright: Copyright (C) 2016 Meltytech, LLC
```
ddennedy: The code is copyright you, which justifies to put it in a separate module. Thus, you should change copyright to yourself in the yml. Also, consider to set the version to 1 (unless you need it this level for your environment). Same applies to the producer yml.

bmatherly: How about moving the code to the "plus" module? It doesn't really have any  external dependencies (except Linux).

PR #158, opened 2016-12-20
Ensure KDE 4 code is built when requested

xzhavilla:
Due to incomplete updates to `qt/configure`, KDE code is now forcibly disabled (both in the `configure` and in the `build` stage).

### DIFF
--- a/src/modules/qt/configure
+++ b/src/modules/qt/configure
@@ -195,9 +195,9 @@ else
 			if [ -d "$kde_includedir" ] && [ -d "$kde_libdir" ]
 			then
 				echo "- KDE version 4.x detected, will enable extra image formats"
-				echo "#define USE_KDE4" >> config.h
 				echo "USE_KDE4=1" >> config.mak
 				echo KDECXXFLAGS=-I$kde_includedir >> config.mak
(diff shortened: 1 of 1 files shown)


src/modules/qt/configure
```diff
@@ -148,6 +148,7 @@ else
 		if [ $? -eq 0 ]
 		then
 			echo "- Qt version 5.x detected"
+			without_kde=true
```
ddennedy: This looks like it will break KDE support on KDE5.

ddennedy:
@j-b-m How are we supporting the qt module with Qt 5 with KDE? This configure script seems to be KDE4 specific in its KDE support.

src/modules/qt/configure
xzhavilla: This is already the case:
```
echo "- Qt version 5.x detected"
# TODO re-enable KDE support when KDE Frameworks 5 widely available
without_kde=true
```

j-b-m:
Sorry for the late answer. Back in Qt4/KDE4, adding extra image formats like gimp's xcf required linking with some KDE libraries. Since Qt5/KF5, this is not necessary anymore, the extra image formats are a runtime dependency that doesn't require any linking, so KDE specific code in qimage producer is not required anymore for Qt5

PR #163, opened 2017-01-02
Pixbuf: fix alpha_buffer creation when no conversion is required

j-b-m:
Following a Kdenlive bug reported by mail, it looks like the pixbuf producer did not correctly create the alpha buffer in some cases (if no format conversion is required). This caused an incorrect transparency. Before committing, I would like to have Dan's opinion on this patch in case I misunderstood the alpha_mask feature..

### DIFF
--- a/src/modules/gtk2/producer_pixbuf.c
+++ b/src/modules/gtk2/producer_pixbuf.c
@@ -507,11 +507,11 @@ static void refresh_image( producer_pixbuf self, mlt_frame frame, mlt_image_form
 		}
 		pthread_mutex_unlock( &g_mutex );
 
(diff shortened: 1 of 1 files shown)


ddennedy:
mlt_frame should not always have an alpha buffer. It should only be added to save an alpha channel for an image stored in a format without alpha. An rgb to rgba conversion must add one. If something requests an alpha buffer and it does not exist, it should treat it as opaque or transparent.

j-b-m:
Thanks Dan I had not realized the obvious fact that an alpha mask is not necessary with an rgba image type. Now I understand the problem that is in my qtblend transition.

PR #173, opened 2017-01-20
Make mlt_playlist_insert_at conformant with documentation

alcinos:
The mode parameter was ignored half of the time.
The doc was also wrong about the return value, and I updated it to make it clearer about what is happenning.

I'm also thinking about adding in and out parameters to this function (defaulting to -1, so that it doesn't break any code relying on this). What do you think?

### DIFF
--- a/src/framework/mlt_playlist.c
+++ b/src/framework/mlt_playlist.c
@@ -1819,13 +1819,17 @@ void mlt_playlist_pad_blanks( mlt_playlist self, mlt_position position, int leng
 }
 
 /** Insert a clip at a specific time.
+ * If the insertion is done in overwrite mode (mode=true):
+ *      - if the position is a blank then it is replaced by the clip. If the blank was shorter than the clip, then the following clip is pushed to make room for the new clip.
+        - if the position is inside a clip, then the new clip is inserted before or after the old one, depending on whether the position is past the middle of the old clip or not. In case it is inserted after, if there is some blank after, it is replaced by the clip similarly as before
+ * If the insertion is not done in overwrite mode, then the behaviour is the same except that blanks are considered as normal clips (and thus never deleted).
  *
  * \public \memberof mlt_playlist_s
  * \param self a playlist
  * \param position the time at which to insert
  * \param producer the producer to insert
  * \param mode true if you want to overwrite any blank section
- * \return true if there was an error
+ * \return -1 if there was an error, else returns the index of the inserted clip
  */
 
 int mlt_playlist_insert_at( mlt_playlist self, mlt_position position, mlt_producer producer, int mode )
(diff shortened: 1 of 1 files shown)


ddennedy:
I am thinking this is going to take a while to analyze and test any potential regressions it may cause to existing applications. That means, it may take a long time because I have more important things to do. It is safer to fix the documentation and add a new function.

src/framework/mlt_playlist.c
```diff
@@ -1819,13 +1819,17 @@ void mlt_playlist_pad_blanks( mlt_playlist self, mlt_position position, int leng
 }
 
 /** Insert a clip at a specific time.
+ * If the insertion is done in overwrite mode (mode=true):
+ *      - if the position is a blank then it is replaced by the clip. If the blank was shorter than the clip, then the following clip is pushed to make room for the new clip.
+        - if the position is inside a clip, then the new clip is inserted before or after the old one, depending on whether the position is past the middle of the old clip or not. In case it is inserted after, if there is some blank after, it is replaced by the clip similarly as before
+ * If the insertion is not done in overwrite mode, then the behaviour is the same except that blanks are considered as normal clips (and thus never deleted).
```
ddennedy: There should be a blank comment line between the doxygen brief doc and the extended description. The doxygen description should be hard wrapped at 80 columns.

src/framework/mlt_playlist.c
```diff
@@ -1839,7 +1843,7 @@ int mlt_playlist_insert_at( mlt_playlist self, mlt_position position, mlt_produc
 		mlt_playlist_clip_info info;
 		mlt_playlist_get_clip_info( self, &info, clip );
 		mlt_events_block( properties, self );
-		if ( clip < self->count && mlt_playlist_is_blank( self, clip ) )
+		if ( mode == 1 && clip < self->count && mlt_playlist_is_blank( self, clip ) )
```
ddennedy: Do not compare with 1. Anything non-zero should be true.

src/framework/mlt_playlist.c
```diff
@@ -1872,13 +1876,8 @@ int mlt_playlist_insert_at( mlt_playlist self, mlt_position position, mlt_produc
...
 		{
-			if ( mode == 1 ) {
-				if ( position == info.start )
-					mlt_playlist_remove( self, clip );
-				else
-					mlt_playlist_blank( self, position - mlt_properties_get_int( properties, "length" ) - 1 );
-			}
-			mlt_playlist_append( self, producer );
+      mlt_playlist_blank( self, position - mlt_properties_get_int( properties, "length" ) - 1 );
+      mlt_playlist_append( self, producer );
```
ddennedy: Use tabs for indentation.

PR #175, opened 2017-01-28
Sliced interlaced-aware image convert implemented

max-verem:
This is a preview version of sliced interlaced-aware pixel format converter. Please, do a tests.


max-verem:
if you failed to compile against your currently used ffmpeg, please follow a patch https://github.com/max-verem/mlt/commit/cb39a14e087b735d1aed527b2ed11fef7a22194f

Issue #172, opened 2017-01-17
interlaced SD&HD PAL&NTSC 4:2:0 mpeg2/dv/h264 content has chroma upsample error on decklink consumer 422 +workaround


walterav1984:
Problem:
During the playback of Red and or Blue high saturated colors and or contrast within interlaced motion area's in SD PAL mpeg2/h264/dv video content which uses yuv interlaced 4:2:0 chroma sampling, there is visible artifacts/distortion like jaggy edges especially moving Red/Orange color contrast on blacks during playback on a SDL/Decklink consumer connected to CRT SD monitor. This ~~might~~ is also the case for NTSC & HD content, but I don't have interlaced HD broadcast CRT monitor to check (HDMI LCD).

`melt -profile dv_pal sdinterlaced420mpeg2dvh264examplevideofile -consumer decklink` #plays incorrect with chroma artifacts

Analyses:
From video file to decklink output the following has to take place, detect field order of file and if necessary change it to field order of SDI decklink output. Followed by the detection of chroma sampling level 4:2:0 and upscale it to decklink native 4:2:2 (possible a lot more routines happen and maybe in different order). The interlacing and field order is detected correctly for (mpeg2dvd upper field & DV lower field), because motion with less saturation/contrast is playback correct and smooth. But the problem is chroma upscaling probably happens "progressive" instead of "interlaced".
Its indeed C.U.E. (chroma upsampling error) because it is mostly visible in the motion affected areas of footage, on contrary  (ICP) interlaced chroma problem is mostly visible in static areas of the footage, see here for explanation:

http://web.archive.org/web/20060719025947/http://www.hometheaterhifi.com/volume_8_2/dvd-benchmark-special-report-chroma-bug-4-2001.html
https://trac.ffmpeg.org/ticket/4168

FIX/Workaround?
As of my understanding "melt" is build around ffmpeg, the following ffmpeg filter libswscale which is also part of melt can fix this in ffmpeg with "-vf scale=interl=1" and choosing a 422 or 4:4:4 output video codec . Investigation by ddennedy / max-verem shows that this interlaced aware code in the melt scaler equivalent is missing.

Workaround PRECONVERT footage:
`ffmpeg -f mpeg -i bug.vob -vf scale=interl=1 -c:v prores test.mov
melt -profile dv_pal test.mov -consumer decklink #plays correct on CRT monitor no chroma errors`

Workaround LIVE/REALTIME streams:
[...]

PR #177, opened 2017-01-29
Sliced interlaced-aware image convert implemented

max-verem:
This rework implement sliced interlace-aware pixel format convertor that should fix artifacts on interlaced 4:2:0 sources

Please do a tests if works properly

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1246,6 +1246,69 @@ static int pick_av_pixel_format( int *pix_fmt )
 	return 0;
 }
 
+struct sliced_pix_fmt_conv_t
+{
+	int width, height;
+	AVFrame *frame;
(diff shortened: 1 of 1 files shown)


bmatherly:
I understand that the code processes the picture in 10 slices. But the slices are not processed in separate threads. So what is the advantage to processing in slices?

ddennedy:
@bmatherly I think he is just preparing for threaded processing by using a slice context and function while we wait to verify it fixes the bug.

bmatherly:
Makes sense. I suppose we will also want to optimize the number of slices when the time comes.

walterav1984:
It fixes chroma upsampling error :-D , but creates a new kind of field sync error ;-(.  Please see #172

walterav1984:
I can confirm this looks Fixed!

PR #181, opened 2017-02-03
Avformat fixes1

max-verem:
This pull should fix https://github.com/mltframework/mlt/issues/178 and https://github.com/mltframework/mlt/issues/172

Issue #166, opened 2017-01-05
melt 6.5.0 hangs on certain .mp4 video file


TheDiveO:
My HDMI recorder generated an MP4 video file (~6MB) which plays and seeks totally fine in ffplay 3.0.5, but hangs melt 6.5.0, when seeking around a certain frame region (approx. frame 383 or so). A normal play from start to end won't hang melt, but seeking in the video around the problematic region causes melt to stop responding and causing some moderate, permanent CPU load; probably a single core. I've only found this issue because Kdenlive hangs while seeking this specific MP4 video in its clip monitor.

My OS is Kubuntu 16.10; ffplay is 3.0.5-0ubuntu0.16.10.1 from the Ubuntu repos; melt is 6.5.0 from the Kdenlive master ppa, which is compiled from this repository, as far as I know.

walterav1984:
This is also the case for me with melt 6.5.x with patch #182 and h264 1080i_50 broadcast transport streams on decklink consumer atsc_1080i_50, it will only play less than the first ~95 frames than it stops. Playback with melt 6.0.0 from the ubuntu 16.04 repo is not affected and plays correctly.

FFProbe output:
`Input #0, mpegts, from '/home/ubuntuadmin/Downloads/NPO-2-HD.ts':
  Duration: 00:38:49.06, start: 82567.858511, bitrate: 13359 kb/s
  Program 19402 
    Metadata:
      service_name    : NPO 2 HD
      service_provider: ZiggoL
    Stream #0:0[0x8fd]: Video: h264 (High) ([27][0][0][0] / 0x001B), yuv420p(tv, bt709), 1920x1080 [SAR 1:1 DAR 16:9], 25 fps, 50 tbr, 90k tbn, 50 tbc
    Stream #0:1[0x907](dut): Audio: mp2 ([3][0][0][0] / 0x0003), 48000 Hz, stereo, s16p, 256 kb/s
    Stream #0:2[0x908](dut): Audio: ac3 (AC-3 / 0x332D4341), 48000 Hz, 5.1(side), fltp, 448 kb/s
    Stream #0:3[0x909](GOS): Audio: mp2 ([3][0][0][0] / 0x0003), 48000 Hz, stereo, s16p, 256 kb/s (visual impaired)
    Stream #0:4[0x961](dut): Subtitle: dvb_teletext ([6][0][0][0] / 0x0006), 492x250
[...]

PR #182, opened 2017-02-03
Vertical slices

max-verem:
Use vertical slices to fix https://github.com/mltframework/mlt/issues/172

### DIFF
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -803,8 +803,9 @@ static AVStream *add_video_stream( mlt_consumer consumer, AVFormatContext *oc, A
 		st->time_base = c->time_base;
 
 		// Default to the codec's first pix_fmt if possible.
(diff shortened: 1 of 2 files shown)


ddennedy:
According to the thread in #172 it sounds like this was not successful on NTSC. Should we consider to close this and merge #181 instead?

walterav1984:
This pull request also fixes HD interlaced h264 4:2:0 chroma upsampling errors but freezes playback of HD h264 after around ~100 frames(might be CPU bound).  Please look for further debug info #166.

walterav1984:
Freezing playback of HD h264 interlaced 1080i50 4:2:0 after ~100 frames is caused in commit fb09410.

Building and running melt commit 58d8ce does not show this behavior freezing behavior for HD h264 1080i50 4:2:0 footage.

max-verem:
please provide a sample that reproduce freezing

PR #183, opened 2017-02-04
Add mlt_slices_get_global()

ddennedy:
This does not remove mlt_slices_init_pool(), but I think we should perhaps do that. Comments welcome,

I chose not to include "idle" as a policy option and split "rt" into "rr" and "fifo" to better closely match POSIX  standard and remove ambiguity (e.g. not clear rt = fifo).

I chose to not put mlt_slices_get_global() into mlt_factory because mlt_factory is mainly used for services, and I just wanted to keep slice-related stuff together.

TODO: 
- ~remove mlt_slices_init_pool()~
- ~convert other services using mlt_slices_init_pool()~
- ~Document environment variables in mlt_slices.h.~
- ~move code mapping mlt_schedule_policy to SCHED_ constants from mlt_slices_get_global() to mlt_slices_init()~ (removing because changing param to mlt_slices_init breaks ABI).
- ~MLT_GLOBAL_SLICES or MLT_GLOBAL_SLICES_COUNT?~ (MLT_SLICES_COUNT as before)

### DIFF
--- a/src/framework/mlt_slices.c
+++ b/src/framework/mlt_slices.c
@@ -40,7 +40,6 @@
 #define ENV_SLICES "MLT_SLICES_COUNT"
 
 static pthread_mutex_t g_lock = PTHREAD_MUTEX_INITIALIZER;
-static mlt_properties pool_map = NULL;
 static mlt_slices globals[mlt_policy_nb] = {NULL, NULL, NULL};
 
 
@@ -133,6 +132,7 @@ static void* mlt_slices_worker( void* p )
 /** Initialize a sliced threading context
  *
  * \public \memberof mlt_slices_s
+ * \deprecated
  * \param threads number of threads to use for job list, 0 for #cpus
  * \param policy scheduling policy of processing threads, -1 for normal
  * \param priority priority value that can be used with the scheduling algorithm, -1 for maximum
@@ -212,6 +212,7 @@ mlt_slices mlt_slices_init( int threads, int policy, int priority )
 /** Destroy sliced threading context
  *
  * \public \memberof mlt_slices_s
+ * \deprecated
  * \param ctx context pointer
  */
 
@@ -231,14 +232,6 @@ void mlt_slices_close( mlt_slices ctx )
 		pthread_mutex_unlock( &g_lock );
 		return;
 	}
-
-	/* remove it from pool */
-	if ( ctx->name )
-	{
-		mlt_properties_set_data( pool_map, ctx->name, NULL, 0, NULL, NULL );
-		mlt_log_debug( NULL, "%s:%d: ctx=[%p][%s] removed\n", __FUNCTION__, __LINE__, ctx, ctx->name );
-	}
-
 	pthread_mutex_unlock( &g_lock );
 
 	/* notify to exit */
@@ -264,6 +257,7 @@ void mlt_slices_close( mlt_slices ctx )
 /** Run sliced execution
  *
  * \public \memberof mlt_slices_s
+ * \deprecated
  * \param ctx context pointer
  * \param jobs number of jobs to proccess
  * \param proc number of jobs to proccess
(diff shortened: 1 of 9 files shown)


src/modules/frei0r/frei0r_helper.c
```diff
@@ -104,23 +103,18 @@ int process_frei0r_item( mlt_service service, double position, double time, mlt_
...
 
-	mlt_service_lock( service );
-
 	if (slice_count >= 0) {
-		slices = mlt_properties_get_data(prop, "mlt_slices", NULL);
-		if (!slices)
-			slices = mlt_slices_init(slice_count, SCHED_OTHER, sched_get_priority_max(SCHED_OTHER));
-		if (slices) {
-			mlt_properties_set_data(prop, "mlt_slices", slices, 0, (mlt_destructor) mlt_slices_close, NULL);
+		if ((slices = mlt_slices_get_global(mlt_policy_normal)))
```
bmatherly: I might be missing something here. But it seems to me that this will never be true because slices is initialized to NULL above.

src/framework/mlt_slices.h
```diff
@@ -38,4 +39,12 @@ extern int mlt_slices_count( mlt_slices ctx );
 
 extern mlt_slices mlt_slices_init_pool( int threads, int policy, int priority, const char* name );
 
+extern mlt_slices mlt_slices_get_global( mlt_schedule_policy policy );
```
bmatherly: I would question the need for this function to be public. Is it just so that it is possible to call mlt_slices_count()? If so, just make a new version of mlt_slices_count() that returns the global number of slices. It could be called:

    extern int mlt_slices_count_global();

It think it would be valid for the implementation to never provide the global contexts to the public.

max-verem: leaving only 3 functions:
```
extern void mlt_slices_run_normal( int jobs, mlt_slices_proc proc, void* cookie );
extern void mlt_slices_run_rr( int jobs, mlt_slices_proc proc, void* cookie );
extern void mlt_slices_run_fifo( int jobs, mlt_slices_proc proc, void* cookie );
```
very good approach...
i dont know why *mlt_slices_count_global* could be needed, but i have not objections against it.

why to rename: *pool_lock* -> *g_lock* - global functionality reusing existing mutex... twice mentioned in new function but 5 times in previous code...

ddennedy: mlt_slices_count is used by frei0r, which requires to know the number of slices ahead of mlt_slices callback function to set dimension of slice on frei0r API.

ddennedy: > why to rename: pool_lock -> g_lock

Because it was no longer specific to the properties pool, and I like to have a g_ prefix on global names.

src/modules/frei0r/frei0r_helper.c
ddennedy: The result of this assignment is almost always likely to be true.

src/framework/mlt_slices.h
ddennedy: I agree with the all the hiding suggestions as well.

ddennedy:
Lots of commits this weekend based on review comments. Might be easier to look at mlt_slices.[ch] here:
https://github.com/mltframework/mlt/tree/mlt_slices_get_global/src/framework

src/framework/mlt_slices.c
```diff
@@ -306,63 +316,71 @@ void mlt_slices_run( mlt_slices ctx, int jobs, mlt_slices_proc proc, void* cooki
...
-int mlt_slices_count(mlt_slices ctx)
+int mlt_slices_count()
+{
+	mlt_slices slices = mlt_slices_get_global( mlt_policy_normal );
+	if (slices)
+		return slices->count;
+	if ((slices = mlt_slices_get_global( mlt_policy_fifo )))
+		return slices->count;
+	if ((slices = mlt_slices_get_global( mlt_policy_rr )))
+		return slices->count;
```
bmatherly: I'm not sure I understand these other checks. The first call to mlt_slices_get_global will create a slices context if one does not exist. So these other calls to mlt_slices_get_global() will never occur.

The disadvantage to this logic is that the first call to mlt_slices_count() will create the global slices context for mlt_policy_normal even if no services are going to use it.

Perhaps you want to try something like this instead:

    int mlt_slices_count()
    {
        int i = 0;
        int slices = 0;
        pthread_mutex_lock( &g_lock );
        for( i = mlt_policy_normal; i < mlt_policy_nb ); i++ )
        {
            if( globals[i] )
            {
                slices = globals[i]->count;
                break;
            }
        }
        pthread_mutex_unlock( &g_lock );
        return slices;
    }

bmatherly:
This all looks good to me. I only had one comment which was a nit.

src/framework/mlt_slices.c
ddennedy: Good point; however, your logic requires there to be some previous call such as mlt_slices_run_ to create a context. However, where this is used such as in frei0r, is not the correct place to call run. We could make separate _normal, _rr, and _fifo versions of these since the service will know which policy it wants.

bmatherly: Correct: there would be a race condition in my proposal. Maybe there should be a single global slice count at the top of the file that gets calculated once from whatever function gets called first.

Anyway this whole comment is probably a micro-optimization. I just thought I would mention it.

ddennedy: See my new commit to replace mlt_slices_count().

max-verem:
when are you going to push this changes?

ddennedy:
@max-verem As soon as you say "OK with me." :)

max-verem:
LGTM
:-)

PR #184, opened 2017-02-05
Minor fixes 20170205

max-verem:
Please review and merge

### DIFF
--- a/src/modules/decklink/consumer_decklink.cpp
+++ b/src/modules/decklink/consumer_decklink.cpp
@@ -756,14 +756,6 @@ protected:
 				uint32_t written = 0;
 #endif
 				BMDTimeValue streamTime = m_count * frequency * m_duration / m_timescale;
-				m_deckLinkOutput->GetBufferedAudioSampleFrameCount( &written );
-				mlt_log_debug( getConsumer(), "%s:%d GetBufferedAudioSampleFrameCount=" DECKLINK_UNSIGNED_FORMAT "\n",
-					__FUNCTION__, __LINE__, written );
-				if ( written > (m_preroll + 1) * samples )
-				{
-					mlt_log_verbose( getConsumer(), "renderAudio: will flush " DECKLINK_UNSIGNED_FORMAT " audiosamples\n", written );
-					m_deckLinkOutput->FlushBufferedAudioSamples();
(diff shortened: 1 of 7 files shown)


src/modules/decklink/consumer_decklink.cpp
```diff
@@ -575,7 +575,7 @@ class DeckLinkConsumer
 					ssize_t size = stride * height;
 
 					// convert lower field first to top field first
-					if ( !progressive )
+					if ( !progressive && m_displayMode->GetFieldDominance() == bmdUpperFieldFirst )
```
ddennedy: Today, it seems consumer_decklink assumes MLT field order is bottom field first, but it really all depends on the "top_field_first" property of mlt_frame, and the consumer property "top_field_first determines" the output field order (default 0). I think this is the only issue I have with this set of patches, but now it blocks merging the pull request.

src/modules/decklink/consumer_decklink.cpp
```diff
@@ -599,6 +587,9 @@ class DeckLinkConsumer
 
 					if ( !progressive && m_displayMode->GetFieldDominance() == bmdUpperFieldFirst )
 					{
+						// Make the first line black for field order correction.
```
bmatherly: I agree with Dan's comments about using "top_field_first" property to determine the mlt frame field order.

Another way to correct the field order would be to delay the output by one field. Display the upper field from this frame and the lower field from the next frame.

src/modules/decklink/consumer_decklink.cpp
walterav1984: Keep in mind this NTSC patch fixes playback for allot of native NTSC example video's besides the ones I correctly created. Try the SD25 SD50 and MPEG examples with this f70d47c build and some previous original melt builds.

http://www.gvgdevelopers.com/concrete/products/k2/test_clips/

Footage with DV 4:1:1 lowerfield plays fine with this patch, also rawvideo uyvy422 that plays natively from Blackmagic Design Media Express also plays correct with this version. This is without any XML hacks like progressive=0 top_field_first=1.

How can this be wrong?

max-verem:
i will provide a interlacing fix in separate branch

PR #188, opened 2017-02-10
Sliced parallel-interlaced-aware pixel format convertor

max-verem:
please review and merge

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -25,6 +25,7 @@
 #include <framework/mlt_deque.h>
 #include <framework/mlt_factory.h>
 #include <framework/mlt_cache.h>
+#include <framework/mlt_slices.h>
 
 // ffmpeg Header files
 #include <libavformat/avformat.h>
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -1246,6 +1247,102 @@ static int pick_av_pixel_format( int *pix_fmt )
...
+	if ( AV_PIX_FMT_YUV420P == ctx->dst_format )
+		dst_v_chr_pos = ( !interlaced ) ? 128 : ( !field ) ? 64 : 192;
+
+	mlt_log_debug( NULL, "%s:%d: [id=%d, idx=%d, jobs=%d], interlaced=%d, field=%d, slices=%d, mul=%d, h=%d, slice_w=%d, slice_x=%d ctx->src_desc=[log2_chroma_h=%d, log2_chroma_w=%d], src_v_chr_pos=%d, dst_v_chr_pos=%d\n",
+		__FUNCTION__, __LINE__, id, idx, jobs, interlaced, field, slices, mul, h, slice_w, slice_x, ctx->src_desc->log2_chroma_h, ctx->src_desc->log2_chroma_w, src_v_chr_pos, dst_v_chr_pos );
+
+	if ( slice_w <= 0 )
+		return 0;
+
+	sws = sws_alloc_context();
```
bmatherly: I don't have a lot of experience with the SWS API. What is the cost of allocating a new context for each slice? Is it expensive?

Would it be possible to create a single context for each frame and then have each slice use the same context?

Would it be possible to create a single context on initialization and use the same context for all slices and all frames?

src/modules/avformat/producer_avformat.c
```diff
@@ -1332,20 +1429,48 @@ static int convert_image( producer_avformat self, AVFrame *frame, uint8_t *buffe
...
-			result = profile->colorspace;
-		sws_scale( context, (const uint8_t* const*) frame->data, frame->linesize, 0, height,
-			output.data, output.linesize);
-		sws_freeContext( context );
+		ctx.src_desc = av_pix_fmt_desc_get( ctx.src_format );
+		ctx.dst_desc = av_pix_fmt_desc_get( ctx.dst_format );
+
+		avpicture_fill( ctx.output, buffer, ctx.dst_format, width, height );
+
+		if ( getenv("MLT_AVFORMAT_SLICED_PIXFMT") )
```
bmatherly: Do we want this to be an option? If it works, why not always uses it?

max-verem: > I don't have a lot of experience with the SWS API. What is the cost of allocating a new context for each slice? Is it expensive?

i think it expensive, but it get a performance gain: https://github.com/mltframework/mlt/issues/172#issuecomment-278329812

> Would it be possible to create a single context for each frame and then have each slice use the same context?

i though about it, but it will require some code to check if sws init condition changes

> Would it be possible to create a single context on initialization and use the same context for all slices and all frames?

i tried that - it is not working.

> Do we want this to be an option? If it works, why not always uses it?

for first time if some artifact appear on a picture we could swap processing method to check if it caused by sliced processing...

bmatherly: > i though about it, but it will require some code to check if sws init condition changes

I see. So maybe it could be a future optimization.

> for first time if some artifact appear on a picture we could swap processing method to check if it caused by sliced processing...

I understand. Perhaps we could remove the check in the future after any bugs are discovered and resolved.

max-verem: >> i though about it, but it will require some code to check if sws init condition changes

> I see. So maybe it could be a future optimization

 did some checks (https://github.com/max-verem/mlt/commit/30b52552da686ca3884b2f600e40e09147497e2f):
```
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=407 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=416 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=394 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=413 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=1002 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=375 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=359 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=374 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=358 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=353 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=355 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=356 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=353 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=391 us
Current Position:         15
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=373 us
sliced_h_pix_fmt_conv_proc:1318: DUR(sws_context_init)=1582 us
convert_image:1485: DUR(convert_image)=10054 us
producer_get_image:1984: DUR(producer_get_image)=15155 us
```
average 356 us

max-verem: >> for first time if some artifact appear on a picture we could swap processing method to check if it caused by sliced processing...

> I understand. Perhaps we could remove the check in the future after any bugs are discovered and resolved.

i can reverse it by using `MLT_AVFORMAT_SLICED_PIXFMT_DISABLE` instead of current, making current method default

bmatherly: > i can reverse it by using MLT_AVFORMAT_SLICED_PIXFMT_DISABLE instead of current, making current method default

My opinion is that sliced conversion should be default and it can be disabled via env in case it causes someone a problem. After some time in used and bugs are worked out, then the env can be removed.

max-verem: OK
i will post updated version

PR #190, opened 2017-02-10
Timestamped logging and timings measurement

max-verem:
This pull request (it nest https://github.com/mltframework/mlt/pull/189) makes debugging of some performance depended task much easy - it add timing information to log messages and provides a method for measuring time of execution of some part of code.

Main changes comes in https://github.com/mltframework/mlt/commit/eafb820bc6888ed05e584f2bf63d149e1e295720

### DIFF
--- a/src/framework/mlt_log.c
+++ b/src/framework/mlt_log.c
@@ -23,6 +23,8 @@
 #include "mlt_service.h"
 
 #include <string.h>
+#include <sys/time.h>
+#include <time.h>
 
 static int log_level = MLT_LOG_WARNING;
 
@@ -33,6 +35,23 @@ void default_callback( void* ptr, int level, const char* fmt, va_list vl )
 	
 	if ( level > log_level )
 		return;
+
+	if ( print_prefix && level >= MLT_LOG_VERBOSE )
+	{
+		struct timeval tv;
+		time_t ltime;
+		struct tm *rtime;
+		char buf[32];
+
+		gettimeofday(&tv, NULL);
+		ltime = tv.tv_sec;
+
+		rtime = localtime( &ltime );
+		strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S" , rtime );
+
+		fprintf( stderr, "| %s.%.3d | ", buf, (int)(tv.tv_usec / 1000) );
+	}
+
 	if ( print_prefix && properties )
 	{
 		char *mlt_type = mlt_properties_get( properties, "mlt_type" );
@@ -81,4 +100,18 @@ void mlt_log_set_level( int level )
 void mlt_log_set_callback( void (*new_callback)( void*, int, const char*, va_list ) )
 {
 	callback = new_callback;
+}
+
+int64_t mlt_log_timings_now( void )
+{
+	int64_t r;
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
+	r = tv.tv_sec;
+
+	r *= 1000000LL;
+	r += tv.tv_usec;
+
+	return r;
 }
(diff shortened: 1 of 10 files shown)


src/framework/mlt_log.c
```diff
@@ -82,3 +101,17 @@ void mlt_log_set_callback( void (*new_callback)( void*, int, const char*, va_lis
...
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
+	r = tv.tv_sec;
+
+	r *= 1000000LL;
+	r += tv.tv_usec;
+
+	return r;
+}
```
ddennedy: How do you feel about making mlt_log_timings_now() a noop if NDEBUG is defined?
http://en.cppreference.com/w/c/error/assert

src/framework/mlt_log.h
```diff
@@ -90,4 +91,16 @@ int mlt_log_get_level( void );
 void mlt_log_set_level( int );
 void mlt_log_set_callback( void (*)( void*, int, const char*, va_list ) );
 
+#define mlt_log_timings_begin() \
+{ \
+	int64_t _mlt_log_timings_begin = mlt_log_timings_now(), _mlt_log_timings_end;
+
+#define mlt_log_timings_end(service, msg) \
+	_mlt_log_timings_end = mlt_log_timings_now(); \
+	mlt_log_verbose( service, "%s:%d: T(%s)=%" PRId64 " us\n", \
```
ddennedy: A lot of users or possibly even apps like Kdenlive or Flowblade use verbose log level, and I am not sure we want this volume. OTOH debug level is so verbose it requires text filtering. Maybe add a new level like MLT_LOG_TIMINGS?

max-verem:
> I do not want to use sliced pix_fmt conversion across all versions that we support. I do not like the risk for the bugs that may introduce. Rather, it should be limited to the lower version that @walterav1984 tested and up.

please elaborate

ddennedy:
@walterav1984 said he is using FFmpeg version 2.8.10. So, this has been tested on that version and something very recent by you. Make a build time dependency on [libswscale version 3.1.101 or higher](https://github.com/FFmpeg/FFmpeg/blob/n2.8.10/libswscale/version.h) to use sliced processing with swscale.

max-verem:
> What compiler warning for the lack of space characters? I have not seen that.

```
mlt_slices.c: In function ‘mlt_slices_init’:
mlt_slices.c:201:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
     if ( priority < 0 )
     ^~
mlt_slices.c:203:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
  pthread_attr_setschedpolicy( &tattr, policy );
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

max-verem:
> Maybe add a new level like MLT_LOG_TIMINGS?

what about:
```
#define MLT_LOG_TIMINGS 44
```
?

max-verem:
> @walterav1984 said he is using FFmpeg version 2.8.10. So, this has been tested on that version and something very recent by you. Make a build time dependency on libswscale version 3.1.101 or higher to use sliced processing with swscale.

do you mean we have to check minimal version of ffmpeg to use this method and leave original for older version? what checking macro should i use then?

walterav1984:
I can do more tests with newer versions of ffmpeg/chroma upsampling but I / we need to make somekind testing framework with multiple formats / footage and use cases because this incredible great sliced chroma upsampling code may affect alot of use cases. From SD-HD NTSC-PAL 4:2:0 - 4:2:2 playback, field orders, SD downconvertion by melt (HD footage in SD profile) etc etc.

ddennedy:
> do you mean we have to check minimal version of ffmpeg to use this method

Yes

> leave original for older version?

Yes

> what checking macro should i use then?

Really? See all the other library version checking macros used in the same file!

ddennedy:
> #define MLT_LOG_TIMINGS 44

Yes

PR #194, opened 2017-02-15
yuv422p16: please review

max-verem:
This branch provides handling 10bit capturing.

Regression testing:

```
melt -profile dv_pal decklink:0 10=1 -consumer avformat:/tmp/test1-`date '+%H%M%S'`.mov mlt_image_format=yuv422p16 properties=DV f=mov threads=12

melt -profile dv_pal decklink:0 10=1 -consumer avformat:/tmp/test2-`date '+%H%M%S'`.mov properties=DV f=mov threads=12

melt -profile dv_pal decklink:0 -consumer avformat:/tmp/test3-`date '+%H%M%S'`.mov mlt_image_format=yuv422p16 properties=DV f=mov threads=12

melt -profile dv_pal decklink:0 -consumer avformat:/tmp/test4-`date '+%H%M%S'`.mov properties=DV f=mov threads=12
```

10-bit capturing:

```
melt -profile dv_pal decklink:0 10=1 -consumer avformat:/tmp/test10-`date '+%H%M%S'`.mov mlt_image_format=yuv422p16 properties=ProRes threads=12
```

### DIFF
--- a/src/framework/mlt_frame.c
+++ b/src/framework/mlt_frame.c
@@ -410,9 +410,34 @@ const char * mlt_image_format_name( mlt_image_format format )
 		case mlt_image_opengl:  return "opengl";
 		case mlt_image_glsl:    return "glsl";
 		case mlt_image_glsl_texture: return "glsl_texture";
+		case mlt_image_yuv422p16: return "yuv422p16";
+		case mlt_image_last:    return "last";
 	}
 	return "invalid";
 }
+
+/** Get the id of image format from short name.
+ *
+ * \public \memberof mlt_frame_s
+ * \param name the image format short name
+ * \return a image format
+ */
+
+mlt_image_format mlt_image_format_id( const char * name )
+{
+	mlt_image_format f;
+
+	for(f = mlt_image_none; f < mlt_image_last; f++)
+	{
+		const char * v = mlt_image_format_name( f );
+		if(!strcmp( v, name ) )
+			return f;
+	}
+
+	return mlt_image_last;
+}
+
+#define UP16(V) ( ( ( (V) + 15 ) / 16 ) * 16 )
 
 /** Get the number of bytes needed for an image.
   *
@@ -445,6 +470,9 @@ int mlt_image_format_size( mlt_image_format format, int width, int height, int *
 		case mlt_image_glsl_texture:
 			if ( bpp ) *bpp = 0;
 			return 4;
+		case mlt_image_yuv422p16:
+			if ( bpp ) *bpp = 2;
+			return height * ( UP16( width * 2 ) + 2 * UP16( width + 1 ) ) ;
 		default:
 			if ( bpp ) *bpp = 0;
 			return 0;
@@ -1132,4 +1160,64 @@ mlt_frame mlt_frame_clone( mlt_frame self, int is_deep )
 	}
 
 	return new_frame;
+}
+
+/** Build a planes pointers of image mapping
+ *
+ * For proper and unified planar image processing, planes sizes and planes pointers should
+ * be provides to processing code.
+ *
+ * \public \memberof mlt_frame_s
+ * \param format the image format
+ * \param width width of the image in pixels
+ * \param height height of the image in pixels
+ * \param[in] data pointer to allocated image
+ * \param[out] planes pointers to plane's pointers will be set
+ * \param[out] strides pointers to plane's strides will be set
+ * \return the number of bytes
+ */
+int mlt_image_format_planes( mlt_image_format format, int width, int height, void* data, unsigned char *planes[4], int strides[4])
+{
+	if ( mlt_image_yuv422p16 == format )
+	{
+		strides[0] = UP16( width * 2 );
+		strides[1] = UP16( width + 1 );
+		strides[2] = UP16( width + 1 );
+		strides[3] = 0;
+
+		planes[0] = (unsigned char*)data;
+		planes[1] = planes[0] + height * strides[0];
+		planes[2] = planes[1] + height * strides[1];
+		planes[3] = 0;
+	}
+	else if ( mlt_image_yuv420p == format )
+	{
+		strides[0] = width;
+		strides[1] = width >> 1;
+		strides[2] = width >> 1;
+		strides[3] = 0;
+
+		planes[0] = (unsigned char*)data;
+		planes[1] = (unsigned char*)data + width * height;
+		planes[2] = (unsigned char*)data + ( 5 * width * height ) / 4;
+		planes[3] = 0;
+	}
+	else
+	{
+		int bpp;
+
+		mlt_image_format_size( format, width, height, &bpp );
+
+		planes[0] = data;
+		planes[1] = 0;
+		planes[2] = 0;
+		planes[3] = 0;
+
+		strides[0] = bpp * width;
+		strides[1] = 0;
+		strides[2] = 0;
+		strides[3] = 0;
+	};
+
+	return 0;
 }
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -454,6 +454,8 @@ static enum AVPixelFormat pick_pix_fmt( mlt_image_format img_fmt )
 		return AV_PIX_FMT_RGBA;
 	case mlt_image_yuv420p:
 		return AV_PIX_FMT_YUV420P;
+	case mlt_image_yuv422p16:
+		return AV_PIX_FMT_YUV422P16LE;
 	default:
 		return AV_PIX_FMT_YUYV422;
 	}
@@ -1144,7 +1146,6 @@ static void *consumer_thread( void *arg )
 	// Need two av pictures for converting
 	AVFrame *converted_avframe = NULL;
 	AVFrame *audio_avframe = NULL;
-	AVFrame *video_avframe = NULL;
 
 	// For receiving audio samples back from the fifo
 	uint8_t *audio_buf_1 = av_malloc( AUDIO_ENCODE_BUFFER_SIZE );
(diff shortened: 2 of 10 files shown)


src/framework/mlt_frame.c
```diff
@@ -1133,3 +1161,63 @@ mlt_frame mlt_frame_clone( mlt_frame self, int is_deep )
...
+ * \public \memberof mlt_frame_s
+ * \param format the image format
+ * \param width width of the image in pixels
+ * \param height height of the image in pixels
+ * \param[in] data pointer to allocated image
+ * \param[out] planes pointers to plane's pointers will be set
+ * \param[out] strides pointers to plane's strides will be set
+ * \return the number of bytes
+ */
+int mlt_image_format_planes( mlt_image_format format, int width, int height, void* data, unsigned char *planes[4], int strides[4])
```
bmatherly: I would like some discussion on this function because it feels like it is optimized for ffmpeg libraries.

I would propose we replace this function with three functions: one to return the pointer to a plane, one to return the height and one to return the size of a line. The signatures would be:

```
void* mlt_image_plane_ptr( mlt_image_format format, int width, int height, void* data, size_t plane )
size_t mlt_image_plane_height( mlt_image_format format, int width, int height, void* data, size_t plane )
size_t mlt_image_plane_linesize( mlt_image_format format, int width, int height, void* data, size_t plane )
```

I feel that the discrete functions would be more generally useful throughout MLT. If you want an ffmpeg optimized function, you could create on in the avformat module that provides the behavior of this function using the three functions I propose.

I don't feel strongly about this, but I thought I would throw it out for discussion.

src/modules/avformat/consumer_avformat.c
```diff
@@ -1316,6 +1317,8 @@ static void *consumer_thread( void *arg )
 					img_fmt = mlt_image_rgb24a;
 				else if ( !strcmp( img_fmt_name, "yuv420p" ) )
 					img_fmt = mlt_image_yuv420p;
+				else if ( !strcmp( img_fmt_name, "yuv422p16" ) )
```
bmatherly: Should this entire case statement just be replaced with the new mlt_image_format_id() function?

src/framework/mlt_frame.c
```diff
@@ -410,10 +410,35 @@ const char * mlt_image_format_name( mlt_image_format format )
 		case mlt_image_opengl:  return "opengl";
 		case mlt_image_glsl:    return "glsl";
 		case mlt_image_glsl_texture: return "glsl_texture";
+		case mlt_image_yuv422p16: return "yuv422p16";
+		case mlt_image_last:    return "last";
```
bmatherly: I think that last should return "invalid" to be consistent with previous behavior. But I'm not sure.

Alternately, maybe we should remove "invalid" and always return "last" as a default.

Either way, I don't think there should be two different error returns. It should either be "invalid" or "last".

src/framework/mlt_frame.c
```diff
@@ -445,6 +470,9 @@ int mlt_image_format_size( mlt_image_format format, int width, int height, int *
 		case mlt_image_glsl_texture:
 			if ( bpp ) *bpp = 0;
 			return 4;
+		case mlt_image_yuv422p16:
+			if ( bpp ) *bpp = 2;
```
max-verem: that is luma plane bpp... or assume bpp as array of four int

src/framework/mlt_frame.c
```diff
@@ -445,6 +470,9 @@ int mlt_image_format_size( mlt_image_format format, int width, int height, int *
...
+			return height * ( UP16( width * 2 ) + 2 * UP16( width + 1 ) ) ;
```
max-verem: because XMM register is 16 bytes wide

max-verem:
> I would like some discussion on this function because it feels like it is optimized for ffmpeg libraries.

yes... it is ffmpeg's approach tested for many years

src/modules/avformat/consumer_avformat.c
max-verem: possibly... i though there is some specific customized

src/modules/core/filter_fieldorder.c
```diff
@@ -83,12 +83,21 @@ static int get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format
 			}
 
 			// Shift the entire image down by one line
-			int bpp;
```
max-verem: i would say: Non-vertical luma/chroma scaled

max-verem:
> To make the new format truly usable, we will have to add appropriate converter functions in core/filter_imageconvert.c
> 
> Without the converter functions, most filters will not be able to be used. I suppose that can be added in a future commit.

*avcolour_space* was updated for this functionality... it seems core/filter_imageconvert.c used only if mlt does not linked against ffmpeg... i can extend arrays and put NULL as conversion function to avoid memory access error...

src/framework/mlt_frame.c
bmatherly: By looking at the other formats, I assume that bpp is meant to indicate the overall bytes per pixel. That is, bpp = total bytes in the entire image / total pixels in the entire image.

src/framework/mlt_frame.c
bmatherly: I don't think this function is intended to return SMID optimized sizes. It doesn't do that for any of the other formats. @ddennedy do you have a preference about whether this function can return a number that could possibly be greater than the size actually needed to store the image?

src/modules/core/filter_fieldorder.c
bmatherly: Yes. That would be more accurate.

src/framework/mlt_frame.c
max-verem: > It doesn't do that for any of the other formats.

sad but true. approach above makes all rows aligned to 16 if head is aligned...

ddennedy:
>  it seems core/filter_imageconvert.c used only if mlt does not linked against ffmpeg

That is correct. Ideally, there are conversion functions for every mlt_image_ format in filter_imageconvert.c, but that is a lot of work and MLT is rarely used without FFmpeg now. 

> extend arrays and put NULL as conversion function

I do not follow you. Where will this go? What arrays?

src/framework/mlt_frame.c
ddennedy: I agree with "invalid".

src/framework/mlt_frame.c
ddennedy: The bpp param is not useful for computing stride for planar formats. Any code that is calling this to get bpp that can work with mlt_image_yuv422p16 is not going to use this param, and it does not matter. Might be safer to set bpp to zero to prevent it from being used for this image format.

src/framework/mlt_frame.c
ddennedy: For this function it should be width * height * 4. If that makes the function unusable in certain contexts then that code should use its own math.

src/modules/avformat/consumer_avformat.c
ddennedy: I agree with Brian.

src/framework/mlt_frame.c
max-verem: are you agree that rows would not be aligned to 16?

max-verem:
line 296: ( https://github.com/mltframework/mlt/blob/master/src/modules/core/filter_imageconvert.c#L296 )
```
static conversion_function conversion_matrix[5][5] = {
```
contains matrix with functions pointer.

i can extend it to 6x6 and set NULL values for function pointer. if somebody decides to use *yuv422p16* without ffmpeg, then he will have to implement it..

src/modules/avformat/consumer_avformat.c
```diff
@@ -1766,36 +1768,16 @@ static void *consumer_thread( void *arg )
...
-							q += stride / 4;
-							memcpy( video_avframe->data[2], q, video_avframe->linesize[2] * height / 2 );
-						}
-						else for ( i = 0; i < height; i ++ )
-						{
-							p = video_avframe->data[0] + i * video_avframe->linesize[0];
-							memcpy( p, q, stride );
-							q += stride;
-						}
+						mlt_image_format_planes( img_fmt, width, height, image, video_avframe.data, video_avframe.linesize );
```
max-verem: @ddennedy have you objections against this? new code avoiding one more memory copying...

src/modules/decklink/producer_decklink.yml
```diff
@@ -113,3 +113,12 @@ parameters:
       It skips frames that has VITC timecode less then specified.
       After reaching first frame with timecode greater or equal then specified this
       property is reset to zero.
+
+  - identifier: 10
```
max-verem: i am also thing about renaming it to *mlt_image_format* like consumer has

max-verem:
i am going to prepare new pull for this functionality, please provide more comments and objections

src/framework/mlt_frame.c
ddennedy: Yes, I agree. Images and image rows on mlt_frame are not required to be 16 byte aligned. MLT framework does not impose these requirements on its services such that implementations of signal processing can be simper and less optimized. Major rework to deal with all that would be required. I think it is great that FFmpeg has support and optimizations for these things, but this is not FFmpeg, and I do not have the time and energy to recreate what they have done (or ensure someone else's changes does not break shit). Also, maybe there should be mlt_image_avformat and additional frame properties that allows any FFmpeg pixel format to be used only within the avformat module.

src/modules/decklink/producer_decklink.yml
ddennedy: yeah, I did not like "10" when I saw it. I was going to suggest "10bit". It is not good behavior to add new image format conversion code into a producer in order to comply with a requested image format unless it already has something at its disposal (for example, avformat producer can use libswscale to comply). So, mlt_image_format does not seem correct to me.

bmatherly: Suggestion:

```
identifier: bitdepth
values
  - 8
  - 10
```

PR #197, opened 2017-02-20
Set 'progressive', 'top_field_first' property

max-verem:
attached set unifies setting 'progressive', 'top_field_first' property for NDI consumer, drop field order changing and set 'progressive' property of decklink consumer

### DIFF
--- a/src/modules/decklink/consumer_decklink.cpp
+++ b/src/modules/decklink/consumer_decklink.cpp
@@ -392,6 +392,7 @@ protected:
 			return false;
 		}
 		mlt_properties_set_int( properties, "top_field_first", m_displayMode->GetFieldDominance() == bmdUpperFieldFirst );
+		mlt_properties_set_int( properties, "progressive", m_displayMode->GetFieldDominance() == bmdProgressiveFrame );
 
 		// Set the keyer
 		if ( m_deckLinkKeyer && ( m_isKeyer = mlt_properties_get_int( properties, "keyer" ) ) )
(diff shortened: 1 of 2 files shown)


src/modules/ndi/consumer_ndi.c
```diff
@@ -280,6 +255,11 @@ static int consumer_ndi_start( mlt_consumer consumer )
 
 	mlt_properties properties = MLT_CONSUMER_PROPERTIES( consumer );
 
+	// Set progressive/interlacing properties
+	mlt_profile profile = mlt_service_profile( MLT_CONSUMER_SERVICE( consumer ) );
+	mlt_properties_set_int( properties, "top_field_first", !profile->progressive );
+	mlt_properties_set_int( properties, "progressive", profile->progressive );
```
ddennedy: FYI, This is already taken care of in mlt_consumer.c apply_profile_properties(), which is called inside mlt_consumer_start(), but it does not hurt anything to do it here as well.

src/modules/decklink/consumer_decklink.cpp
```diff
@@ -392,6 +392,7 @@ class DeckLinkConsumer
 			return false;
 		}
 		mlt_properties_set_int( properties, "top_field_first", m_displayMode->GetFieldDominance() == bmdUpperFieldFirst );
+		mlt_properties_set_int( properties, "progressive", m_displayMode->GetFieldDominance() == bmdProgressiveFrame );
```
ddennedy: The call to getDisplayMode() already requires the mlt_profile.progressive attribute to match the decklink's mode.

ddennedy:
My comments do not really require changes, but let me know if you prefer to submit a revised pull request or just merge this one.

max-verem:
i will update this patchset, duplicating property setting could introduce some misbehavior in future.

PR #195, opened 2017-02-19
Limited yuv422p16 support

max-verem:
This patchset implement minimal yuv422p16  support.

PR #199, opened 2017-02-20
Review please: Move colorspace convertion to standalone function

max-verem:
Colorspace convertion has 3 implementation over avformat module, first at consumer, second at producer, third at avcolorspace filter. What do you thing about factorizing this operation into the single function.

I moved color convertion function from producer into standalone code in a hope it could be used for other parts:
https://github.com/max-verem/mlt/commit/4802b3c555dc99b39ab85a44ba809e2957dd6a21

Please review and comment last commit from that pull request...

### DIFF
--- a/src/framework/mlt_frame.h
+++ b/src/framework/mlt_frame.h
@@ -150,6 +150,8 @@ extern int mlt_image_format_size( mlt_image_format format, int width, int height
 extern const char * mlt_audio_format_name( mlt_audio_format format );
 extern int mlt_audio_format_size( mlt_audio_format format, int samples, int channels );
 extern void mlt_frame_write_ppm( mlt_frame frame );
+extern int mlt_image_format_planes( mlt_image_format format, int width, int height, void* data, unsigned char *planes[4], int strides[4]);
+extern mlt_image_format mlt_image_format_id( const char * name );
 
 /** This macro scales RGB into the YUV gamut - y is scaled by 219/255 and uv by 224/255. */
 #define RGB2YUV_601_SCALED(r, g, b, y, u, v)\
--- a/src/modules/avformat/factory.c
+++ b/src/modules/avformat/factory.c
@@ -23,6 +23,7 @@
 #include <float.h>
 
 #include <framework/mlt.h>
+#include <framework/mlt_slices.h>
 
 extern mlt_consumer consumer_avformat_init( mlt_profile profile, char *file );
 extern mlt_filter filter_avcolour_space_init( void *arg );
@@ -42,6 +43,7 @@ extern mlt_filter filter_avfilter_init( mlt_profile, mlt_service_type, const cha
 #endif
 #include <libavutil/opt.h>
 
+#include "factory.h"
 
 // A static flag used to determine if avformat has been initialised
 static int avformat_initialised = 0;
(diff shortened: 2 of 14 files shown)


ddennedy:
You should have done the colorspace refactoring in a separate branch. Sigh. It will take me a while before I can take a good look at this, and I do not approve merging it until I do.

max-verem:
refactoring based on a branch of yuv422p16 support and contains one patch only... it reuses some functionality from that branch

src/modules/avformat/factory.h
```diff
@@ -0,0 +1,42 @@
+/*
```
bmatherly: I like the idea of making these functions shared. But I would suggest creating a new file rather than adding on to factory. My suggestion would be to create a new file named common.h and common.c and put these new functions in there.

src/framework/mlt_frame.h
```diff
@@ -150,6 +150,8 @@ extern int mlt_image_format_size( mlt_image_format format, int width, int height
 extern const char * mlt_audio_format_name( mlt_audio_format format );
 extern int mlt_audio_format_size( mlt_audio_format format, int samples, int channels );
 extern void mlt_frame_write_ppm( mlt_frame frame );
+extern int mlt_image_format_planes( mlt_image_format format, int width, int height, void* data, unsigned char *planes[4], int strides[4]);
```
bmatherly: I still think this would be more useful as three functions:

```
mlt_image_get_plane_pointer()
mlt_image_get_plane_width()
mlt_image_get_plane_height()
```

But I don't feel strongly about it.

To further my point that this function is ffmpeg specific, MLT does not have any planar formats with three planes. So planes[4] could be planes[3]. But then it would not work with ffmpeg functions.

ddennedy:
This seems to obsolete pull request #195.
I have been slammed with work, but I hope to give this some attention this weekend.

src/modules/avformat/factory.h
max-verem: i can move it to new ones common.(c,h)

src/framework/mlt_frame.h
max-verem: your approach make misunderstanding about width and stride, moreover, simple mlt_image_format_planes would be replaced with N calls: mlt_image_get_plane_count, mlt_image_get_plane_pointer(0..N),  mlt_image_get_plane_stride(0..N). I suggest to use ffmpeg approach that was checked for years...

max-verem:
> This seems to obsolete pull request #195.

it does not... in this request i am only asking just to review ONE commit https://github.com/mltframework/mlt/pull/199/commits/4802b3c555dc99b39ab85a44ba809e2957dd6a21 that relates to factorizing color conversion function....

src/modules/avformat/factory.h
ddennedy: agreed

src/modules/avformat/factory.c
```diff
@@ -437,3 +439,251 @@ MLT_REPOSITORY
...
+		mlt_slices_run_normal( c, sliced_h_pix_fmt_conv_proc, &ctx );
+	else
+		for ( i = 0 ; i < c; i++ )
+			sliced_h_pix_fmt_conv_proc( i, i, c, &ctx );
+
+	return dst_colorspace;
+}
+#endif
+
+int avformat_map_pixfmt_av2mlt( int f )
```
ddennedy: This function seems to serve same purpose as pick_image_format()  but it is different and not yet used. I think it should be expanded to map more of the av pixel formats to be a suitable replacement for pick_image_format(). Also, it should take enum AVPixelFormat as the argument and return mlt_image_format.

PR #201, opened 2017-02-20
Fix memory leak in frei0r module

alcinos:
This works as expected when call_back fill_param_info is not called. Not tested otherwise.


ddennedy:
Thank you for following my explanation so quickly. :-)

PR #204, opened 2017-02-23
Minor fixes 20170223

max-verem:
Please review, comment
If not all patches acceptable then cherry-pick most usefull

### DIFF
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -852,8 +852,10 @@ static void *consumer_read_ahead_thread( void *arg )
 		pthread_cond_broadcast( &priv->queue_cond );
 		pthread_mutex_unlock( &priv->queue_mutex );
 
+		mlt_log_timings_begin();
 		// Get the next frame
 		frame = mlt_consumer_get_frame( self );
+		mlt_log_timings_end(NULL, "mlt_consumer_get_frame");
 
 		// If there's no frame, we're probably stopped...
 		if ( frame == NULL )
@@ -894,7 +896,9 @@ static void *consumer_read_ahead_thread( void *arg )
 
 				// Get the image
 				mlt_events_fire( MLT_CONSUMER_PROPERTIES( self ), "consumer-frame-render", frame, NULL );
+				mlt_log_timings_begin();
 				mlt_frame_get_image( frame, &image, &priv->image_format, &width, &height, 0 );
+				mlt_log_timings_end(NULL, "mlt_frame_get_image");
 			}
 
 			// Indicate the rendered image is available.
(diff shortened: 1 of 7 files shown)


src/framework/mlt_consumer.c
```diff
@@ -852,8 +852,10 @@ static void *consumer_read_ahead_thread( void *arg )
 		pthread_cond_broadcast( &priv->queue_cond );
 		pthread_mutex_unlock( &priv->queue_mutex );
 
+		mlt_log_timings_begin();
 		// Get the next frame
 		frame = mlt_consumer_get_frame( self );
+		mlt_log_timings_end(NULL, "mlt_consumer_get_frame");
```
bmatherly: Should the service be provided instead of NULL?

The style in this file is to have space after (

`mlt_log_timings_end( self, "mlt_consumer_get_frame" );`

Same comment for the next two instances.

src/modules/avformat/filter_avdeinterlace.c
```diff
@@ -313,8 +314,9 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
 
 		// Fill the picture
 		avpicture_fill( output, *image, AV_PIX_FMT_YUYV422, *width, *height );
+		mlt_log_timings_begin();
 		mlt_avpicture_deinterlace( output, output, AV_PIX_FMT_YUYV422, *width, *height );
-
+		mlt_log_timings_end(NULL, "mlt_avpicture_deinterlace");
```
bmatherly: whitespace within ( )

src/modules/core/filter_fieldorder.c
```diff
@@ -97,6 +99,8 @@ static int get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format
 			// Set the normalised field order
 			mlt_properties_set_int( properties, "top_field_first", tff );
 			mlt_properties_set_int( properties, "meta.top_field_first", tff );
+
+			mlt_log_timings_end(NULL, "shifting_fields");
```
bmatherly: whitespace in ( )

src/modules/core/transition_composite.c
```diff
@@ -1398,10 +1400,12 @@ static int transition_get_image( mlt_frame a_frame, uint8_t **image, mlt_image_f
...
 
 				// Composite the b_frame on the a_frame
+				mlt_log_timings_begin();
 				if ( invert )
-					composite_yuv( *image, width_b, height_b, image_b, *width, *height, alpha_a, alpha_b, result, progressive ? -1 : field, luma_bitmap, luma_softness, line_fn, sliced );
+					composite_yuv( *image, width_b, height_b, image_b, *width, *height, alpha_a, alpha_b, result, field_id, luma_bitmap, luma_softness, line_fn, sliced );
 				else
-					composite_yuv( *image, *width, *height, image_b, width_b, height_b, alpha_b, alpha_a, result, progressive ? -1 : field, luma_bitmap, luma_softness, line_fn, sliced );
+					composite_yuv( *image, *width, *height, image_b, width_b, height_b, alpha_b, alpha_a, result, field_id, luma_bitmap, luma_softness, line_fn, sliced );
+				mlt_log_timings_end(NULL, "composite_yuv");
```
bmatherly: `mlt_log_timings_end( self, "composite_yuv" );`

ddennedy:
I am not concerned with the white space differences. I am fine with people using "a more(common)" style. However, I strongly dislike when there is no spacing in (statements+expressions-that->do*not/include->spaces+and){cram} everything together!

src/framework/mlt_consumer.c
ddennedy: > Should the service be provided instead of NULL?

In general, yes, but I suppose it can create some reading clutter. Before I merge, I just want to get @max-verem feedback about using mlt_service instead of NULL in calls to mlt_log_timings(). Otherwise, I am fine with the rest of the pull request.

max-verem: I put it NULL because it useless for analyzing and make output less readable...

PR #208, opened 2017-03-01
Colorspace factorize #2

max-verem:
Please review.

NB: I noticed strange chroma loss (first 16 pixels of image) if mlt_image_yuv422p16 used, but could be fixed aligning planes strides to 32 (https://github.com/mltframework/mlt/commit/41d0a0a881b578e8b366da290b324c6e4ec1778a)

### DIFF
--- a/src/framework/mlt_frame.c
+++ b/src/framework/mlt_frame.c
@@ -437,6 +437,8 @@ mlt_image_format mlt_image_format_id( const char * name )
 	return mlt_image_invalid;
 }
 
+#define ALIGN_TO( V, L ) ( ( (V + L - 1) / L ) * L )
+
 /** Get the number of bytes needed for an image.
   *
   * \public \memberof mlt_frame_s
@@ -470,7 +472,7 @@ int mlt_image_format_size( mlt_image_format format, int width, int height, int *
 			return 4;
 		case mlt_image_yuv422p16:
 			if ( bpp ) *bpp = 0;
-			return 4 * height * width ;
+			return height * ALIGN_TO( width * 2, 32 ) + 2 * height * ALIGN_TO( width, 32 ) ;
 		default:
 			if ( bpp ) *bpp = 0;
 			return 0;
@@ -1184,9 +1186,9 @@ int mlt_image_format_planes( mlt_image_format format, int width, int height, voi
 {
 	if ( mlt_image_yuv422p16 == format )
 	{
-		strides[0] = width * 2;
-		strides[1] = width;
-		strides[2] = width;
+		strides[0] = ALIGN_TO( width * 2, 32 );
+		strides[1] = ALIGN_TO( width, 32 );
+		strides[2] = ALIGN_TO( width, 32 );
 		strides[3] = 0;
 
 		planes[0] = (unsigned char*)data;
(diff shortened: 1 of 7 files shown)


src/modules/avformat/consumer_avformat.c
```diff
@@ -1768,6 +1753,7 @@ static void *consumer_thread( void *arg )
 
 						mlt_image_format_planes( img_fmt, width, height, image, video_avframe.data, video_avframe.linesize );
 
+#if LIBSWSCALE_VERSION_INT < AV_VERSION_INT( 3, 1, 101 )
 						// Do the colour space conversion
 						int flags = SWS_BICUBIC;
 						struct SwsContext *context = sws_getContext( width, height, pick_pix_fmt( img_fmt ),
```
ddennedy: pick_pix_fmt() was removed in this file.

src/modules/avformat/consumer_avformat.c
```diff
@@ -1776,6 +1762,22 @@ static void *consumer_thread( void *arg )
 							converted_avframe->data, converted_avframe->linesize);
 						sws_freeContext( context );
 
+#else
+						video_avframe.interlaced_frame = !mlt_properties_get_int( frame_properties, "progressive" );
+						video_avframe.format = avformat_map_pixfmt_mlt2av( img_fmt );
```
ddennedy: Behavior of avformat_map_pixfmt_mlt2av() differs slightly than pick_pix_fmt(), but it might be OK. I do not think anyone tries to use mlt_image_opengl or mlt_image_glsl_texture with the avformat consumer,

src/framework/mlt_frame.c
```diff
@@ -470,7 +472,7 @@ int mlt_image_format_size( mlt_image_format format, int width, int height, int *
 			return 4;
 		case mlt_image_yuv422p16:
 			if ( bpp ) *bpp = 0;
-			return 4 * height * width ;
+			return height * ALIGN_TO( width * 2, 32 ) + 2 * height * ALIGN_TO( width, 32 ) ;
```
ddennedy: This raises again the question about whether to add row alignment to MLT images. It is not necessarily a bad idea, but it needs to be indicated on the frame with a property (e.g. image_alignment) and the new property factored into these calculations. And then it must be done for all image formats to keep things consistent. Next question is: does all existing image processing code for yuv422, rgb24, and rgb24a need to be adjusted to deal with images whose rows are not aligned? On the other hand, we could just add a note to the documentation comment to mlt_image_yuv422p16 that says all planes are aligned to 32 bytes and strides a multiple of 32.

bmatherly: How about if we declare that yuv422p16 shall be aligned and all other (previous) formats are not? I know it would not be consistent, but it might be a way to transition to a new method going forward.

Also, in the future, someone could add a format like rgb24_aligned, so that the exiting formats could eventually be aligned as well.

src/modules/avformat/consumer_avformat.c
max-verem: yes. i missed that line to replace.

src/framework/mlt_frame.c
max-verem: Making all formats aligned will broke definitely existing pixel operation. New *yuv422p16* has no serious pixel processing routines except field-shift that works with strides properly...

ddennedy: Document the alignment requirements for mlt_image_yuv422p16.

PR #211, opened 2017-03-05
extend filter_dynamictext with an origin frame and more formatting operations

timon37:
Currently dynamic text is extremely limited, I hacked it up to support
more use-cases, mainly:
 - having an in-clip stopwatch (the docs mention this should work but
it didn't for me with 16.12.2)
 - negative countdowns (e.g. rocket launch)

One other case I think would be very useful, but I didn't take a look
at (I don't need it atm):
 - scaled time for slow-motion video or time lapse (then you should
also add year-month-day)
 - stop-frame after/before which the counter is paused (e.g. to keep it visible a few second before starting a countdown, or to keep the elapsed time visible after an event).

I just added an origin-frame property (that's the frame at which the time is 0).
The formatting is split into separate keywords for hours, minutes,
seconds, and milliseconds.
These can be chained, so if you set "#time_m#:#time_s#" the seconds
will never go beyond 59 (since minutes count that), but if you only
specify #time_s# the seconds will count arbitrarily large. The main
limitation is that the keywords have to be in most-significant-first
order for this to work.

There's a few issues with the patches though:
1. using an origin-time might be better than an origin-frame since it
doesn't depend on the framerate
2. how is compatibility/versioning handled between mlt and other
programs including kdenlive
3. getting the profile_fps as part of substitute_keywords is probably not nice?
4. uhh all those strncats, strlens... are quite ugly and inefficient (though that part's irrelevant)

I generally would prefer if somebody took this over, extended, and polished it, since I don't really want to  become a contributor.
I don't need any credit, and I can relinquish all copyright claims:)

### DIFF
--- a/src/modules/plus/filter_dynamictext.c
+++ b/src/modules/plus/filter_dynamictext.c
@@ -167,17 +167,59 @@ static void get_resource_str( mlt_filter filter, mlt_frame frame, char* text )
 
 /** Perform substitution for keywords that are enclosed in "# #".
 */
-static void substitute_keywords(mlt_filter filter, char* result, char* value, mlt_frame frame)
+static void substitute_keywords(mlt_filter filter, char* result, char* value, mlt_frame frame, mlt_position origin)
 {
 	char keyword[MAX_TEXT_LEN] = "";
 	int pos = 0;
 	int is_keyword = 0;
-
+	
+	mlt_position frames = mlt_frame_get_position( frame ) - origin;
+	int frames_negative = frames < 0 ? 1 : 0;
+	frames = abs( frames );
+	
+	double fps = 0;
+	{
+		mlt_profile profile = mlt_properties_get_data( MLT_FILTER_PROPERTIES( filter ), "_profile", NULL );
+		if ( profile )
+			fps = mlt_profile_fps( profile );
+	}
+	
 	while ( get_next_token(value, &pos, keyword, &is_keyword) )
 	{
+		size_t len = strlen( result );
+		size_t left = MAX_TEXT_LEN - len;
 		if(!is_keyword)
 		{
-			strncat( result, keyword, MAX_TEXT_LEN - strlen( result ) - 1 );
+			strncat( result, keyword, left - 1 );
+		}
+		else if ( !strcmp( keyword, "time_sign" ) )
+		{
+			if ( frames_negative )
+				strncat( result, "-", left - 1 );
+		}
+		else if ( !strcmp( keyword, "time_h" ) )
+		{
+			int num = frames / ( fps * 60 * 60 );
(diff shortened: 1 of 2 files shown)


bmatherly:
Before I consider this patch, have you looked at the count producer?
https://www.mltframework.org/plugins/ProducerCount/

The count producer supports counting up and down from arbitrary values. And it can be set to have a transparent background and overlayed on a track with a transition.

timon37:
Hmm, no I was using kdenlive and only found documentation about dynamic_text, I'll take a look if they integrate the count producer and if fits better for the use cases.

timon37:
I finally found a second to look into it, not sure whether this is a kdenlive limitation, but that plugin is used to generate clips, which are then used like any other video-clips.
The font/colors are not customizable. The background can't be made transparent.
It also requires scaling/panning with an effect to position it appropriately (instead of directly in the filter settings).

Too me it seems too cumbersome to be practical for the given use-cases, but I've got very limited knowledge about how either mlt or kdenlive operate and what the exact plan is (for all I know, it might make more sense to improve the generators in kdenlive and add more settings to the count-producer).

So frankly up to you or somebody from kdenlive to do whatever you feel is right, I solved my problem locally so I'm happy. If you don't want/need the code but feel the use-cases are valuable I can invest some time to document them better (if you've got a place for use-cases).

bmatherly:
Thanks again for your contribution. I am still inclined to say that these features would be better added to the count producer. The purpose of dynamic text filter is to overlay frame properties that can't be accessed from another track. The purpose of the count producer is to count up and down from arbitrary values. The features you desire (color, size, position, transparency) could all be added to the count producer.

However, I am willing to entertain a patch to add these features to dynamic text with the following modifications:
1) Replace "time_h", "time_m", "time_s", and "time_ms" with a single keyword "time" (or maybe "frametime") that also takes a date format string just like filedate and localtime.
2) Rename "originframe" to "offset"
3) Replace "timesign" with "direction" - modeled after the count producer:
https://github.com/mltframework/mlt/blob/master/src/modules/plus/producer_count.yml#L17
When the direction is "down", simply subtract the position from the duration so it counts from duration to zero.

bmatherly:
@timon37  Do you have any interest in resolving my previous comments? There have been some requests from Shotcut users for stopwatch features. So these changes could be of interest.

timon37:
Unfortunately I'm really busy at the moment, and it looks unlikely to change at the very least within this month probably quite a few more:(

bmatherly:
Now that we have the [timer filter](https://github.com/mltframework/mlt/commit/53215aaa5ab25f2bf7b8ca2ebcae4daaa98da843), I am going to close this request. Additional features could be added to the timer filter. But the dynamic text filter is not a good place for this feature.

Issue #217, opened 2017-03-17
linux melt ndi producer works with alpha channel NDI feed however BLUE/RED channel swapped


walterav1984:
Misbehavior:
When feeding linux melt ndi producer a "alpha channel" containing NDI feed from CasparCG 2.07,  the transparency works and is overlayed on a melt track :+1: , but the "blue" and "red" colors from the fill are swapped.

Expected behavior:
Feeding the same CasparCG NDI feed to "Open Broadcaster" with "obs-ndi" @Palakis also shows transparency and correct fill colors "blue" is blue and "red" is red.

Requirements:
melt build #215 with Newtek NDI SDK
CasparCG tricaster airsend plugin (updated to NDI compatible  is Windows only)
(Win7prof64, PAL-wide, tricaster plugin: NewTek Network Video Send.exe + NewTek NDI AirSend Updater.exe)

Commands:
```
melt -profile dv_pal_wide somesdpalwidefootage.dv  -track "ndi:CASPARCG-SERVER (1)" -transition movit.overlay -consumer qglsl 0=sdl 1=decklink

melt -profile dv_pal_wide -producer "ndi:CASPARCG-SERVER (1)" -consumer sdl
```

My guess is that is has to do with some RGBA BGRA translation?

Palakis:
Can confirm there's a pixel format mismatch in the NDI producer code : https://github.com/mltframework/mlt/blob/master/src/modules/ndi/producer_ndi.c#L265-L269

On [line 115 in producer_ndi.c](https://github.com/mltframework/mlt/blob/master/src/modules/ndi/producer_ndi.c#L115), `NDIlib_recv_color_format_e_UYVY_BGRA` should be replaced with `NDIlib_recv_color_format_e_UYVY_RGBA` if the NDI SDK used for the builds supports it (supported at least in NDI SDK v2).

walterav1984:
I tested fix #218 and can confirm the color blue/red swapping on the key "fill" is fixed.

@Palakis Thank you :+1:

PR #218, opened 2017-03-17
Fix color swap issue in NDI source receiver

Palakis:
Fixes issue #217.
And also allow RGBX frames since they have the same structure as RGBA frames, with the alpha channel forced to 255.

P.S. : Building the NDI module with these changes might require the NDI SDK v2.


walterav1984:
I can confirm this Fixes #217 for me.

Issue #216, opened 2017-03-10
linux melt console unresponsive if ndi producer has no active ndi video feed


walterav1984:
UPDATE:
Besides the NDI consumer, the NDI producer was also working if the correct syntax/arguments is used `melt -profile dv_pal_wide -producer  "ndi:SERVER-NAME (FEEDNAME)" -consumer sdl`.

Bug/Design?
By using the "wrong" NDI producer arguments or mistyping the NDI feedname and therefor not selecting a active NDI feed, the melt console "hangs/freezes" on a black screen both on SDL or Decklink. It probably never initializes and waits for the NDI producer? 
`melt -profile dv_pal_wide -producer ndi:insufficientparameterdemo1 -consumer sdl`
`melt -profile dv_pal_wide -producer "ndi:wrong-servername (feed999)" -consumer sdl`

Using -verbose or -timings won't show more debug info, melt has to be killed with `killall melt -9` since key press of 'q' or other console key bindings are not effective.

When correctly setting up a NDI producer in melt with a active NDI feed, but killing the NDI feed halfway during playback from its source site, also freezes the melt console (not responding on 'q' keypress) but the SDL/Decklink shows the current last frame from the NDI feed instead of black screen.

However reactivating the NDI feed from its source makes the melt console responsive to keypress input again. 

I tried latest github melt build of today and 1b9ad4f, with both NDI versions and compile instructions described in this issue #215.

ddennedy:
Please be aware that not many of us (including myself) use or know anything about NDI.

max-verem:
```
melt -profile dv_pal_wide -producer "ndi:DEV-5 (demo1)" -consumer ndi:demo2
```
use NDIlib_Find to find exact name, also check if avahi installed and firewalld (iptables) configured properly (or even disabled)

https://sourceforge.net/p/mlt/mailman/message/35674069/
https://sourceforge.net/p/mlt/mailman/message/35674093/

walterav1984:
It works it was indeed the syntax for the producer "ndi:SERVERNAME (FEEDNAME)". Thanks again.

Is it correct behavior that the blackscreen is unresponse to normal melt CLI interaction like key press "q" for quit or "space" play until an active NDI feed is fed?

max-verem:
you can reproduce it by simple command:
```
melt -profile dv_pal_wide -producer "ndi:DEV-5 (demo1)" -consumer null
```
[...]

PR #221, opened 2017-03-23
Make NDI producer non-blocking

max-verem:
This patch should fix blocking NDI producer during signal absence or source discovery. It should fix https://github.com/mltframework/mlt/issues/216


walterav1984:
The console remains responsize, however it produces white frames for me with alternating every x-th frame a ndi frame and that it goes back to white again. 

Probably reason you are closing the pull request.

PR #223, opened 2017-03-27
Makefiles: fix SDL detection through pkg-config; don't fail distclean if already clean

vpinon:
Hello,
I'm on a new run for Kdenlive windows builds with MXE.
1) SDL detection was failing since last time, corrected.
2) Building unreleased code from a git clone I had to run distclean, but that failed if already clean...
Thanks for the hard work!
Vincent

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,12 @@ all clean:
 	done
 
 distclean:
-	rm mlt-config packages.dat; \
+	rm -f mlt-config packages.dat
 	list='$(SUBDIRS)'; \
 	for subdir in $$list; do \
 		$(MAKE) -C $$subdir $@ || exit 1; \
-	done; \
-	rm config.mak;
+	done
+	echo > config.mak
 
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -13,12 +13,12 @@ all clean:
...
 distclean:
-	rm mlt-config packages.dat; \
+	rm -f mlt-config packages.dat
 	list='$(SUBDIRS)'; \
 	for subdir in $$list; do \
 		$(MAKE) -C $$subdir $@ || exit 1; \
-	done; \
-	rm config.mak;
+	done
+	echo > config.mak
```
ddennedy: Why are you making config.mak an empty file instead of removing it?

vpinon: config.mak is inconditionnaly included in all modules makefiles, so "make distclean" fails on this include if run on an already clean tree... but maybe this should be done before the loop!
an other solution would have been to surround all these includes by file existence checks...

PR #226, opened 2017-04-03
Add function to disconnect all producers

alcinos:


alcinos:
Which version number should I use ?

ddennedy:
MLT_6.6.0 and MLTPP_6.6.0 (these are the next release versions)

ddennedy:
I approve now.

alcinos:
I've changed logic a bit because the disconnection of the consumer done in disconnect_producer was an unwanted side-effect

PR #245, opened 2017-07-21
sdl consumer: Fix potential deadlock

chouquette:
If the thread might be waiting on the condition variable, signal it
first, and then join

### DIFF
--- a/src/modules/sdl/consumer_sdl.c
+++ b/src/modules/sdl/consumer_sdl.c
@@ -281,10 +281,6 @@ int consumer_stop( mlt_consumer parent )
 		// Kill the thread and clean up
 		self->joined = 1;
 		self->running = 0;
-#ifndef _WIN32
-		if ( self->thread )
-#endif
-			pthread_join( self->thread, NULL );
 
 		// internal cleanup
 		if ( self->sdl_overlay != NULL )
@@ -305,6 +301,10 @@ int consumer_stop( mlt_consumer parent )
 			SDL_Quit( );
 			pthread_mutex_unlock( &mlt_sdl_mutex );
 		}
+#ifndef _WIN32
+		if ( self->thread )
+#endif
+			pthread_join( self->thread, NULL );
 	}
 
 	return 0;


ddennedy:
I am going to revert this per #291 because it is causing Flowblade to crash frequently. How did you discover this? What app does it affect?

ddennedy:
Oh, I see you are working on VLMC. What do you mean by "potential?" There is some race condition that makes the deadlock intermittent? We need to find a different fix. Your change is moving something outside the protection of a mutex lock. You mention "signal", what is that signal? I do not see it. SQL_Quit()?

chouquette:
Hi,

It's a bit old to remember precisely, though from what I can see in the commit, it's a really bad idea to have 
```
pthread_mutex_lock( &self->audio_mutex );
pthread_cond_broadcast( &self->audio_cond );
pthread_mutex_unlock( &self->audio_mutex );
```
after the pthread_join.

If the joined thread is waiting for the `audio_cond` `pthread_cond_wait` to be signaled (what I meant by signal is calling `pthread_cond_signal`/`pthread_cond_broadcast`) there is no chance that the thread will be successfully join, causing the application to hand.
I suppose it did affect VLMC at some point, though I suppose it would affect any application for which `if ( !mlt_properties_get_int( MLT_CONSUMER_PROPERTIES( parent ), "audio_off" ) )` yields true.

I'm sorry I can't really comment more, this was most likely to assist our GSoC student that was working on moving from our homemade solution to MLT, I haven't worked with MLT that much myself.

ddennedy:
In a typical scenario, you are correct that it is bad behavior to not signal a wait condition prior to joining a thread. However, in this case, we cannot tell the thread to exit `self->running = 0;` and then simply proceed to do some cleanup `SDL_FreeYUVOverlay( self->sdl_overlay );` without waiting for the thread to stop. I think that was causing the crash.

I think the reason no one here is reproducing the deadlock (after many years, even though that is not foolproof), is because SDL audio is still calling sdl_fill_audio() periodically, which signals the audio_cond  wait condition. This is because we have not stopped the SDL audio subsystem. If you experience the problem again (in case SDL audio unexpectedly stops invoking callback? depending on SDL audio driver?), we should consider a change like this:
```
--- src/modules/sdl/consumer_sdl.c
+++ src/modules/sdl/consumer_sdl.c
@@ -282,6 +282,13 @@ int consumer_stop( mlt_consumer parent )
 		self->joined = 1;
 		self->running = 0;
 
+		if ( !mlt_properties_get_int( MLT_CONSUMER_PROPERTIES( parent ), "audio_off" ) )
+		{
+			pthread_mutex_lock( &self->audio_mutex );
+			pthread_cond_broadcast( &self->audio_cond );
+			pthread_mutex_unlock( &self->audio_mutex );
+		}
+
 #ifndef _WIN32
 		if ( self->thread )
 #endif
@@ -293,12 +300,7 @@ int consumer_stop( mlt_consumer parent )
 		self->sdl_overlay = NULL;
 
 		if ( !mlt_properties_get_int( MLT_CONSUMER_PROPERTIES( parent ), "audio_off" ) )
-		{
-			pthread_mutex_lock( &self->audio_mutex );
-			pthread_cond_broadcast( &self->audio_cond );
-			pthread_mutex_unlock( &self->audio_mutex );
 			SDL_QuitSubSystem( SDL_INIT_AUDIO );
-		}
 
 		if ( mlt_properties_get_int( MLT_CONSUMER_PROPERTIES( parent ), "sdl_started" ) == 0 )
 		{
```
Basically,
1. tell the consumer thread to exit
2. signal the audio wait condition
3. wait for the consumer thread to exit by joining it
4. close SDL 

However, I need to make a release this weekend because there is pent up demand for it, and I do not want to include a risky change at this point. @jliljebl I did some basic testing of this change in Flowblade without a problem. Can you test it please?

jliljebl:
Based on diff I applied on my tree this patch on top of the revert patch: https://github.com/jliljebl/mlt/commit/88e6dee7a98b7b008ae6756488b08bb061c34594

This one works, no crashes.

chouquette:
The change looks good to me as well :) Thanks!

PR #248, opened 2017-08-08
mlt_property.h: Replace include xlocale.h by locale.h

schnitzeltony:
xlocale.h was removed in glibc 2.26

I could test glibc case only. If it won't work for other environments, please consider this patch as ping.

### DIFF
--- a/src/framework/mlt_property.h
+++ b/src/framework/mlt_property.h
@@ -31,7 +31,7 @@
 #endif
 
 #if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD_version >= 900506)
-#include <xlocale.h>
+#include <locale.h>
 #else
 typedef char* locale_t;
 #endif


src/framework/mlt_property.h
```diff
@@ -31,7 +31,7 @@
 #endif
 
 #if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD_version >= 900506)
```
ddennedy: This #if needs to be broken up if you are addressing a glibc change. Also, I think there should be a glibc version check as well with perhaps older versions still #include xlocale.h.

ddennedy:
What does "patch as ping" mean?

schnitzeltony:
On Tue, Aug 8, 2017 at 7:24 PM, Dan Dennedy <notifications@github.com>
wrote:

> What does "patch as ping" mean?
>
Patch as ping meant: Although this patch might not be acceptable - you are
aware of this issue.

However: Travis check passed - is it done on glibc <  2.26? If so: Would it
be enough to split out glibc case only?

ddennedy:
Travis is using Ubuntu 14.04 with glibc 2.19. It must also build on Debian 8, which is also v2.19. 

> If so: Would it be enough to split out glibc case only?

Yes, I believe so,

plater:
mlt, shotcut and webvfx all have this issue in openSUSE:Factory which has recently updated to glibc-2.26 can I safely patch them to use locale.h until this issue is resolved in version updates?

src/framework/mlt_property.h
plater: #if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD_version >= 900506)
+#if __GLIBC_MINOR__ >= 26
+#include <locale.h>
+#else
 #include <xlocale.h>
+#endif
 #else
 typedef char* locale_t;
 #endif
is this ok?

schnitzeltony:
On Fri, Aug 25, 2017 at 9:35 AM, Dave Plater <notifications@github.com>
wrote:

> mlt, shotcut and webvfx all have this issue in openSUSE:Factory which has
> recently updated to glibc-2.26 can I safely patch them to use locale.h
> until this issue is resolved in version updates?
>
> OOps - by business stress I forgot this totally - and I don't have much
time to take care at the moment. Somebody to help out here?

plater:
I've made a patch
===================================================================
Index: src/framework/mlt_property.h
===================================================================
--- src/framework/mlt_property.h.orig   2016-11-16 08:53:11.000000000 +0200
+++ src/framework/mlt_property.h        2017-08-25 14:37:05.144802219 +0200
@@ -31,7 +31,11 @@
 #endif
 
 #if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD_version >= 900506)
+#if __GLIBC_MINOR__ >= 26 && !defined(__APPLE__)
+#include <locale.h>
+#else
 #include <xlocale.h>
+#endif
 #else
 typedef char* locale_t;
 #endif
It's on it's way to Factory, I'm sure FreeBSD uses a normal glibc but somebody pointed out that Apple have they're own.

ddennedy:
The last comment patch looks good to me, and I will apply it over the weekend if you do not want to bother to make a pull request. Thank you for your help.

Hoshpak:
I believe something was lost here, possibly due to github formatting. I tried this patch on my system and it would still include xlocale.h and fail to compile. While looking for the cause, I discovered that ```GLIBC_MINOR``` was not defined anywhere however ```__GLIBC_MINOR__``` was. After changing that, I was able to compile again. I think the same goes for ```APPLE``` versus ```__APPLE__``` but don't have a system here to verify it.

plater:
I never noticed that either, yes APPLE also lost it's underscores. Maybe this helps
https://build.opensuse.org/package/view_file/home:plater/libmlt/libmlt-nomorexlocale_h.patch?expand=1
or the download link
https://build.opensuse.org/source/home:plater/libmlt/libmlt-nomorexlocale_h.patch?rev=9356a09614d5e9080559752c24065dbe
I'm still new to github, I can't understand why you can't attach patches. Maybe it's better to add a .txt extension and attach it.

plater:
Got it, I need a ` on each end

ddennedy:
I just pushed a commit for this underscore issue fbf6a51

PR #253, opened 2017-09-05
producer_decklink_init: Extract resource (card) from arg

siddharudh:
Modules such as melted may pass arg with root_dir prefix, which is handled in this commit.

Initially, I thought of fixing it for all filenames having "service:resource" pattern in get_fullname function of melted_unit_commands.c. But, that may break avformat producer which requires full path for addressing resources. Also, noticed similar root_dir prefix removal being done in producer_color.c of core module, hence thought of handling it in producer_decklink.c.

My requirement is to switch between files playout and bypassing IN feed (live switching) using DeckLink cards having Full Duplex support.

SET root=/clips
UADD decklink:0

### Play a file
LOAD U0 sample.mp4
PLAY U0

### Bypass incoming feed
LOAD U0 decklink:0 0 500
PLAY U0

It's working fine with this commit, so please advise if this modification is fine and pull it if it doesn't break any other use cases.

Thanks.

### DIFF
--- a/src/modules/decklink/producer_decklink.cpp
+++ b/src/modules/decklink/producer_decklink.cpp
@@ -1,6 +1,6 @@
 /*
  * producer_decklink.c -- input from Blackmagic Design DeckLink
- * Copyright (C) 2011 Dan Dennedy <dan@dennedy.org>
+ * Copyright (C) 2011-2017 Dan Dennedy <dan@dennedy.org>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


src/modules/decklink/producer_decklink.cpp
```diff
@@ -853,7 +853,15 @@ mlt_producer producer_decklink_init( mlt_profile profile, mlt_service_type type,
 	// If allocated and initializes
 	if ( decklink && !mlt_producer_init( producer, decklink ) )
 	{
-		if ( decklink->open( arg? atoi( arg ) : 0 ) )
+		// Extract resource (card) from arg, removing path prefix, if any.
+		// (modules such as melted may pass arg with root_dir prefix)
+		char *arg_dup = strdup( arg ? arg : "" );
```
ddennedy: I do not think a strdup() is necessary, and you can default (when arg == NULL) to "0" thereby removing the strlen() and free() lines below.

PR #257, opened 2017-09-13
Correctly handle negative lift values

j-b-m:
Patch submitted by Dušan Hanuš, a Kdenlive user, allowing the lift/gamma/gain filter to correctly handle negative lift values. I submitted a similar change to Movit's version of the filter that should be merged later today by Steinar.

### DIFF
--- a/src/modules/plus/filter_lift_gamma_gain.c
+++ b/src/modules/plus/filter_lift_gamma_gain.c
@@ -67,6 +67,11 @@ static void refresh_lut( mlt_filter filter, mlt_frame frame )
 			g += glift * ( 1.0 - g );
 			b += blift * ( 1.0 - b );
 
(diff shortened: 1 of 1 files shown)


src/modules/plus/filter_lift_gamma_gain.c
```diff
@@ -67,6 +67,11 @@ static void refresh_lut( mlt_filter filter, mlt_frame frame )
 			g += glift * ( 1.0 - g );
 			b += blift * ( 1.0 - b );
 
+			// Clamp negative values
+			r = r < 0.0 ? 0.0 : r;
```
ddennedy: P.S. mlt_types.h defines CLAMP, MIN, and MAX macros if not already defined.

PR #261, opened 2017-10-04
Add metadata to NDI module

Palakis:
There's currently no metadata for the NDI producer and consumer, and there no visibility for these services in the MLT documentation. This contribution aims to address this, as it would bring the said visibility, but most importantly help users willing to use one of these services.

### DIFF
--- a/src/modules/ndi/consumer_ndi.yml
+++ b/src/modules/ndi/consumer_ndi.yml
@@ -1,0 +1,24 @@
+schema_version: 0.1
+type: consumer
+identifier: ndi
+title: NDI Output
+creator: Maksym Veremeyenko
+license: LGPLv2.1
+language: en
+tags:
+  - Video
+  - Network
+description: Audio/Video over IP output using NewTek's NDI technology
+parameters:
+  - identifier: argument
+    type: string
+    title: NDI Source name
+    description: Name used for discovery by NDI receivers on the network
+    required: yes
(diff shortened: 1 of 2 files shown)


src/modules/ndi/consumer_ndi.yml
```diff
@@ -0,0 +1,24 @@
+schema_version: 0.1
+type: consumer
+identifier: ndi
+title: NDI Output
+creator: Maksym Veremeyenko
+license: LGPLv2.1
+language: en
+tags:
+  - Video
```
ddennedy: This list should include "Audio"

src/modules/ndi/consumer_ndi.yml
```diff
@@ -0,0 +1,24 @@
+schema_version: 0.1
+type: consumer
+identifier: ndi
+title: NDI Output
```
ddennedy: There should be a version attribute in here.

src/modules/ndi/producer_ndi.yml
```diff
@@ -0,0 +1,18 @@
+schema_version: 0.1
+type: producer
+identifier: ndi
+title: NDI Input
+creator: Maksym Veremeyenko
+license: LGPLv2.1
+language: en
+tags:
+  - Video
```
ddennedy: version and "Audio" tag here as well

PR #262, opened 2017-10-04
Multistream PCM saving fix

max-verem:
I had a chance to test XDCAMHD files created by MLT against AVID AMA link.

They playing fine except last frame -MediaComposer throw a popup error message dialog: 

AudioStreamReader::ExecuteRequestSingleAsync() CONSISTENCY CHECK FAILURE !?!

After dumping these files i noticed that last audio chunk/packet in file contains an 128 or 256 samples instead of 1920.

I found issue comes from consumer_avformat. This happens because it operates with kind of wrong fifo length for storing samples frames. Moveover, drain fifo and codec not implemented for multi-stream output.

I reworked it, so please comment/reject or pull

### DIFF
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -1091,20 +1091,246 @@ static void write_transmitter( mlt_listener listener, mlt_properties owner, mlt_
 	listener( owner, service, (uint8_t*) args[0], *p_size );
 }
 
+typedef struct encode_ctx_desc
+{
+	mlt_consumer consumer;
+	int audio_outbuf_size;
+	int audio_input_nb_samples;
+	uint8_t audio_outbuf[AUDIO_BUFFER_SIZE], audio_buf_1[AUDIO_ENCODE_BUFFER_SIZE], audio_buf_2[AUDIO_ENCODE_BUFFER_SIZE];
+
+	int channels;
+	int total_channels;
+	int frequency;
+	int sample_bytes;
+
+	sample_fifo fifo;
+
+	AVFormatContext *oc;
+	AVStream *video_st;
+	AVStream *audio_st[ MAX_AUDIO_STREAMS ];
+	int64_t sample_count[ MAX_AUDIO_STREAMS ];
+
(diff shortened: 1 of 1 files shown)


ddennedy:
This is a very heavy change that will require much analysis and testing. I am not confident to merge without that effort, which may take some time.

max-verem:
Actually only last commit in this pull request fix this behaviour. But require codec draining code part rework...

And yes, it related to https://github.com/mltframework/shotcut/issues/451

ddennedy:
The visual review of the code changes looks good, but on testing against FFmpeg v3.2 with AAC I get the following error:
[aac @ 000000001a0ac400] more samples than frame size (avcodec_encode_audio2)
[consumer avformat] error with audio encode 1

src/modules/avformat/consumer_avformat.c
```diff
@@ -1131,14 +1104,23 @@ typedef struct encode_ctx_desc
...
+	int frame_length = ctx->audio_input_frame_size * ctx->channels * ctx->sample_bytes;
 
-	n = FFMIN( FFMIN( ctx->channels * ctx->audio_input_nb_samples, sample_fifo_used( ctx->fifo ) / ctx->sample_bytes ), AUDIO_ENCODE_BUFFER_SIZE );
+	// Get samples count to fetch from fifo
+	if ( sample_fifo_used( ctx->fifo ) < frame_length )
+		samples = sample_fifo_used( ctx->fifo ) / ( ctx->channels * ctx->sample_bytes );
+	else
+	{
+		int frames = FFMIN( sample_fifo_used( ctx->fifo ), AUDIO_ENCODE_BUFFER_SIZE ) / frame_length;
+		samples = frames * ctx->audio_input_frame_size;
```
ddennedy: Many codecs cannot accept more than one frame, but this multiplies the AVCodec.frame_size by `frames`, which goes into into AVFrame.nb_samples below.

ddennedy: Removing `frames *` ends up feeding PCM one sample (*channels) at-a-time, which is not efficient.

PR #263, opened 2017-10-04
NDI input handling improvements

max-verem:
Provided patches improve handling NDI input stream, avoid blocking and provide a more adequate audio packets alignment/synchronization with video frames.

Please comment/reject or merge


ddennedy:
I am fine with accepting just about any changes you make to this module without my input as it is largely yours and I have never used NDI. However, when reviewing the recently contributed metadata yaml, I noticed that the arguments to the producer and consumer are not saved to mlt_properties with a name like "resource". It should in almost all cases. If you are serializing an MLT object tree, it needs a named property to serialize properly with its argument. You should consider to make that correction.

PR #265, opened 2017-10-18
pixbuf producer csv files support

yuroller:
I updated gtk2 pixbuf producer in order to read description of slideshow from a csv file.
Format of csv file is "filename;ttl"; example:
image0.png;50
image1.png;40

image0.png will be shown for 50 frames, image1.png will be shown for 40 frames; repeat if loop=1
producer ttl property will be ignored; loop works.

### DIFF
--- a/src/modules/gtk2/producer_pixbuf.c
+++ b/src/modules/gtk2/producer_pixbuf.c
@@ -50,6 +50,7 @@ struct producer_pixbuf_s
 
 	// File name list
 	mlt_properties filenames;
+	mlt_position *outs;
 	int count;
 	int image_idx;
 	int pixbuf_idx;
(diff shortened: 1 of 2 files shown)


ddennedy:
MLT has playlist functionality that this duplicates, an pixbuf is already sufficiently complicated. I am inclined to reject it. Plead your case.

bmatherly:
I agree that playlist should fill this need. If we have a good reason to add "slideshow" capabilities to a producer, then I suggest we create a wrapper producer that encapsulates the image producer so that it can use either pixbuf or qimage.

yuroller:
Before implementing csv file support in pixbuf producer, I was generating a xml-mlt script using a hold producer for each png image, specifying a length for each image; then I inserted all hold producers in a playlist.
At some point I had to deal with more than 16000 png images (other than video tracks, overlays, audio tracks), and the memory mlt was using started to explode (more than 2-3GB).
I tried to understand what was happening and it seemed to me that for each producer an audio resampler was instantiated with its static buffer for conversions.
I decided to extend pixbuf producer functionality with csv support, because doing so I was instantiating a single producer for all the 16000 png images: the same video rendering now occupies about 500MB ram.

ddennedy:
You have made a good case. The hold producer was unnecessary, but due to automatically attached filters and pixbuf producer, you would still have high memory usage. Also, the integration of the patch is clean. So, I will accept it.

PR #270, opened 2017-11-13
Compilation fixes

vpinon:
Hello,
I had 2 build errors:
- one on ubuntu artful: glibc 2.26 seems not properly handled, and anyway glibc devs claim that xlocale.h should never had been used... if I understand correctly?
- one in mingw cross-build: stderr is not available in libebur; maybe something missing in linker command, but as all other fprintf are commented out why not this one?

### DIFF
--- a/src/framework/mlt_property.h
+++ b/src/framework/mlt_property.h
@@ -30,12 +30,10 @@
 #include <sys/param.h>
 #endif
 
-#if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD_version >= 900506)
-#  if __GLIBC_MINOR__ >= 26 && !defined(__APPLE__)
-#    include <locale.h>
-#  else
-#    include <xlocale.h>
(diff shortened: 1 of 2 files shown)


bmatherly:
I can't speak for the xlocal.h change.

But please try to dig a little deeper on the ebur128.c change. The file includes both <stdio.h> and <stdlib.h>. So it should have everything it needs. I try to keep that file in sync with upstream:
https://github.com/jiixyj/libebur128/blob/master/ebur128/ebur128.c

ddennedy:
I checked the xlocale.h change on a couple of OSes. I want to check a couple more.

vpinon:
Please forget about the ebur patch...
MinGW cross-compilation just works today.
Probably I forgot again to run a clean/distclean first...
Sorry and sorry again!

ddennedy:
I cherry-picked the xlocale.h change.

PR #279, opened 2018-01-01
Fix AAC encoding with FFMpeg git master.

bmatherly:
In the latest FFMpeg master, the AAC encoder requires that the
channel_layout field be set on the avcodec context.

This should fix the lipsync regression test failures on teamcity.

### DIFF
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -634,6 +634,7 @@ static AVStream *add_audio_stream( mlt_consumer consumer, AVFormatContext *oc, A
 		c->time_base = ( AVRational ){ 1, c->sample_rate };
 #endif
 		c->channels = channels;
+		c->channel_layout = av_get_default_channel_layout( channels );
 
 		if ( mlt_properties_get( properties, "alang" ) != NULL )
 			av_dict_set( &oc->metadata, "language", mlt_properties_get( properties, "alang" ), 0 );


bmatherly:
I am currently building against all supported ffmpeg versions for this change. Stand-by for results.

ddennedy:
Looks good to me. We already use av_get_default_channel_layout() in the producer without version check.

bmatherly:
You are right. Thanks for checking.

ddennedy:
One problem with this is that it overrides a user-supplied channel_layout specified as a property (AVOption). All we need to do is move the new line above calls to apply_properties().

ddennedy:
Can you do a quick test of this change?
```
@@ -582,12 +582,13 @@ static AVStream *add_audio_stream( mlt_consumer consumer, AVFormatContext *oc, A
 		// Establish defaults from AVOptions
 		avcodec_get_context_defaults3( c, codec );
 
 		c->codec_id = codec->id;
 		c->codec_type = AVMEDIA_TYPE_AUDIO;
 		c->sample_fmt = pick_sample_fmt( properties, codec );
+		c->channel_layout = av_get_default_channel_layout( channels );
 
 #if 0 // disabled until some audio codecs are multi-threaded
 		// Setup multi-threading
 		int thread_count = mlt_properties_get_int( properties, "threads" );
 		if ( thread_count == 0 && getenv( "MLT_AVFORMAT_THREADS" ) )
 			thread_count = atoi( getenv( "MLT_AVFORMAT_THREADS" ) );
@@ -631,13 +632,12 @@ static AVStream *add_audio_stream( mlt_consumer consumer, AVFormatContext *oc, A
 #if LIBAVFORMAT_VERSION_INT >= ((55<<16)+(44<<8)+0)
 		st->time_base = ( AVRational ){ 1, c->sample_rate };
 #else
 		c->time_base = ( AVRational ){ 1, c->sample_rate };
 #endif
 		c->channels = channels;
-		c->channel_layout = av_get_default_channel_layout( channels );
 
 		if ( mlt_properties_get( properties, "alang" ) != NULL )
 			av_dict_set( &oc->metadata, "language", mlt_properties_get( properties, "alang" ), 0 );
 	}
 	else
 	{
```

bmatherly:
Good catch. Yes, that change still works fine.

PR #291, opened 2018-01-17
Revert 'sdl consumer:Fix potential deadlock' a04e6d9c35f052567b50d28b…

jliljebl:
…94edb322bd28f6f9 to fix Flowblade crashes

The pull request reverts this commit: https://github.com/mltframework/mlt/commit/a04e6d9c35f052567b50d28b94edb322bd28f6f9

The commit makes Flowblade crash constantly and reverting it regains former steady functionality.

### DIFF
--- a/src/modules/sdl/consumer_sdl.c
+++ b/src/modules/sdl/consumer_sdl.c
@@ -282,6 +282,11 @@ int consumer_stop( mlt_consumer parent )
 		self->joined = 1;
 		self->running = 0;
 
+#ifndef _WIN32
+		if ( self->thread )
+#endif
+			pthread_join( self->thread, NULL );
+
 		// internal cleanup
 		if ( self->sdl_overlay != NULL )
 			SDL_FreeYUVOverlay( self->sdl_overlay );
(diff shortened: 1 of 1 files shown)


ddennedy:
I am trying to find a simple and reliable way to reproduce the crash in case the other developer comes back with another proposed change. I do not know flowblade very well to make it crash with the change proposed to revert. Can you suggest something?

jliljebl:
Its some kind of race and does not always happen. The code sequence that happens is that I do first consumer.stop() and then consumer.start() and then it crashes.

On my system the crash can be triggered:
1) On start up
2) Opening media and selecting "Change to File Profile"
3) Creating a new Project.

All these cause the consumer.stop()/consumer.start() sequence to happen which crashes.

Note that this is 50%-60% on my system, it is actually possible that it does not happen on your system, since it is not a deterministic bug.

ddennedy:
I have not been able to reproduce it. I also tried Shotcut where I know some points where it restarts the consumer. It might be a difference between our versions of PulseAudio (I have 10.0 running on Ubuntu 17.04) or something else. However, I will take your word for it.

PR #294, opened 2018-01-29
Fix mulit-channel audio.

bmatherly:
At some framerates, audio with > 6 channels would result in an audio
buffer larger than that allocated for SDL. This change will copy from the
buffer in pieces if it won't all fit at once.

### DIFF
--- a/src/modules/sdl/consumer_sdl_audio.c
+++ b/src/modules/sdl/consumer_sdl_audio.c
@@ -313,7 +313,6 @@ static int consumer_play_audio( consumer_sdl self, mlt_frame frame, int init_aud
 	int samples = mlt_sample_calculator( mlt_properties_get_double( self->properties, "fps" ), frequency, counter++ );
 	
 	int16_t *pcm;
-	int bytes;
 
 	mlt_frame_get_audio( frame, (void**) &pcm, &afmt, &frequency, &channels, &samples );
 	*duration = ( ( samples * 1000 ) / frequency );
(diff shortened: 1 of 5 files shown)


src/modules/sdl/consumer_sdl_audio.c
```diff
@@ -356,19 +355,33 @@ static int consumer_play_audio( consumer_sdl self, mlt_frame frame, int init_aud
...
+		while ( bytes_copied < bytes_to_copy )
 		{
-			if ( scrub || mlt_properties_get_double( properties, "_speed" ) == 1 )
-				memcpy( &self->audio_buffer[ self->audio_avail ], pcm, bytes );
-			else
-				memset( &self->audio_buffer[ self->audio_avail ], 0, bytes );
-			self->audio_avail += bytes;
+			int bytes = bytes_to_copy - bytes_copied;
+			int buffer_space = sizeof( self->audio_buffer) - self->audio_avail;
+			while ( buffer_space == 0 )
```
ddennedy: I think we should keep a check on self->running here and the `while ( bytes_copied < bytes_to_copy )` above.

src/modules/sdl/consumer_sdl_audio.c
```diff
@@ -356,19 +355,33 @@ static int consumer_play_audio( consumer_sdl self, mlt_frame frame, int init_aud
 	if ( init_audio == 0 )
 	{
 		mlt_properties properties = MLT_FRAME_PROPERTIES( frame );
-		bytes = ( samples * channels * 2 );
+		int bytes_copied = 0;
+		int bytes_to_copy = ( samples * channels * sizeof( int16_t ) );
```
ddennedy: sizeof( *pcm ) is preferable over a named type, but that is a nit. Only change perhaps if making other changes.

ddennedy:
consumer_play_audio() in consumer_sdl.c and consumer_sdl2.c also need this change.

ddennedy:
Hey, I am learning that SDL has weak support for multichannel audio. On Windows, I get this with sdl2 module:
`[consumer sdl2_audio] SDL failed to open audio: DirectSound CreateSoundBuffer: Invalid parameter`
Changing the SDL audio driver from DirectSound to XAudio2 requires 32-bit float audio and downgrades my request channel count from 6 to 2.

I get this with rtaudio:
`[consumer rtaudio] RtApiDs::probeDeviceOpen: DirectSound does not support more than 2 channels per device.`

Searching the web for "SDL windows surround" gives few relevant results but only one that shows some patch adding support for ALSA back in 2007. There is also talk about OpenAL and DirectSound3D. So, this is turning into a big job! VLC is also not working here on Windows for multichannel. I even attempted advanced, explicit configuration. PotPlayer and Windows 10 "Movies & TV" apps work, but they are both written specifically for Windows.

bmatherly:
Thanks for testing this in Windows. I guess that perhaps the world is not yet ready for surround sound support in Shotcut. 

I'm going to go ahead and make the changes you requested for this review because they could fix other issues as well (this problem would also occur with very low frame rates).

But maybe we should sit on the Shotcut changes for surround sound. I think that it would be best for Shotcut to have the same features on all platforms.

bmatherly:
When you use VLC, do you actually get surround sound out (4 or 6 channels), or does it just mix the surround to stereo?

bmatherly:
I think I resolved all the comments in the second commit. I made sure that all three implementations match. I'd be happy to squash the commits if you prefer.

ddennedy:
At first try, I was having a problem with VLC downmixing everying to stereo, but I figured out some things, and now it does surround.

ddennedy:
> I guess that perhaps the world is not yet ready for surround sound support in Shotcut.

Correct. I want to revisit this later when I can look into extending SDL or RtAudio to support our needs, or possibly add a new consumer (qt, portaudio?). At the moment, I just have too many WIP on Shotcut (Qt 5.9, OpenGL rework to use Movit master, keyframes) that I do not want to address this now.

bmatherly:
> Change the default for sdl and sdl2 new scrub_audio properties to 1

Good catch. Done.

PR #295, opened 2018-01-29
Extend audio channel conversion.

bmatherly:
* Add more comprehensive surround mixing combinations.
* Properly handle conversions to/from mono and stereo.
* Provide two conversion modes:
  - mix: for stereo/surround audio channels
  - dupdrop: for discrete (unrelated) audio channels
* Allow the consumer to specify the conversion mode to use.
* The default mode is "mix"

### DIFF
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -712,8 +712,9 @@ mlt_frame mlt_consumer_get_frame( mlt_consumer self )
 		if ( test_card != NULL )
 			mlt_properties_set_data( frame_properties, "test_card_producer", test_card, 0, NULL, NULL );
 
-		// Pass along the interpolation and deinterlace options
+		// Pass along the channel convert, interpolation and deinterlace options
 		// TODO: get rid of consumer_deinterlace and use profile.progressive
+		mlt_properties_set( frame_properties, "audiochannel.convert", mlt_properties_get( properties, "channelconvert" ) );
 		mlt_properties_set( frame_properties, "rescale.interp", mlt_properties_get( properties, "rescale" ) );
 		mlt_properties_set_int( frame_properties, "consumer_deinterlace", mlt_properties_get_int( properties, "progressive" ) | mlt_properties_get_int( properties, "deinterlace" ) );
 		mlt_properties_set( frame_properties, "deinterlace_method", mlt_properties_get( properties, "deinterlace_method" ) );
--- a/src/modules/core/filter_audiochannels.c
+++ b/src/modules/core/filter_audiochannels.c
@@ -23,6 +23,388 @@
 
 #include <string.h>
 
+// Mix levels taken from ATSC A/52 assuming maximum center and surround mix levels.
+#define C_MIX 0.70710678118654752440  // 1/sqrt(2)
+#define S_MIX 0.5
+
+enum Channel
+{
+	CHAN_LF  = 0,
+	CHAN_RF  = 1,
+	CHAN_C   = 2,
+	CHAN_LFE = 3,
+	CHAN_LS  = 4,
+	CHAN_RS  = 5,
+	CHAN_LB  = 6,
+	CHAN_RB  = 7,
+	CHAN_MAX = 8,
+};
+
+enum ChannelMask
+{
+	CHAN_MASK_LF  = 1 << CHAN_LF,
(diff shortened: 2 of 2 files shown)


src/framework/mlt_consumer.c
```diff
@@ -712,8 +712,9 @@ mlt_frame mlt_consumer_get_frame( mlt_consumer self )
 		if ( test_card != NULL )
 			mlt_properties_set_data( frame_properties, "test_card_producer", test_card, 0, NULL, NULL );
 
-		// Pass along the interpolation and deinterlace options
+		// Pass along the channel convert, interpolation and deinterlace options
 		// TODO: get rid of consumer_deinterlace and use profile.progressive
+		mlt_properties_set( frame_properties, "audiochannel.convert", mlt_properties_get( properties, "channelconvert" ) );
```
bmatherly: I chose "channelconvert " as the consumer property name for specifying "mix" or "dupdrop" modes.

Any suggestions for a different name?
Should this new consumer property be documented somewhere?

src/modules/core/filter_audiochannels.c
```diff
@@ -23,216 +23,416 @@
...
+			pi += in_channels;
 		}
-		if ( *format != new_format && frame->convert_audio )
-			frame->convert_audio( frame, buffer, format, new_format );
+	}
+}
 
-		// Perform the downmix. Operate on the buffer in place to avoid realloc.
-		if ( *format == mlt_audio_s16 )
+static void dup_drop_channels( void* in_buffer, void* out_buffer, int in_channels, int out_channels, int samples, int format )
```
bmatherly: This function is extracted from the original implementation - it duplicates or drops channels to reach the requested number of channels. No modifications here - just relocated.

src/modules/core/filter_audiochannels.c
```diff
@@ -23,216 +23,416 @@
...
+
+	if ( channels_avail == *channels )
+	{
+		return error;
+	}
+
+	int size = mlt_audio_format_size( *format, *samples, *channels );
+	float *new_buffer = mlt_pool_alloc( size );
+	char *convert = mlt_properties_get( MLT_FRAME_PROPERTIES( frame ), "audiochannel.convert" );
+	if ( channels_avail > 8 || ( convert && !strcmp( convert, "dupdrop" ) ) )
```
bmatherly: Default mode is "mix". The new surround mixing will be used unless the user/application specifies "channelconvert=dupdrop" on the consumer. I chose mix as the default because that makes avcodecs that don't support codec downmix (like AAC) behave the same as avcodecs that do support codec downmix (like ac3).

src/framework/mlt_consumer.c
ddennedy: There is a convention that consumer properties propagated from the consumer to the frame begin with "consumer_" (consumer_deinterlace, consumer_tff, etc.) although I see how you are targeting the corresponding filter with your name for the frame property. That is sort of an inversion of the convention, which has been to publish properties and let whatever filter use them. If you are going to target this filter, then use the exact name "audiochannels." Since, we use the plural form "channels" everywhere else including consumer and frame properties, I prefer "channels_convert" as the base property name.

ddennedy: Hmm, do we need to propagate this property in mlt_tractor.c:producer_get_audio()?

src/modules/core/filter_audiochannels.c
ddennedy: If we publish the frame property as "consumer_channels_convert", then the avformat producer could read that and if "dupdrop" then override the request_channel_layout to prevent the need to supply that property explicitly on the producer to get desired result. Alternatively, we can always set request_channel_layout from av_get_default_channel_layout(codec_context->channels) to defer all audio channel conversion to our filter.

src/framework/mlt_consumer.c
bmatherly: I had in mind that any filter could intercept this property and use/apply it. So I don't think it has to specifically target the audiochannels filter. Or example, we could create a new filter that uses avresample and that filter should also use this property. I will change it to "consumer_channels_convert"

bmatherly:
I withdraw this pull request in favor of https://github.com/mltframework/mlt/pull/305
I believe https://github.com/mltframework/mlt/pull/305 resolves all comments posted in this review.

PR #298, opened 2018-01-30
musl libc support

jnbr:
musl has no macro, so it can not be detected straigthforward.

HAVE_STRTOD_L tells that strdod_l() is available
HAVE_LOCALE_H tells that the header locale.h is available

### DIFF
--- a/src/framework/mlt_property.h
+++ b/src/framework/mlt_property.h
@@ -30,7 +30,7 @@
 #include <sys/param.h>
 #endif
 
-#if defined(__GLIBC__) && !defined(__APPLE__)
+#if defined(__GLIBC__) && !defined(__APPLE__) || HAVE_LOCALE_H
 #  include <locale.h>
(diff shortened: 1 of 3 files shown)


bmatherly:
LGTM. But Dan is the expert on this subject. We should wait for him to review after he gets back.

src/framework/mlt_property.h
```diff
@@ -30,7 +30,7 @@
 #include <sys/param.h>
 #endif
 
-#if defined(__GLIBC__) && !defined(__APPLE__)
+#if defined(__GLIBC__) && !defined(__APPLE__) || HAVE_LOCALE_H
```
ddennedy: This should have parenthesis to clarify the order of logical operations. && is higher precedence than ||:
`#if (defined(__GLIBC__) && !defined(__APPLE__)) || HAVE_LOCALE_H`

PR #305, opened 2018-02-19
Add filter_swresample.

bmatherly:
While resolving comments for https://github.com/mltframework/mlt/pull/295, I realized that I was reinventing a lot of functionality that swresample already does very well. So I decided to take a different approach.

This adds a new audio normalize filter that is aware of the channel configuration and can convert between configurations. Since it uses swresample, it also has the side-effect of converting sample formats and sample frequencies.

In general, it works by adding a new "chan_cfg" property to frames when audio is added (currently only supported in producer_avformat. The user can specify a "chan_cfg" on the consumer and filter_swresample will attempt to convert if they do not match. If the frame or consumer "chan_cfg" property is not set, it will default to "auto" which will pick the most probable configuration for the number of channels (this matches existing functionality).

### DIFF
--- a/src/framework/mlt_frame.c
+++ b/src/framework/mlt_frame.c
@@ -1228,3 +1228,128 @@ int mlt_image_format_planes( mlt_image_format format, int width, int height, voi
 
 	return 0;
 }
+
+/** Get the short name for a channel configuration.
+ *
+ * You do not need to deallocate the returned string.
+ * \public \member of mlt_frame_s
+ * \param cfg a channel configuration enum
+ * \return a string for the name of the channel configuration
+ */
+
+const char * mlt_chan_cfg_name( mlt_chan_cfg cfg )
+{
+	switch ( cfg )
+	{
+		case mlt_chan_auto:           return "auto";
+		case mlt_chan_independent:    return "independent";
+		case mlt_chan_mono:           return "mono";
+		case mlt_chan_stereo:         return "stereo";
+		case mlt_chan_2p1:            return "2.1";
+		case mlt_chan_3p0:            return "3.0";
+		case mlt_chan_3p0_back:       return "3.0(back)";
+		case mlt_chan_3p1:            return "3.1";
+		case mlt_chan_4p0:            return "4.0";
+		case mlt_chan_4p1:            return "4.1";
+		case mlt_chan_quad_back:      return "quad(back)";
+		case mlt_chan_quad_side:      return "quad(side)";
+		case mlt_chan_5p0:            return "5.0";
+		case mlt_chan_5p1:            return "5.1";
+		case mlt_chan_5p0_back:       return "5.0(back)";
+		case mlt_chan_5p1_back:       return "5.1(back)";
+		case mlt_chan_6p0:            return "6.0";
+		case mlt_chan_6p0_front:      return "6.0(front)";
+		case mlt_chan_hexagonal:      return "hexagonal";
+		case mlt_chan_6p1:            return "6.1";
+		case mlt_chan_6p1_back:       return "6.1(back)";
+		case mlt_chan_6p1_front:      return "6.1(front)";
+		case mlt_chan_7p0:            return "7.0";
+		case mlt_chan_7p0_front:      return "7.0(front)";
+		case mlt_chan_7p1:            return "7.1";
+		case mlt_chan_7p1_wide_side:  return "7.1(wide-side)";
+		case mlt_chan_7p1_wide_back:  return "7.1(wide-back)";
+	}
+	return "invalid";
+}
+
+/** Get the id of channel configuration from short name.
+ *
+ * \public \memberof mlt_frame_s
+ * \param name the channel configuration short name
+ * \return a channel configuration
+ */
+
+mlt_chan_cfg mlt_chan_cfg_id( const char * name )
+{
+	if( name )
+	{
+		mlt_chan_cfg c;
+		for( c = mlt_chan_auto; c <= mlt_chan_7p1_wide_back; c++ )
+		{
+			const char * v = mlt_chan_cfg_name( c );
+			if( !strcmp( v, name ) )
+				return c;
+		}
+	}
+	return mlt_chan_auto;
+}
+
+/** Get the number of channels for a channel configuration.
+ *
+ * \public \memberof mlt_frame_s
+ * \param cfg a channel configuration enum
+ * \return the number of channels for the channel configuration
+ */
+
+int mlt_chan_cfg_channels( mlt_chan_cfg cfg )
+{
+	switch ( cfg )
+	{
+		case mlt_chan_auto:           return 0;
+		case mlt_chan_independent:    return 0;
+		case mlt_chan_mono:           return 1;
+		case mlt_chan_stereo:         return 2;
+		case mlt_chan_2p1:            return 3;
+		case mlt_chan_3p0:            return 3;
+		case mlt_chan_3p0_back:       return 3;
+		case mlt_chan_4p0:            return 4;
+		case mlt_chan_3p1:            return 4;
+		case mlt_chan_quad_back:      return 4;
+		case mlt_chan_quad_side:      return 4;
+		case mlt_chan_5p0:            return 5;
+		case mlt_chan_5p0_back:       return 5;
+		case mlt_chan_4p1:            return 5;
+		case mlt_chan_5p1:            return 6;
+		case mlt_chan_5p1_back:       return 6;
+		case mlt_chan_6p0:            return 6;
+		case mlt_chan_6p0_front:      return 6;
+		case mlt_chan_hexagonal:      return 6;
+		case mlt_chan_6p1:            return 7;
+		case mlt_chan_6p1_back:       return 7;
+		case mlt_chan_6p1_front:      return 7;
+		case mlt_chan_7p0:            return 7;
+		case mlt_chan_7p0_front:      return 7;
+		case mlt_chan_7p1:            return 8;
+		case mlt_chan_7p1_wide_side:  return 8;
+		case mlt_chan_7p1_wide_back:  return 8;
+	}
+	return 0;
+}
+
+/** Get a default channel configuration for a given number of channels.
+ *
+ * \public \memberof mlt_frame_s
+ * \param channels the number of channels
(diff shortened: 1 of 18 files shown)


src/framework/mlt_frame.c
```diff
@@ -1228,3 +1228,128 @@ int mlt_image_format_planes( mlt_image_format format, int width, int height, voi
...
+
+/** Get the short name for a channel configuration.
+ *
+ * You do not need to deallocate the returned string.
+ * \public \member of mlt_frame_s
+ * \param cfg a channel configuration enum
+ * \return a string for the name of the channel configuration
+ */
+
+const char * mlt_chan_cfg_name( mlt_chan_cfg cfg )
```
bmatherly: These are the strings that the user would use to specify the channel conifguration. E.g:

`melt clip.mp4 -consumer sdl channels=4 chan_cfg="quad(back)"`

Any comments on the names? We could replace the "()" with "_". For example: "quad_back".

ddennedy:
Nevermind previous comment I made about swresample not doing resampling. I see that it does and deleted that comment.

bmatherly:
All your review comments should be resolved.

I can squash these commits if you want.

One other thought: we could move all the mlt_channel_layout_* functions from mlt_frame.h to a new file called mlt_audio.h. There are pros and cons - I thought I would throw it out there.

ddennedy:
Yes, it is starting to feel like we new files in framework/ such as mlt_utils, mlt_audio, mlt_image. However, not sure we should do that here and now. I will squash while merging. Nothing else needed now except for me for some testing.

ddennedy:
It appears that not all codecs support all channel layouts. However, at ffmpeg command line with ac3, both channel layouts "4.0" and "quad" work. However, in melt with this branch I get:
```
ddennedy@MS-7593:~/Videos$ melt -verbose A06.mlt -consumer avformat:tests/test.mp4 progressive=1 s=768x576 channels=4 frequency=44100 channel_layout=4.0 acodec=ac3
no more csLADSPA plugins
+-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+
|1=-10| |2= -5| |3= -2| |4= -1| |5=  0| |6=  1| |7=  2| |8=  5| |9= 10|
+-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+
+---------------------------------------------------------------------+
|               H = back 1 minute,  L = forward 1 minute              |
|                 h = previous frame,  l = next frame                 |
|           g = start of clip, j = next clip, k = previous clip       |
|                0 = restart, q = quit, space = play                  |
+---------------------------------------------------------------------+
[libx264 @ 0x7f69d0001c60] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f69d0001c60] profile High, level 3.1
[libx264 @ 0x7f69d0001c60] 264 - core 148 r2744 b97ae06 - H.264/MPEG-4 AVC codec - Copyleft 2003-2016 - http://www.videolan.org/x264.html - options: cabac=1 ref=3 deblock=1:0:0 analyse=0x3:0x113 me=hex subme=7 psy=1 psy_rd=1.00:0.00 mixed_ref=1 me_range=16 chroma_me=1 trellis=1 8x8dct=1 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=-2 threads=12 lookahead_threads=2 sliced_threads=0 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=3 b_pyramid=2 b_adapt=1 b_bias=0 direct=1 weightb=1 open_gop=0 weightp=2 keyint=250 keyint_min=25 scenecut=40 intra_refresh=0 rc_lookahead=40 rc=crf mbtree=1 crf=23.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=1:1.00
[ac3_fixed @ 0x7f69d00049c0] Channel layout 'mono' with 1 channels does not match number of specified channels 4
consumer_avformat.c: Unable to encode audio - disabling audio output.
```
channel_layout=quad works but strangely if I do not specify the channel_layout, then it succeeds and gives me 3.1. However, ffmpeg without channel_layout gives 4.0 as a result. This is because our layout order in enum mlt_channel_layout differs from libavutil/channel_layout.c:channel_layout_map. Now, if I tell melt channel_layout=3.1, then it also fails:
```
ddennedy@MS-7593:~/Videos$ melt -verbose A06.mlt -consumer avformat:tests/test.mp4 progressive=1 s=768x576 channels=4 frequency=44100 channel_layout=3.1 acodec=ac3
no more csLADSPA plugins
+-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+
|1=-10| |2= -5| |3= -2| |4= -1| |5=  0| |6=  1| |7=  2| |8=  5| |9= 10|
+-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+ +-----+
+---------------------------------------------------------------------+
|               H = back 1 minute,  L = forward 1 minute              |
|                 h = previous frame,  l = next frame                 |
|           g = start of clip, j = next clip, k = previous clip       |
|                0 = restart, q = quit, space = play                  |
+---------------------------------------------------------------------+
[libx264 @ 0x7f5e08001c60] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f5e08001c60] profile High, level 3.1
[libx264 @ 0x7f5e08001c60] 264 - core 148 r2744 b97ae06 - H.264/MPEG-4 AVC codec - Copyleft 2003-2016 - http://www.videolan.org/x264.html - options: cabac=1 ref=3 deblock=1:0:0 analyse=0x3:0x113 me=hex subme=7 psy=1 psy_rd=1.00:0.00 mixed_ref=1 me_range=16 chroma_me=1 trellis=1 8x8dct=1 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=-2 threads=12 lookahead_threads=2 sliced_threads=0 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=3 b_pyramid=2 b_adapt=1 b_bias=0 direct=1 weightb=1 open_gop=0 weightp=2 keyint=250 keyint_min=25 scenecut=40 intra_refresh=0 rc_lookahead=40 rc=crf mbtree=1 crf=23.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=1:1.00
[ac3_fixed @ 0x7f5e080049c0] Channel layout 'stereo' with 2 channels does not match number of specified channels 4
consumer_avformat.c: Unable to encode audio - disabling audio output.
```
I think something is mismapped in consumer_avformat.c.

bmatherly:
Thanks for your thorough testing, Dan. I hadn't tested avformat consumer thoroughly after renaming the property. I guess this is the disadvantage of using the "channel_layout" name: avcodec expects that property to be specified as the AV_CH_LAYOUT_* integer value. So the MLT code interprets the string and maps it to the correct AV_CH_LAYOUT value. But then the pass_properties function passes it to avcodec where it is interpreted as an integer and overwrites the value that MLT correctly set previously. So "4.0" gets interpreted as the number "4" which maps to "mono" and "3.1" gets interpreted as "3" which maps to "stereo". I've offered a commit which works around this by skipping "channel_layout" in pass_properties(). Let me know what you think of this approach.

ddennedy:
> the pass_properties function passes it to avcodec where it is interpreted as an integer

Why is it interpreted as an integer? I am fairly sure there are other AVOptions that take constants and converts from string. The AVOptions docs say:

>  When setting options, you often have a string read directly from the
user. In such a case, simply passing it to av_opt_set() is enough. For
non-string type options, av_opt_set() will parse the string according to the
option type.

And this is what we do in apply_properties(). Is this related to seeing weird Eval warnings on some properties when using the avformat consumer? I wonder if the problem is we are not supplying the AV_OPT_SEARCH_CHILDREN flag when calling av_opt_set().

In libavutil/opt.c:av_opt_set() I see specific handling for the channel_layout type that should accept our string.

ddennedy:
I figured out that the errors like:
> [libx264 @ 0x7f1c5c001c60] [Eval @ 0x7f1c633a5c60] Undefined constant or missing '(' in 'main'
[libx264 @ 0x7f1c5c001c60] Unable to parse option value "main"

are due to supplying the option flags to the search_flags in av_opt_find(). The following fixes that but not the channel_layout:
```
@@ -444,7 +444,8 @@ static void apply_properties( void *obj, mlt_properties properties, int flags )
 	for ( i = 0; i < count; i++ )
 	{
 		const char *opt_name = mlt_properties_get_name( properties, i );
-		const AVOption *opt = av_opt_find( obj, opt_name, NULL, flags, flags );
+		int search_flags = AV_OPT_SEARCH_CHILDREN;
+		const AVOption *opt = av_opt_find( obj, opt_name, NULL, flags, search_flags );
 
 		// If option not found, see if it was prefixed with a or v (-vb)
 		if ( !opt && (
@@ -453,7 +454,7 @@ static void apply_properties( void *obj, mlt_properties properties, int flags )
 			opt = av_opt_find( obj, ++opt_name, NULL, flags, flags );
 		// Apply option if found
 		if ( opt )
-			av_opt_set( obj, opt_name, mlt_properties_get_value( properties, i), 0 );
+			av_opt_set( obj, opt_name, mlt_properties_get_value( properties, i), search_flags );
 	}
 }
 ```

ddennedy:
Aha, figured it out! The AVOption "channel_layout" declared in libavcodec/options_table.h is not declared as AV_OPT_TYPE_UINT64 instead of AV_OPT_TYPE_CHANNEL_LAYOUT, which would have let av_opt_set() do its intended conversion. There are some other options of other classes that are declared as AV_OPT_TYPE_CHANNEL_LAYOUT. I am not sure why not this one in libavcodec, and if it is intentional. I see it is still this way in their git master. Maybe it is some libav compatibility thing. Searching the libav source tree, I see indeed they do not have that type. So, I think your workaround is the only solution.

bmatherly:
> So, I think your workaround is the only solution.

Ugh. I was really hoping that there would be another way. The UINT64 bitmask is completely unworkable from a user interface perspective. You couldn't expect people to be able to provide that from a command line.

PR #306, opened 2018-02-24
Add support for more channel combinations.

bmatherly:
This is not tested. I don't have a card to test with.

@ddennedy would you be willing to test this sometime? It is not urgent.

### DIFF
--- a/src/modules/decklink/consumer_decklink.cpp
+++ b/src/modules/decklink/consumer_decklink.cpp
@@ -73,7 +73,8 @@ private:
 	BMDTimeScale                m_timescale;
 	double                      m_fps;
 	uint64_t                    m_count;
-	int                         m_channels;
+	int                         m_outChannels;
+	int                         m_inChannels;
 	bool                        m_isAudio;
 	int                         m_isKeyer;
 	IDeckLinkKeyer*             m_deckLinkKeyer;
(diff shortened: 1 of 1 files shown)


ddennedy:
OK, I will. It looks like my SDI-to-HDMI mini converter only does stereo output, but it does have DIP switches to choose a particular channel pair from the SDI.

src/modules/decklink/consumer_decklink.cpp
```diff
@@ -380,7 +381,23 @@ class DeckLinkConsumer
 		// Initialize members
 		m_count = 0;
 		preroll = preroll < PREROLL_MINIMUM ? PREROLL_MINIMUM : preroll;
-		m_channels = mlt_properties_get_int( properties, "channels" );
+		m_inChannels = mlt_properties_get_int( properties, "channels" );
+		if( m_inChannels <= 2 )
+		{
+			m_outChannels = 2;
+		}
+		else if( m_inChannels <=4 )
```
ddennedy: The docs say "only 2, 8 or 16 channel output is supported." and indeed it fails on 4 for me in testing.

BTW my SDI to HDMI mini converter does multi-channel out as my receiver displays something about the number of channels, but it has a strange output with mixing channels because it is configured for 5.1 and thinks the input is 6.1 perhaps due to some incompatible signaling.

ddennedy:
Regarding the receiver, I literally meant "6.1" not 7.1! I have never seen that before. I suspect it may due to the fact that it is a 7 channel amp (sub is pre-out), and this is some quirk or bug with it.

PR #311, opened 2018-03-15
Fix for loop initial declarations.

bmatherly:
Proposed fix for:
https://github.com/mltframework/shotcut/issues/521

I don't get that error in my environment. So I expect that this fixes it, but I don't know for sure. It must be related to gcc version? Newer versions enable C99 by default?

### DIFF
--- a/src/modules/avformat/filter_swresample.c
+++ b/src/modules/avformat/filter_swresample.c
@@ -76,7 +76,8 @@ static void audio_format_planes( mlt_audio_format format, int samples, int chann
 {
 	int plane_count = audio_plane_count( format, channels );
 	size_t plane_size = audio_plane_size( format, samples, channels );
-	for( int p = 0; p < plane_count; p++ )
+	int p = 0;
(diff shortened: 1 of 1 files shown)


src/modules/avformat/filter_swresample.c
```diff
@@ -255,7 +259,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 		audio_format_planes( in_format, in_samples, in_channels, *buffer, pdata->in_buffers );
 		audio_format_planes( out_format, alloc_samples, out_channels, out_buffer, pdata->out_buffers );
 
-		int out_samples = swr_convert( pdata->ctx, pdata->out_buffers, alloc_samples, (const uint8_t**)pdata->in_buffers, in_samples );
```
ddennedy: BTW this C99 feature of intermingling variable declarations is permitted without a compiler flag and used throughout the code, but I am not going to reject the pull request, of course.

PR #323, opened 2018-04-07
Fix rotoscoping crash

vpinon:
mlt_image_format_size surprisingly increases height by 1 line, which causes memory copy out of source range.

This is the short-term minimal fix.

After a rapid look through the code, it seems to me that this extra line is unexpected in most places where this function is called.
Do you remember what it is for? Then would should add a comment...
Don't you think it would be better to remove this height change in the function, and adjust the few places where it is necessary? This could create regressions, in MLT but also in software calling it...


ddennedy:
I am well aware of this height quirk and have not removed it yet. As I recall, there are some image processing functions that can overshoot the end of the image buffer.

PR #327, opened 2018-04-12
Fixing GCC warnings

vpinon:
Hello,
I'm in the process of cleaning the compile warnings we get by building MLT.
Some trivial changes...
the 1st commit in decklink module looks as a bug to me, but I don't have the hardware to test
the 2nd is mostly cosmetic, to avoid missing real issues

then the 3rd, 4th and 5th are targeting recent FFmpeg API.
This is a work in progress as consumer part is still unfixed, but before spending more time on it I would like to know what you think of it.
Handling "codec" to "codecpar" transition makes it difficult to support LibAv < 57.5 (2015/08): is it acceptable to drop older versions, or I should I use a macro to avoid making the switch too ugly?

### DIFF
--- a/src/framework/mlt_frame.h
+++ b/src/framework/mlt_frame.h
@@ -162,6 +162,10 @@ extern mlt_channel_layout mlt_channel_layout_default( int channels );
   y = ((263*r + 516*g + 100*b) >> 10) + 16;\
   u = ((-152*r - 300*g + 450*b) >> 10) + 128;\
   v = ((450*r - 377*g - 73*b) >> 10) + 128;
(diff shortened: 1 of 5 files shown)


vpinon:
Thinking of it: with my 5th commit before the 4th, I could reuse codec_context instead of codecpar, and the changes would be much less consequent, allowing to switch the code depending on libav version.
I'll try to get it done tomorrow.
Have a nice evening :)

vpinon:
Hello,
Skipping the FFmpeg API updates, the 2 simple commits are still interesting.
OK to merge like this?

PR #334, opened 2018-04-28
Fix mono channel configuration for ladspa plugins.

bmatherly:
Some plugins only operate on 2 channels. This change will satisfy those plugins when only one channel is requested by the consumer.

### DIFF
--- a/src/modules/jackrack/filter_ladspa.c
+++ b/src/modules/jackrack/filter_ladspa.c
@@ -71,10 +71,27 @@ static jack_rack_t* initialise_jack_rack( mlt_properties properties, int channel
 
 			if ( plugin && plugin->desc && plugin->copies == 0 )
 			{
-				mlt_log_warning( properties, "Not compatible with %d channels. Requesting %lu channels instead.\n",
-					channels, plugin->desc->channels );
-				jackrack = initialise_jack_rack( properties, plugin->desc->channels );
+				// Calculate the number of channels that will work with this plugin
+				int request_channels = plugin->desc->channels;
+				while ( request_channels < channels )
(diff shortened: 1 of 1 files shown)


src/modules/jackrack/filter_ladspa.c
```diff
@@ -71,10 +71,27 @@ static jack_rack_t* initialise_jack_rack( mlt_properties properties, int channel
 
 			if ( plugin && plugin->desc && plugin->copies == 0 )
 			{
-				mlt_log_warning( properties, "Not compatible with %d channels. Requesting %lu channels instead.\n",
-					channels, plugin->desc->channels );
-				jackrack = initialise_jack_rack( properties, plugin->desc->channels );
+				// Calculate the number of channels that will work with this plugin
+				int request_channels = plugin->desc->channels;
+				while ( request_channels < channels )
+					request_channels += plugin->desc->channels;
```
ddennedy: Is this loop supposed to be like `MAX(1, floor(channels / plugin->desc->channels)) * plugin->desc->channels`?

bmatherly: In the case where a plugin supports 2 channels and the consumer requests 5 channels, the the ideal request_channels would be 6. I think your suggestion would only result in 4 channels.

Using ceil() instead of floor() might work:
> MAX( 1, ceil( (double)channels / (double)plugin->desc->channels) ) * plugin->desc->channels;

I had played around with some math based approaches but didn't find anything that I thought was more readable than the simple loop. I'd be happy to change it to use math if you prefer.

ddennedy: ok, thanks for explanation, no change required.

PR #336, opened 2018-05-04
Fix luma transition can't use pgm file

vvck:
when use commamd param like : -transition luma:luma1.pgm,the file name will be cut at first call create_transition.


ddennedy:
Please provide a better explanation of the problem. Is create_transition() truncating the command line after the ':' ?

vvck:
yes,   I find the create_transition() is been called twice, when the first call 'id' will be cut, so the second call it will be no file name at all.  this bug is exist all transition , use like : -transition xxx:xxx .

PR #340, opened 2018-05-21
Newline

bmatherly:
Proposed fix for issue reported here:
https://forum.shotcut.org/t/text-preset-is-not-saving-more-than-one-line-of-content/2192


ddennedy:
I do not like this heavy of a change especially as mlt_properies_load() is deprecated (although unlikely to be removed). Some options:

1. add unit tests in src/tests/test_properties
2. convert Shotcut to use the YAML APIs
3. encode and decode the newline in Shotcut

If option 1, then the change should follow the Java (and other scripting langs) convention of a backslash at end of line to continue to next line.

bmatherly:
Closed in favor of https://github.com/mltframework/shotcut/pull/565

Issue #339, opened 2018-05-19
crash: mlt_animation_next_key


eszlari:
> Kdenlive Crashes when attempting to use "White Balance" color correction

https://github.com/flathub/org.kde.kdenlive/issues/10
```
#0  0x00007ffff71885f1 in mlt_animation_next_key (self=0x0, item=item@entry=0x7fffffffbf90, position=-1) at mlt_animation.c:527
#1  0x00007ffff6f5286a in Mlt::Animation::next_key(int) (this=<optimized out>, position=<optimized out>) at MltAnimation.cpp:109
#2  0x00000000005aa192 in KeyframeView::loadKeyframes(QLocale const&, QDomElement const&, int, int) ()
[...]
```

faridosc:
I just tested 18.04.1 AppImage and it seems to work fine. Ar you using MLT 6.8 in the flatpak by any chance?

eszlari:
Yes, see the MLT commit in https://github.com/flathub/org.kde.kdenlive/blob/master/org.kde.kdenlive.json

faridosc:
AFAIK it is better to stick with MLT 6.6 for now until Kdenlive devs build it against a later version. (Just an opinion though.)

eszlari:
This was fixed in kdenlive: https://github.com/KDE/kdenlive/commit/9b09dbb2b0de78d4eaf4011c8a462fdb3c6916b6

lukefromdc:
Still getting crashes with mlt master on opening a very large and complex file. Effects include "lift-gamma-gain," "position and zoom" and many others. Can't post the file itself as some of the raw clips are sensitive along with their filenames.

PR #342, opened 2018-05-24
Revert "Fix parsing non-animation string as animation" due to crash on opening large kdenlive files

lukefromdc:
This reverts commit 5256000109836b240aa627d5159b85305d179e33.
Commit caused segfaults on opening long Kdenlive projects

### DIFF
--- a/src/framework/mlt_property.c
+++ b/src/framework/mlt_property.c
@@ -1246,7 +1246,7 @@ static void refresh_animation( mlt_property self, double fps, locale_t locale, i
 double mlt_property_anim_get_double( mlt_property self, double fps, locale_t locale, int position, int length )
 {
 	double result;
-	if ( self->animation || ( self->prop_string && strchr( self->prop_string, '=' ) ) )
+	if ( self->animation || ( ( self->types & mlt_prop_string ) && self->prop_string ) )
 	{
 		struct mlt_animation_item_s item;
 		item.property = mlt_property_init();
@@ -1281,7 +1281,7 @@ double mlt_property_anim_get_double( mlt_property self, double fps, locale_t loc
 int mlt_property_anim_get_int( mlt_property self, double fps, locale_t locale, int position, int length )
 {
 	int result;
-	if ( self->animation || ( self->prop_string && strchr( self->prop_string, '=' ) ) )
+	if ( self->animation || ( ( self->types & mlt_prop_string ) && self->prop_string ) )
 	{
 		struct mlt_animation_item_s item;
 		item.property = mlt_property_init();
@@ -1316,7 +1316,7 @@ int mlt_property_anim_get_int( mlt_property self, double fps, locale_t locale, i
 char* mlt_property_anim_get_string( mlt_property self, double fps, locale_t locale, int position, int length )
 {
 	char *result;
-	if ( self->animation || ( self->prop_string && strchr( self->prop_string, '=' ) ) )
+	if ( self->animation || ( ( self->types & mlt_prop_string ) && self->prop_string ) )
 	{
 		struct mlt_animation_item_s item;
 		item.property = mlt_property_init();
@@ -1660,7 +1660,7 @@ int mlt_property_anim_set_rect( mlt_property self, mlt_rect value, double fps, l
 mlt_rect mlt_property_anim_get_rect( mlt_property self, double fps, locale_t locale, int position, int length )
 {
 	mlt_rect result;
-	if ( self->animation || ( self->prop_string && strchr( self->prop_string, '=' ) ) )
+	if ( self->animation || ( ( self->types & mlt_prop_string ) && self->prop_string ) )
 	{
 		struct mlt_animation_item_s item;
 		item.property = mlt_property_init();


ddennedy:
Gonna need something better than that.

lukefromdc:
Closing given I don't know enough about this for "something better than that" but I will need to keep using my local branch until someone else comes up with a fix for this. Issues are not reported here, so I had no other way to even make you aware of these crashes. I deal with some long kdenlive files with hundreds of clips, and a crucial one was crashing. A bisect found that before https://github.com/mltframework/mlt/commit/5256000109836b240aa627d5159b85305d179e33 it worked and after it crashed

ddennedy:
OK, I was hoping to get more analysis on the problem and a better fix. I probably need to expand the mutex lock to include strchr(). I will try that. Something like (just one of the functions)
```
--- src/framework/mlt_property.c
+++ src/framework/mlt_property.c
@@ -1244,25 +1244,26 @@ static void refresh_animation( mlt_property self, double fps, locale_t locale, i
  */
 
 double mlt_property_anim_get_double( mlt_property self, double fps, locale_t locale, int position, int length )
 {
 	double result;
+	pthread_mutex_lock( &self->mutex );
 	if ( self->animation || ( self->prop_string && strchr( self->prop_string, '=' ) ) )
 	{
 		struct mlt_animation_item_s item;
 		item.property = mlt_property_init();
 
-		pthread_mutex_lock( &self->mutex );
 		refresh_animation( self, fps, locale, length );
 		mlt_animation_get_item( self->animation, &item, position );
 		pthread_mutex_unlock( &self->mutex );
 		result = mlt_property_get_double( item.property, fps, locale );
 
 		mlt_property_close( item.property );
 	}
 	else
 	{
+		pthread_mutex_unlock( &self->mutex );
 		result = mlt_property_get_double( self, fps, locale );
 	}
 	return result;
 }
 
```

lukefromdc:
Thanks

ddennedy:
I do not have a good test case. Much appreciated if you can test commit 2b4fdaf.

lukefromdc:
I just tried it, stil got the same crash. I have atached the results from gdb in one tarball. One file is the messages and the other the backtrace.  The key to it seems to be this:

```

Thread 1 "kdenlive" received signal SIGSEGV, Segmentation fault.
mlt_animation_next_key (self=0x0, item=item@entry=0x7fffffffb2b0, position=-1)
    at mlt_animation.c:527
527		animation_node node = self->nodes;


```
[gdb_results_and_Backtrace.tar.gz](https://github.com/mltframework/mlt/files/2037165/gdb_results_and_Backtrace.tar.gz)

lukefromdc:
Will be back tonight after something I need to do if more tests are ready then

ddennedy:
OK, that crash is probably due to a bug in kdenlive mentioned in #339.
I think the MLT change may have exposed this problem in kdenlive.

lukefromdc:
I just rebuilt kdenlive with https://github.com/KDE/kdenlive/commit/9b09dbb2b0de78d4eaf4011c8a462fdb3c6916b6 which was supposed to fix this but only mentioned box-blur(which is not in the project in question) and still got the same crash, with this in gdb again:
```
Thread 1 "kdenlive" received signal SIGSEGV, Segmentation fault.
mlt_animation_next_key (self=0x0, item=item@entry=0x7fffffffb2b0, position=-1)
    at mlt_animation.c:527
527		animation_node node = self->nodes;
(gdb) 

```
Again my temporary branch with the reverted commit still works. Master crashes, master with https://github.com/mltframework/mlt/commit/2b4fdaf61ce70e51b448337c0b8ded8eba7f0661 applied still crashes with kdenlive 18.04 after https://github.com/KDE/kdenlive/commit/9b09dbb2b0de78d4eaf4011c8a462fdb3c6916b6 .

lukefromdc:
Lots of effects in the test file, both "lift-gamma-gain" and "position and zoom" are in about the right positions for the crash

ddennedy:
The backtrace shows an obvious null pointer dereference. Applications are encourage to call the is_valid() method on the C++ wrapper objects before using, which is what the Kdenlive patch was about. The gdb backtrace does not make it clear who the caller was:
```
#2  0x000055555577d86c in  ()
#3  0x00005555556f6c63 in  ()
#4  0x0000555555704895 in  ()
#5  0x000055555578c309 in  ()
#6  0x000055555578b513 in  ()
#7  0x0000555555784a63 in  ()
#8  0x0000555555787aa3 in  ()
#9  0x0000555555783c96 in  ()
#10 0x00005555556d3669 in  ()
#11 0x00005555556d24d1 in  ()
#12 0x00005555556d186c in  ()
#13 0x0000555555b3866f in  ()
```
I added pointer checks in mlt_animation.c, which is good to have regardless the caller or whether is_valid() is used. Thank you for help so far. I hope you can give this a try as well.

lukefromdc:
That got it! File opens fine with both of these commits plus the latest kdenlive, so I will pull the current Master and build a new Debian package from it.

PR #343, opened 2018-05-30
Handle key with colon in YAML.

bmatherly:
Example: "shotcut:animIn: 68" would parse as:
 "shotcut=animIn: 68"
when it should be:
 "shotut:animIn=68"

This change will look for a colon folowed by a space as a key/value
separator.


ddennedy:
I chose to handle this the more compliant manner instead.

Issue #346, opened 2018-06-01
Updating mlt-xml.dtd


camillem:
Hi, 
First, thanks for this most useful piece of software that  is MLT!

The dtd  file seems a little bit outdated compared to the rest of the code, and it doesn't allow to validate mlt files that I think are conformant: I made simple test with Shotcut and Kdenlive and get the same kind of errors with both applications.
Most problems are missing child elements and attributes.
Would that be relevant that I submit a PR with changes to adapt the dtd to the output of Shotcut?
One trickier point is the ID "main bin" in the file, as in xml, ID can't contain spaces, AFAICT.
Do you think that this could be changed without breaking too many things?
Thanks, 
Camille

ddennedy:
> Would that be relevant that I submit a PR with changes to adapt the dtd to the output of Shotcut?

Yes, but it should not contain any "shotcut:" attributes. 

> Do you think that this could be changed without breaking too many things?

Yes, I would make Shotcut start output as "main_bin" and then make whatever code is looking for "main bin" look for either (backwards compatibility).

camillem:
Thanks for the fastest answer !

ddennedy:
I have the Shotcut change staged now for testing, but I am busy making new release. I will get to it soon.

camillem:
Thanks very much.
I have a few (like 2) additional changes but I would like to test the whole thing more thoroughly.
Would you have a pointer to a correct-but-not-so-simple mlt file that I could test the DTD on?

ddennedy:
[test.mlt.txt](https://github.com/mltframework/mlt/files/2070229/test.mlt.txt)

PR #347, opened 2018-06-01
Updating mlt-xml.dtd according to Shotcut output

camillem:
Hi, 
In relation to issue #346 .
I made one commit per change, so you can easily revert them individually.
Just let me  know if you want me to do things differently.
Thanks, 
Camille


ddennedy:
Thank you

PR #359, opened 2018-08-08
Add generic text filter.

bmatherly:
The text filter can be encapsulated by specialized text filters like
dynamictext to add additional functionality.

I've been sitting on this while I develop the timer filter (which will use filter_text). But I am tired of merging in bug fixes that you have been making. So I hope this can be merged while I continue to work on the timer filter.


bmatherly:
I can understand your caution. This change really just splits the dynamictext filter into two filters: one for displaying text and then the dynamictext filter is a wrapper that performs keyword substitution. But I suppose there is always some risk or regression. We can sit on this change for a while. Alternately, I could create the text filter without changing the dynamictext filter to use it.

PR #365, opened 2018-08-29
Add swig support for nodejs

599316527:


src/swig/Makefile
```diff
@@ -6,10 +6,10 @@ all clean:
 	for subdir in $$list; do \
 		if [ -x $$subdir/build -a ! -f .$$subdir -o $@ = clean ] ; \
 		then echo -n Building $$subdir... ; \
-			cd $$subdir && output=$$(CXXFLAGS="$(CXXFLAGS)" ./build $@ 2>&1) ; \
```
ddennedy: Great contribution! But why this change?

599316527: Putting output into a variable and echoing the variable works weird in my mac. Line-breaks are missing.

compare

And I believe that build shell's output which might be very important and helpful for those who build their own libraries to find out which step goes wrong or warning shall be always shown.

ddennedy: Same on Linux, and I noticed some useful warnings (e.g. operator=) while testing this change even when successful.

PR #373, opened 2018-10-14
Add new Spot Remover filter.

bmatherly:
This filter implements a similar capability as the avfilter.delogo filter. However, I did not port the code from delogo because I didn't really understand it. So the exact interpolation algorithm is different.

Unlike delogo, this filter supports keyframes. Also, it does not crash if the rectangle is outside of the active video.

### DIFF
--- a/src/modules/plus/filter_spot_remover.c
+++ b/src/modules/plus/filter_spot_remover.c
@@ -1,0 +1,232 @@
+/*
+ * filter_remover.cpp -- filter to interpolate pixels to cover an area
+ * Copyright (c) 2018 Meltytech, LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/modules/plus/filter_spot_remover.c
```diff
@@ -0,0 +1,232 @@
...
+{
+	int yStop = rect.y + rect.h;
+	int xStop = rect.x + rect.w;
+	int rowSize = rowCount * step;
+	for ( int y = rect.y; y < yStop; y++ )
+	{
+		uint8_t* xValueL = chan + ( y * rowSize ) + ( ( (int)rect.x - 1 ) * step );
+		uint8_t* xValueR = xValueL + ( (int)rect.w * step );
+		uint8_t* p = chan + ( y * rowSize ) + ( (int)rect.x * step );
+		float yRatio = 1.0 - ( ( y - rect.y ) / rect.h );
```
ddennedy: This is a nitpick, but I would have chosen double here especially since the literal value `1.0` is double. Add 'f' after a literal numeric to make it float.

bmatherly: Good point. Fixed it here https://github.com/mltframework/mlt/commit/8dd6f12d7dd543109288982d0a8a0b820d620c54 along with another typo:

Issue #274, opened 2017-12-09
frei0r plugins prevent mlt from multiprocessing.


plater:
I'm currently working on problem with openSUSE mlt where exporting video from kdenlive and shotcut doesn't use the same processing power as ffmpeg for the same file. I can't duplicate this but the two reporters say that this problem only occurs when frei0r plugins are present. Now the one reporter has removed all frei0r plugins listed as not thread safe in src/modules/frei0r/not_thread_safe.txt and melt works as expected using maximum cpu resources. 
What I can't understand is the reporter downloaded the stand alone shotcut and kdenlive packages and they work as expected. shotcut-linux-x86_64-171203.tar.bz2 has a full compliment of frei0r plugins and yet it works. I tried to get the reporter to build shotcut sources using build-shotcut.sh but this fails to build movit properly and fails mlt.
What is wrong with my packaging?

plater:
See https://bugzilla.novell.com/show_bug.cgi?id=1068792

ddennedy:
This bug report is difficult to follow. For starters, Shotcut (and perhaps Kdenlive appimage but not sure) are using MLT and frei0r git master when released each month. There has been a lot of changes in the past year in both projects to facilitate more concurrency, and you need both to achieve it.

For the FFmpeg integration, there was an ABI change in FFmpeg 3.2 with respect multi-threading that required a change in MLT:
https://github.com/mltframework/mlt/commit/0f8e807aa911c53e221e940556370ee983f49c64#diff-4f241487ea1a24089964f1fad57aaa89

Basically, the default in libavcodec changed from auto to 1 (but not the ffmpeg cmd line util), and MLT was not supplying a value if its corresponding property was not set by the user. However, none of that explains the frei0r interaction.
If I understand correctly, the problem is reproduced by playing a video clip with melt with frei0r installed and observing CPU usage and thread activity. And then comparing without frei0r installed.. All that not_thread_safe.txt in MLT does is introduce mutex locks to inhibit concurrency. The MLT-frei0r integration does not create threads or use sched_getaffinity(). Also, your note "I still can't understand why the frei0r-plugins are used in a simple export unless they are part of the production." is correct. They are not used.

[...]

PR #376, opened 2018-10-19
Disable OpenCV module by default

j-b-m:
As discussed, disable OpenCV by default
MLT issue #274


ddennedy:
I prefer to do this by modifying only the opencv module's configure script, which I am taking care of.

PR #377, opened 2018-10-25
Fix plug-in search directory error on Windows

vgawen:
Fix plug-in search directory error on Windows


ddennedy:
You should explain, something like: commit 5582a548 broke windows. Also, indentation is incorrect. I will fix it. Thanks

PR #379, opened 2018-11-03
Filter qtext

bmatherly:
When I apply the dynamic text filter in Shotcut on my system it causes slight lag in the preview. I set out to see if I could find an optimization. This change adds a new qtext filter which draws directly on to the video frame.

Here are some examples that show the performance change:

Generate 10000 color frames:
```
time ./shotcut/Shotcut/Shotcut.app/melt colour:red length=10000 out=10000 -consumer null terminate_on_pause=1 real_time=4
real	0m4.415s
user	0m6.723s
sys	0m0.557s
```

Generate 10000 color frames and apply the original dynamictext filter:
```
time ./shotcut/Shotcut/Shotcut.app/melt colour:red length=10000 out=10000 -filter dynamictext:hello size=500 geometry="0 0 720 480" -consumer null terminate_on_pause=1 real_time=4
real	0m20.823s
user	0m49.622s
sys	0m1.853s
```

Generate 10000 color frames and appy the new dynamictext filter using qtext:
```
time ./shotcut/Shotcut/Shotcut.app/melt colour:red length=10000 out=10000 -filter dynamictext:hello size=500 geometry="0 0 720 480" -consumer null terminate_on_pause=1 real_time=4
eal	0m7.605s
user	0m20.214s
sys	0m0.705s
```

This is obviously a large change that requires vetting before release. Also, there is no urgency. Please let me know there are any changes I should make or any tests you would like me to run.

### DIFF
--- a/src/modules/core/filter_resize.c
+++ b/src/modules/core/filter_resize.c
@@ -62,7 +62,7 @@ static uint8_t *resize_alpha( uint8_t *input, int owidth, int oheight, int iwidt
 	return output;
 }
 
-static void resize_image( uint8_t *output, int owidth, int oheight, uint8_t *input, int iwidth, int iheight, int bpp )
+static void resize_image( uint8_t *output, int owidth, int oheight, uint8_t *input, int iwidth, int iheight, int bpp, mlt_image_format format, uint8_t alpha_value )
 {
 	// Calculate strides
 	int istride = iwidth * bpp;
@@ -85,7 +85,17 @@ static void resize_image( uint8_t *output, int owidth, int oheight, uint8_t *inp
 		return;
 	}
 
-	if ( bpp == 2 )
+	if ( format == mlt_image_rgb24a )
+	{
+		while ( size -- )
+		{
+			*p ++ = 0;
+			*p ++ = 0;
+			*p ++ = 0;
+			*p ++ = alpha_value;
+		}
+	}
+	else if ( bpp == 2 )
 	{
 		while( size -- )
 		{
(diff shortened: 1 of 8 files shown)


bmatherly:
> Can you please rebase this against master.

Done

ddennedy:
It looks good in my testing so far using the dynamictext filter with qtext. The only reservation I have is that rect handling seems to be slightly different between the old filter and the new one version from this branch. I created two videos with text on them in Shotcut 18.11.18 and dev copy. Then, I put both on the Timeline on different tracks and toggle the visibility of the upper track and the rectangle is slightly larger in this branch version:
image

bmatherly:
> Then, I put both on the Timeline on different tracks and toggle the visibility of the upper track and the rectangle is slightly larger in this branch version:

Thanks for looking at this. Have you also compared the output of the dynamictext filter when using the pango producer?

I think we might be able to calibrate the new filter to more closely match the current filter by tweaking the pad and outline variables:
https://github.com/mltframework/mlt/pull/379/files#diff-9b32116d0b6a329d1c7d6d0cceb7d9e4R33

I should have some time this weekend to look into it if you don't beat me to it.

ddennedy:
> Have you also compared the output of the dynamictext filter when using the pango producer?

When I compare this branch dynamictext with the pango producer, I get no outlines but the same size. IOW, hiding the upper track visibility (current dynamictext with qtext on bottom) does not make the text appear to get smaller when hidden. Is that clear?

ddennedy:
> When I zoom in using the current text filter, I see a thin black line along the left edge

I see that! Last night I thought to myself perhaps this issue is actually an artifact of the current version's embedded compositor. I am going to accept this now.

bmatherly:
> I am going to accept this now.

I hope this can debut as part of a Shotcut Beta release. If so, beta testers should be encouraged to exercise the Text filter. I plan to make myself available to help resolve issues as they are identified.

Issue #392, opened 2018-12-25
Code issues and potential bugs


shlyakpavel:
Hello. I have some notes on the code here. Most of the issues are found by PVS studio static analyzer. I hope some experienced devs find time to check this out
1. Null pointer is freed here
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/linsys/consumer_SDIstream.c#L371
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/linsys/consumer_SDIstream.c#L381
2. Memory is used after being reallocated.
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/framework/mlt_multitrack.c#L434
3. Null pointer is freed here. I guess it should be better to refactor this code so it does not use `goto`
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/framework/mlt_playlist.c#L100
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/framework/mlt_playlist.c#L106-L107
4. `frame` is first used, then checked against null
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/framework/mlt_consumer.c#L805-L807
5. `decklink` is never null as the memory is allocated with 'new' operator. It can throw exception, but not return null https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/decklink/consumer_decklink.cpp#L988
6. Condition is obviously always true as we have already returned from the function otherwise
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/motion_est/producer_slowmotion.c#L286
7. Intention here does not represent the actual code behavior
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/plusgpl/cJSON.c#L166
In fact, `ptr2+=len;` is always executed, no matter what the condition result is.
8. This function does a real mess. It is dangerous to print a string into itself.
https://github.com/mltframework/mlt/blob/fafaa42e72cee6efb093d959829c756d52e067db/src/modules/plusgpl/filter_telecide.c#L122
Refer https://www.viva64.com/en/w/v541/
[...]

PR #395, opened 2019-01-02
Linsys: do not free null pointer

shlyakpavel:
This seems to be a dangerous copy-pasted line.
Refer issue https://github.com/mltframework/mlt/issues/392 part 1

### DIFF
--- a/src/modules/linsys/consumer_SDIstream.c
+++ b/src/modules/linsys/consumer_SDIstream.c
@@ -378,7 +378,6 @@ static void *consumer_thread(void *arg) {
 		}
 	} else if (this->device_file_video &&
 			strstr(this->device_file_video, "sdivideotx")) {
-		free(this->device_file_audio);
 		this->device_file_audio = strdup("/dev/sdiaudiotx0");
 	}
 


ddennedy:
free() accepts a null pointer. Explain more.

shlyakpavel:
> free() accepts a null pointer. Explain more.

This is dangerous in terms of misunderstanding that might occur when changing the code. The pointer is ALWAYS null so it takes much time to understand that the free() is just a copy-paste, but not intended behavior

bmatherly:
> If ptr is NULL, no operation is performed.

https://linux.die.net/man/3/free

Not a bug if the pointer was initialized to NULL. Theoretically there could be a bug if the pointer was never initialized or if it was previously freed and not re-initialized to NULL.

In the case of #392 part 1, consumer_SDIstream_s  is initialized with calloc(). So the pointer is initialized to NULL. And by inspection, I can't see a case where the pointer would be freed twice.

ddennedy:
OK, reading this again in the context of the issue and full source, there is a simpler and more concrete explanation: the else block containing this line is only entered when the pointer is null; so the free() line does nothing. It's just that comments like "seems to be copy-paste" or "leads to misunderstanding" are not as concrete.

PR #407, opened 2019-02-07
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/src/swig/python/codecs.py
+++ b/src/swig/python/codecs.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # Import required modules
+from __future__ import print_function
 import mlt
 
 # Start the mlt system
@@ -21,4 +22,4 @@ codecs = mlt.Properties( c.get_data( 'vcodec' ) )
 
 # Print the list of codecs
 for i in range( 0, codecs.count()):
-	print codecs.get( i )
+	print(codecs.get( i ))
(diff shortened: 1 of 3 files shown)


jliljebl:
There is no need to this for Flowblade at this time, Python 3 port is not happening for the next 2 releases at least.

cclauss:
These changes are compatible with both Python 2 and Python 3.  They allow further experimentation towards the goals of a Python 3 port and they provide confidence to downstream projects that the port will be completed in the 328 days remaining until Python 2 end of life.

ddennedy:
This does not fix src/swig/python/build, which also uses `print`. I verified the following works.
```
diff --git a/src/swig/python/build b/src/swig/python/build
index ab69aec9..ecbc6641 100755
--- a/src/swig/python/build
+++ b/src/swig/python/build
@@ -13,7 +13,7 @@ path=`which python 2> /dev/null`
 if [ $? = 0 ]
 then
        # Change this as needed
-       export PYTHON_INCLUDE=`python -c "import sys;print \"%s/include/python%d.%d\"%(sys.prefix,sys.version_info[0],sys.version_info[1])"`
+       export PYTHON_INCLUDE=`python -c "from __future__ import print_function;import sys;print(\"%s/include/python%d.%d\"%(sys.prefix,sys.version_info[0],sys.version_info[1]))"`

        [ ! -d "$PYTHON_INCLUDE" ] && echo python development missing && exit 1
```

ddennedy:
See 7b42f20d for the change to the build file.

PR #412, opened 2019-02-16
Add createdate keyword to dynamictext

bmatherly:


ddennedy:
The problem with this is it is already possible to include any meta.attr, I doubt this attribute is consistently available and in the same format, and there is a variety of ways locales format dates. It's a gateway to a support issues.

Separately, the Shotcut text filter should expand the variable substitution UI to be a drop-down that includes all of the existing bars variables plus the meta.attr of the original producer.

bmatherly:
That is a good comment and I am on the fence about this change. Here is the story:

I have some files that for some reason are having the file mtime property modified. So when I use the text overlay with the #filedate# keyword, the date shown on the overlay is not actually the day of the recording. So I went looking for other properties.

"creation_time" is not a format specific mettadata. FFmpeg works pretty hard to find a value to populate that field with for most formats. FFmpeg always stores the value as ISO 8601. So if I just use #meta.attr.creation_time.markup# then MLT will overlay something that looks like this: "2016-01-01T22:30:00.000000Z". That's not really a very nice thing to look at.

The keyword added by this proposed change makes a best effort to determine the most likely date of origin.
1) First, it looks for the creation_date attribute. If present, it parses the ISO 8601 format and converts to the user selected format.
2) If creation_date is not found, it reverts to using the file modification time just like #filedate# does today.

Right now I'm trying to decide how to improve this workflow in Shotcut. Maybe just using this new #createdate# would be good. But it wouldn't support other future features (like sorting by creation date in the playlist). I'm also looking at other options like helping the user update the file date easily.

I just figured that since this work might be useful to someone, I might was well put it in a safe place. Plus, I might end up using it in Shotcut.

ddennedy:
Ok good to know about FFmpeg conforming this data. But I think the output format should be a property to also allow date, time, or date and time.

bmatherly:
Yes, the user can optionally specify any strftime compatible format
https://github.com/mltframework/mlt/pull/412/files#diff-387f6fa1839e3f6c3d0c333f224c213aR174
This is the same method used for #filedate#, #localfiledate# and #localtime# and is specified in the documentation here:
https://github.com/mltframework/mlt/blob/master/src/modules/plus/filter_dynamictext.yml#L33

PR #413, opened 2019-02-19
Mark frei0r IIRblur not thread safe

j-b-m:
From my tests, frei0r IIRblur is not thread safe and produces corrupted result when using threads


ddennedy:
When finding something like this, we should also look into the implementation in frei0r to see what can be done without significantly impacting performance. My brief review looks like the img buffer used in its inst struct can be removed and allocated for each call to f0r_update(). I will accept this, but consider to make a change on frei0r. If you do not, I might soon.

PR #414, opened 2019-02-21
Fix memory leak in jackrack module

alcinos:
A couple of buffers were not properly freed.


ddennedy:
The if checks should be remove. [g_free()](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-free) takes NULL. We have actually made commits to remove if tests before free and delete calls and have a coccinelle code quality rule for it.

alcinos:
Should be fixed.

PR #415, opened 2019-02-25
Fix memory leak in consumer_xml

alcinos:
This PR overloads consumer_xml close function to ensure proper memory management, and avoid the leak of the consumer pointer.


ddennedy:
Thank you

PR #417, opened 2019-02-25
add convenience constructor for Mlt::Producer

alcinos:
The motivation behind is the following. Let's assume you want to create a producer, and use the same profile as another producer that you have a pointer to. The natural way of doing so would look like this:
```
Producer *original; // obtained externally
Producer copy(*original->profile(), "color", "red");
```
However, this will create a memory leak, since the Producer::profile() function creates a Profile instance that will never be freed.

Another approach could be similar to:
```
Producer *original; // obtained externally
Profile originalProfile(original->get_profile());
Producer copy(originalProfile, "color", "red");
```
However, this solution is not possible: if the producer outlives the `originalProfile`, then, since profiles are not ref-counted, the destructor of Profile will close the profile and the producer will be in a very unstable state (accessing any profile property, directly or indirectly, will cause a segfault).

One "good" solution would be to add a reference counter to the profile objects, in a similar spirit to the properties. However, this is a substantial design change, and as a hacky solution I propose to add a constructor to profile that allows to avoid the creation of a cpp Profile. The leak-free, correct version of the snippet above would then become
```
Producer *original; // obtained externally
Producer copy(original->get_profile(), "color", "red");
```

I'm happy to discuss alternative design fixes.

### DIFF
--- a/src/mlt++/Makefile
+++ b/src/mlt++/Makefile
@@ -10,7 +10,7 @@ LIBFLAGS += -install_name $(libdir)/$(SONAME) -current_version $(version) -compa
 else ifeq ($(targetos), MinGW)
 NAME = libmlt++$(LIBSUF)
 TARGET = libmlt++-$(soversion)$(LIBSUF)
-CXXFLAGS += -DMLTPP_EXPORTS
+CXXFLAGS += -DMLTPP_EXPORTS --std=c++11
 LIBFLAGS += -Wl,--output-def,libmlt++.def
 else
 NAME = libmlt++$(LIBSUF)
--- a/src/mlt++/MltConsumer.cpp
+++ b/src/mlt++/MltConsumer.cpp
@@ -38,6 +38,11 @@ Consumer::Consumer( Profile& profile ) :
 }
 
 Consumer::Consumer( Profile& profile, const char *id, const char *arg ) :
+	Consumer ( profile.get_profile(), id, arg )
+{
+}
+
+Consumer::Consumer( mlt_profile profile, const char *id, const char *arg ) :
 	instance( NULL )
 {
 	if ( id == NULL || arg != NULL )
--- a/src/mlt++/MltConsumer.h
+++ b/src/mlt++/MltConsumer.h
@@ -40,6 +40,7 @@ namespace Mlt
 			Consumer( );
 			Consumer( Profile& profile );
 			Consumer( Profile& profile, const char *id , const char *service = NULL );
+			Consumer( mlt_profile profile, const char *id , const char *service = NULL );
 			Consumer( Service &consumer );
 			Consumer( Consumer &consumer );
 			Consumer( mlt_consumer consumer );
--- a/src/mlt++/MltFilter.cpp
+++ b/src/mlt++/MltFilter.cpp
@@ -25,6 +25,11 @@
 using namespace Mlt;
 
 Filter::Filter( Profile& profile, const char *id, const char *arg ) :
+	Filter( profile.get_profile(), id, arg )
+{
+}
+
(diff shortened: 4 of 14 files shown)


alcinos:
I didn't update the version file, I'm assuming I should put the function in version 6.14.0?

ddennedy:
I agree with your findings, but the intention of the current design is that the app manages the Profile(s) (typically one or a few known-purpose) and can pass it by reference when needed by a constructor. With that said, I can see how the constructor you suggest can be useful in some use cases. Do you anticipate this style of constructor is useful for other subclasses of Service?

Yes, this must be added to the mlt++.vers file. Thanks

alcinos:
You're right about propagating to other Service classes. I applied this to the ones for which it was the most likely to be useful.

ddennedy:
The code duplication bothers me, and the constructors that take Profile& should be refactored to use the new constructors by invoking profile.get_profile():
```
Producer::Producer( Profile& profile, const char *id, const char *service ) :
    Producer( profile.get_profile(), id, service )
{
}
```

alcinos:
I agree, however constructor delegation is only available since c++11 (cf https://en.cppreference.com/w/cpp/language/initializer_list). It seems that the c++ target doesn't define a standard, which makes me assume it's c++03?

ddennedy:
What about?
```
Producer::Producer( Profile& profile, const char *id, const char *service ) :
	instance( NULL ),
	parent_( NULL )
{
	Profile( profile.get_profile(), id, service );
}
```

alcinos:
This just creates a local temporary object that is immediately discarded. I believe the only correct way to do it in c++03 is to abstract init code in an external function that is called by both constructors:
```
void Producer::init( mlt_profile profile, const char *id, const char *arg ) 
{
// init code
}

Producer::Producer( Profile& profile, const char *id, const char *service ) :
	instance( NULL ),
	parent_( NULL )
{
    init( profile.get_profile(), id, service );
}


Producer::Producer( mlt_profile profile, const char *id, const char *service ) :
	instance( NULL ),
	parent_( NULL )
{
    init( profile, id, service );
}
```
In this case, the init method should probably be private.
Would you prefer that?

ddennedy:
Hmm.. yes, I was thinking about how you can invoke a base method within an overridden method. Let's go with c++11. [SWIG largely supports it](http://www.swig.org/Doc3.0/CPlusPlus11.html), which was my main concern.

alcinos:
Ok, I edited the Makefile accordingly, and used the delegating constructor everywhere. Tell me if something else needs to be addressed.

src/mlt++/Makefile
```diff
@@ -10,7 +10,7 @@ LIBFLAGS += -install_name $(libdir)/$(SONAME) -current_version $(version) -compa
 else ifeq ($(targetos), MinGW)
 NAME = libmlt++$(LIBSUF)
 TARGET = libmlt++-$(soversion)$(LIBSUF)
-CXXFLAGS += -DMLTPP_EXPORTS
+CXXFLAGS += -DMLTPP_EXPORTS --std=c++11
```
ddennedy: Travis CI failed. Here is the way we specify c++11 in src/modules/qt/Makefile. We should do it the same way:
https://github.com/mltframework/mlt/blob/master/src/modules/qt/Makefile#L46

Or, possibly put it in into the top level configure that generates the config.mak that all Makefiles include. Then, we can safely, gradually migrate all C++ code to c++11.

alcinos:
Sorry didn't have time to update this PR, but there was a typo in my Makefile modification: I added the flag in a position that is guarded by the if on l10 (is the os MinGW). Didn't notice it since it's not indented.

It would be better to add it eg. on line 22, and it would probably fix the CI. Do you want me to PR that?

PR #418, opened 2019-02-26
Add reorder function to mlt playlist.

bmatherly:
This change adds a function to allow a playlist to be arbitrarily reordered I would be open to discussion of other ways to accomplish this.

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -532,4 +532,5 @@ MLT_6.12.0 {
 MLT_6.14.0 {
   global:
     mlt_frame_get_unique_properties;
+    mlt_playlist_reorder;
 } MLT_6.12.0;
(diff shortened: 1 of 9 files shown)


ddennedy:
I feel like we should have unit tests for mlt_playlist/Mlt::Playlist. Are you up to adding at least one for this new method? It would exercise some other methods in the course of it.

bmatherly:
> I feel like we should have unit tests for mlt_playlist/Mlt::Playlist. Are you up to adding at least one for this new method? It would exercise some other methods in the course of it.

Yes, I would be happy to add some unit tests. But before I do, could you comment on the general approach? I have some concern that it might be confusing. But I couldn't think of a better way. I had considered putting the sorting in the playlist class. For example, I could add a function mlt_playlist_sort_by_filedate(). But then a new function would have to be added every time an application wants to invent a new way to rearrange the list. 

If you are comfortable with the current approach, I'll put together some tests.

ddennedy:
I like the general approach.

bmatherly:
I added some tests for playlist. Let me know what you think.

PR #428, opened 2019-03-26
Add get/set datetime to producer.

bmatherly:
"datetime" is a reserved producer property that can hold the origin
(creation) time of the producer. The datetime property is always in
UTC.

Here is an alternative method to associate time with a producer. The advantage of this method is that it can be applied to non-file producers. The disadvantage is that it requires new producer functions and a new reserved producer property (datetime).

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -533,4 +533,6 @@ MLT_6.14.0 {
   global:
     mlt_frame_get_unique_properties;
     mlt_playlist_reorder;
+    mlt_producer_set_creation_time;
+    mlt_producer_get_creation_time;
 } MLT_6.12.0;
--- a/src/framework/mlt_producer.c
+++ b/src/framework/mlt_producer.c
@@ -30,6 +30,10 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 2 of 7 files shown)


ddennedy:
I'm a little concerned about the vagueness of datetime: creation, modified, or possibly even something else like a base. Is that intentional? perhaps document the intention.

bmatherly:
Great comment. I intended for it to represent the date the file was created. I can see how that is not clear from the code. Now that I think about this some more, I think I would like to propose a slightly different approach.

When do you plan to release the next Shotcut beta? I wold not like to let the current implementation get out because it modifies files and I would not like to propagate that.

ddennedy-gpsw:
> When do you plan to release the next Shotcut beta?

I was planning for March 30.

bmatherly:
> I was planning for March 30.

In that case, I might submit a patch to comment out the "Set date time" menu option until I feel more comfortable with a long-term solution.

bmatherly:
I put some more thought into this and tried some different options. I propose a similar approach that uses a different name for the functions. I landed on "creation_time" for the function and property names. I think this is congruent with what people will expect for this property.

ddennedy:
I think creation_time works good too.

PR #431, opened 2019-04-12
fix python include detection

tpgxyz:
Fix python detection. Based on OpenMandriva patch
https://github.com/OpenMandrivaAssociation/mlt/blob/master/mlt-6.14.0-fix-python3-detect.patch

### DIFF
--- a/src/swig/python/build
+++ b/src/swig/python/build
@@ -13,7 +13,7 @@ path=`which python 2> /dev/null`
 if [ $? = 0 ]
 then
 	# Change this as needed
-	export PYTHON_INCLUDE=`python -c "import sys;print(\"{}/include/python{}.{}\".format(sys.prefix,*sys.version_info))"`
+	export PYTHON_INCLUDE=`python -c "from sysconfig import get_paths; info = get_paths(); print(info['include'])"`
 
 	[ ! -d "$PYTHON_INCLUDE" ] && echo python development missing && exit 1
 


tpgxyz:
@ddennedy Here are the results:

Original call
`[root@lazur tpg]# python -c "import sys;print(\"{}/include/python{}.{}\".format(sys.prefix,*sys.version_info))"`
/usr/include/python3.7

Patched call:
`[root@lazur tpg]# python -c "from sysconfig import get_paths; info = get_paths(); print(info['include'])"`
/usr/include/python3.7m

OpenMandriva's python is compiled with https://www.python.org/dev/peps/pep-3149 support

ddennedy:
Unfortunately, this change is making our Debian 8 nightly build fail:

http://build.mltframework.org/viewLog.html?tab=buildLog&buildTypeId=bt26&buildId=61604

> [03:47:57][Step 1/1] Building python...python development missing

I think it is time to upgrade these to Debian 9 since that is the stable version for nearly 2 years. Still, it is good to know about this. I am a little surprised since the python package version is 2.7.x

https://packages.debian.org/jessie/python-dev

ddennedy:
This change also breaks on Debian stable. `python -c "from sysconfig import get_paths; info = get_paths(); print(info['include'])"` returns `/usr/local/include/python2.7`, when it should be /usr/include obviously. Maybe this is a Debian bug, but I am going to revert it as a result. It needs to build out-of-the-box on Debian, which is higher priority than OpenMandriva, sorry to say.

tpgxyz:
On OpenMandriva it works without issues:
```

[tpg@lazur ~]$ python2 -c "from sysconfig import get_paths; info = get_paths(); print(info['include'])"
/usr/include/python2.7
```

```
[tpg@lazur ~]$ python -c "from sysconfig import get_paths; info = get_paths(); print(info['include'])"
/usr/include/python3.7m
[tpg@lazur ~]$ 
```

PR #437, opened 2019-05-01
Fix race in mlt_consumer

alcinos:
Access to the speed needs to be atomic. 
The (potential) concurrent accesses are:

- line 806 (write)
- line 1579 (read)

### DIFF
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -31,6 +31,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <sys/time.h>
+#include <stdatomic.h>
 
 /** Define this if you want an automatic deinterlace (if necessary) when the
  * consumer's producer is not running at normal speed.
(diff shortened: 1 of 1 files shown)


ddennedy:
I see this is C11, but we are currently on C99. I do not mind bumping up the requirement, but do you anticipate many people needing to add `-std=c11` to their CFLAGS? According to [this reference](https://stackoverflow.com/a/14737642/4355276) to gcc docs, it was added in gcc v5.1.0. According to distrowatch.com, even the latest version of CentOS does not have this yet. Neither does Debian 8 even though that is oldstable now. Ubuntu since 15.10.

See also __sync_val_compare_and_swap() used in mlt_consumer.c. Maybe these should be converted to [atomic_compare_exchange_strong()](https://en.cppreference.com/w/c/atomic/atomic_compare_exchange).

alcinos:
According to [the official wiki](http://gcc.gnu.org/wiki/C11Status), the part we care about (atomics) was added in gcc 4.9. This makes it compatible with Debian 8 and Ubuntu 14.10.

As you note, CentOs is stuck on 4.8.5, however gcc 4.9 has been available for more than 5 years as of now (first release April 22th 2014), I'd find it a bit lame to make the code more obscure than needed to support such ancient compilers. 

About the CFLAGS, you are right. The default standard is gnu11 since gcc 5.1 but I will make it explicit to ease compilation on slightly older compilers.

PR #438, opened 2019-05-02
Fix some potential races in consumer_sdl2_audio

alcinos:
Made the playing flag atomic, and added protection on the cleanup of audio_avail.

### DIFF
--- a/src/modules/sdl2/consumer_sdl2_audio.c
+++ b/src/modules/sdl2/consumer_sdl2_audio.c
@@ -31,6 +31,7 @@
 #include <pthread.h>
 #include <SDL.h>
 #include <sys/time.h>
+#include <stdatomic.h>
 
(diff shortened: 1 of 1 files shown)


ddennedy:
It looks like this change is needed for all the other sdl and sdl2 consumers. Were you planning to do those too, or should I after accepting this?

alcinos:
So I'm fixing the races that I actually observe in the wild using TSAN, and for which I can confirm that:
1. It's not a false positive, and there are actually situations where it can be racy
2. Adding the atomic flag is indeed a valid fix, at least from what I can see from the conditions that trigger the race. It might be the case that in other consumers the situation is more complex and stronger synchronization primitives are required.

So I'm planning to PR fixes for whatever I catch in my net, but I am not confident enough about the inner workings of these consumers to blindly fix potential races.

PR #442, opened 2019-05-10
Fix interpolation in rotoscoping filter

alcinos:
`position` should evaluate to 1 when `time==pos2` and 0 when `time==pos1`


yoda-jm:
I confirm it fixed the rendering issue I reported here: https://bugs.kde.org/show_bug.cgi?id=407418

PR #455, opened 2019-06-04
 mlt_service_identify identify the producer type

vvck:
When user mlt_parse to parse a pruducer load by xml , there is something wrong with the node of `<producer>`, beacuse of  the producer 's service object can‘t get right type using with  `mlt_service_identify()` function . 
> mlt_type : "mlt_producer" , resource : "\<producer>"

the type return  unknown_type,  the parse will break.


ddennedy:
> the parse will break.

mlt_parser.[hc]? Your own? If mlt_parser, it is rarely used; so I would not be surprised.

You can also use properties to detect things. Take a look at consumer_xml.c as it is able to parse the object graph fine. It uses property "mlt_type" that is set in mlt_factory.c.

I am not entirely opposed to this, but I worry that it will break some apps that rely on using `unknown_type` as a condition to look at other properties such as mlt_service. I will have to check the code bases of Flowblade and Kdenlive (already checked Shotcut) as they are the only ones I really care about.

vvck:
Yes,  I used mlt_parser.[hc] to parse a mlt producer loaded from xml ,but my ultimate aim is to parse all item in it,  I just identify the  tractor 、multitrack、playlist  when resource property is <tractor > <multitrack> <playlist>,but only producer type is return unkown type .
As it can clearly find the producer type use <producer> mark, I suggest it will be much better to return the producer type    or   Is  unknown_type only  be equal to producer type?

ddennedy:
I wrote unit tests that illustrates a problem:
```
/*
 * Copyright (C) 2019 Meltytech, LLC
 *
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.
 *
 * This library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 * Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
 * License along with consumer library; if not, write to the Free Software
 * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */

#include <QtTest>
#include <mlt++/Mlt.h>
using namespace Mlt;

class TestService : public QObject
{
    Q_OBJECT
    
public:
    TestService()
    {
        repo = Factory::init();
    }

    ~TestService()
    {
        Factory::close();
    }
    
private Q_SLOTS:
    void CanIdentifyServicesFromFactory()
    {
        Profile profile;
        Producer producer(profile, "color");
        QCOMPARE(mlt_service_identify(producer.get_service()), producer_type);
        Filter filter(profile, "resize");
        QCOMPARE(mlt_service_identify(filter.get_service()), filter_type);
        Transition transition(profile, "mix");
        QCOMPARE(mlt_service_identify(transition.get_service()), transition_type);
        Consumer consumer(profile, "null");
        QCOMPARE(mlt_service_identify(consumer.get_service()), consumer_type);
    }
    
    void CanIdentifyServicesFromAPI()
    {
        mlt_profile profile = mlt_profile_init(NULL);

        mlt_playlist playlist = mlt_playlist_new(profile);
        QCOMPARE(mlt_service_identify(MLT_PLAYLIST_SERVICE(playlist)), playlist_type);
        mlt_tractor tractor = mlt_tractor_new();
        QCOMPARE(mlt_service_identify(MLT_TRACTOR_SERVICE(tractor)), tractor_type);
        QCOMPARE(mlt_service_identify(MLT_MULTITRACK_SERVICE(mlt_tractor_multitrack(tractor))), multitrack_type);

        mlt_producer producer = mlt_producer_new(profile);
        QCOMPARE(mlt_service_identify(MLT_PRODUCER_SERVICE(producer)), producer_type);
        mlt_filter filter = mlt_filter_new();
        QCOMPARE(mlt_service_identify(MLT_FILTER_SERVICE(filter)), filter_type);
        mlt_transition transition = mlt_transition_new();
        QCOMPARE(mlt_service_identify(MLT_TRANSITION_SERVICE(transition)), transition_type);
        mlt_consumer consumer = mlt_consumer_new(profile);
        QCOMPARE(mlt_service_identify(MLT_CONSUMER_SERVICE(consumer)), consumer_type);
    }

private:
    Repository* repo;
};

QTEST_APPLESS_MAIN(TestService)

#include "test_service.moc"
```

All tests in `CanIdentifyServicesFromFactory()` pass. In `CanIdentifyServicesFromAPI()`, the tests pass until it reaches `mlt_producer_new()` on line 62. This is your case, and your change lets it pass but then tests on `mlt_filter_new()`, `mlt_transition_new()`, and `mlt_consumer_new()` all fail. This is because mlt_factory takes care of setting property `mlt_type`. This makes me wonder: _how is your producer that is not identified is created and why it is failing to be identified?_

To best address the unit test, I think other changes should be made including **removal** of
`mlt_properties_set( properties, "resource", "<producer>" );` inside `mlt_producer_init()`! Because producers that do not set a resource property (e.g. frei0r) end up serializing this (ugly), and this property should only be set by the derived producer if needed.

Then, setting mlt_type property inside each service class' init function like this:

```
diff --git a/src/framework/mlt_consumer.c b/src/framework/mlt_consumer.c
index 61d21700..21f02403 100644
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -126,6 +126,7 @@ int mlt_consumer_init( mlt_consumer self, void *child, mlt_profile profile )
                        mlt_properties_set_data( properties, "_profile", profile, 0, (mlt_destructor)mlt_profile_close, NULL );
                }
                apply_profile_properties( self, profile, properties );
+               mlt_properties_set( properties, "mlt_type", "consumer" );
 
                // Default rescaler for all consumers
                mlt_properties_set( properties, "rescale", "bilinear" );
diff --git a/src/framework/mlt_filter.c b/src/framework/mlt_filter.c
index 7df9be79..8d37a2e5 100644
--- a/src/framework/mlt_filter.c
+++ b/src/framework/mlt_filter.c
@@ -55,6 +55,7 @@ int mlt_filter_init( mlt_filter self, void *child )
                service->close_object = self;
 
                // Default in, out, track properties
+               mlt_properties_set( properties, "mlt_type", "filter" );
                mlt_properties_set_position( properties, "in", 0 );
                mlt_properties_set_position( properties, "out", 0 );
 
diff --git a/src/framework/mlt_producer.c b/src/framework/mlt_producer.c
index 8cff0066..8d758743 100644
--- a/src/framework/mlt_producer.c
+++ b/src/framework/mlt_producer.c
@@ -91,7 +91,7 @@ int mlt_producer_init( mlt_producer self, void *child )
                        mlt_properties properties = MLT_SERVICE_PROPERTIES( parent );
 
                        // Set the default properties
-                       mlt_properties_set( properties, "mlt_type", "mlt_producer" );
+                       mlt_properties_set( properties, "mlt_type", "producer" );
                        mlt_properties_set_position( properties, "_position", 0.0 );
                        mlt_properties_set_double( properties, "_frame", 0 );
                        mlt_properties_set_double( properties, "_speed", 1.0 );
@@ -101,7 +101,6 @@ int mlt_producer_init( mlt_producer self, void *child )
                        mlt_properties_set_position( properties, "out", MAX(0, p - 1) );
                        mlt_properties_set_position( properties, "length", p );
                        mlt_properties_set( properties, "eof", "pause" );
-                       mlt_properties_set( properties, "resource", "<producer>" );
 
                        // Override service get_frame
                        parent->get_frame = producer_get_frame;
diff --git a/src/framework/mlt_transition.c b/src/framework/mlt_transition.c
index 179aed3a..d407ffc2 100644
--- a/src/framework/mlt_transition.c
+++ b/src/framework/mlt_transition.c
@@ -54,6 +54,7 @@ int mlt_transition_init( mlt_transition self, void *child )
                service->close = ( mlt_destructor )mlt_transition_close;
                service->close_object = self;
 
+               mlt_properties_set( properties, "mlt_type", "transition" );
                mlt_properties_set_position( properties, "in", 0 );
                mlt_properties_set_position( properties, "out", 0 );
                mlt_properties_set_int( properties, "a_track", 0 );
```

ddennedy:
I found a use for mlt_parser in Shotcut and ran into the issue reported. I found that matching resource to "<producer>" was not enough. Those cases were so-called "cut" producers due to `mlt_producer_cut()` used in mlt_playlist.c. There were still some other producers that have a legitimate file resource that have mlt_type = "mlt_producer" instead of "producer". Trying to change things liked mlt_producer.c to set mlt_type to "producer" raised all sorts of bugs where fixing them would be risky. So, it was best to leave that alone, and simply match on "mlt_producer", which catches all cases. My change in 8f74af36 also includes the unit tests and setting mlt_type in mlt_consumer, mlt_transition, and mlt_filter.

PR #473, opened 2019-09-04
Fix returning image data for Python 3 bindigs

jliljebl:
Both frame.get_image() and mlt.frame_get_waveform() return image data as char[].

In Python 2 strings are byte arrays and we were able to use PyString_FromStringAndSize to create the returned PyObject.

In Python 3 using PyUnicode_FromStringAndSize will try to create UTF-8 string from image data and obviously fail.

We need to use PyByteArray_FromStringAndSize instead.

Tested and works on my system.


ddennedy:
thanks!

PR #482, opened 2019-10-10
Fix broken -profile on melt command line

j-b-m:
The -profile command line param from melt was broken with this commit:
https://github.com/mltframework/mlt/commit/33493c70c7b076b03181296f2a549437a457fbb3

To reproduce:
melt -profile atsc_1080p_25 color:red

produces a pal sized result, profiles were not found since mlt_factory was not initialized
This patch fixes the problem


ddennedy:
Good catch. Thank you

PR #483, opened 2019-10-10
fix building with python2 and python3

mikhailnov:
I wanted to build mlt with both python2 and python3 for ROSA Linux and had to patch like this to allow this to be done.
https://abf.io/import/mlt/commit/eff21006ee45d596cf4456ff2469f51f5d55cde9

### DIFF
--- a/src/swig/python/build
+++ b/src/swig/python/build
@@ -1,6 +1,8 @@
 #!/bin/sh
 
 CXX=${CXX:-g++}
+PYTHON=${PYTHON:-python3}
+path=`which "$PYTHON" 2> /dev/null`
 
 if [ "$1" = "clean" ]
 then
@@ -8,14 +10,12 @@ then
 	exit 0
 fi
 
-path=`which python3 2> /dev/null`
-
 if [ $? = 0 ]
 then
 	# Change this as needed
-	export PYTHON_INCLUDE=`python3 -c "import sys;print(\"{}/include/python{}.{}\".format(sys.prefix,*sys.version_info))"`
+	export PYTHON_INCLUDE=`"$PYTHON" -c "import sys;print(\"{}/include/python{}.{}\".format(sys.prefix,*sys.version_info))"`
 
-	[ ! -d "$PYTHON_INCLUDE" ] && echo python development missing && exit 1
+	[ ! -d "$PYTHON_INCLUDE" ] && echo "$PYTHON" development missing && exit 1
 
 	ln -sf ../mlt.i
 
@@ -26,7 +26,7 @@ then
 	${CXX} -fPIC -D_GNU_SOURCE ${CXXFLAGS} -c -I../.. -I$PYTHON_INCLUDE mlt_wrap.cxx || exit $?
 
 	# Create the module
-	${CXX} ${CXXFLAGS} -shared mlt_wrap.o -L../../mlt++ -lmlt++ -L../../framework -lmlt $(python3-config --ldflags) -o _mlt.so || exit $?
+	${CXX} ${CXXFLAGS} -shared mlt_wrap.o -L../../mlt++ -lmlt++ -L../../framework -lmlt $("${PYTHON}-config" --ldflags) -o _mlt.so || exit $?
 else
 	echo Python not installed.
 	exit 1
--- a/src/swig/python/build
+++ b/src/swig/python/build
@@ -13,9 +13,8 @@ fi
 if [ $? = 0 ]
 then
 	# Change this as needed
-	export PYTHON_INCLUDE=`"$PYTHON" -c "import sys;print(\"{}/include/python{}.{}\".format(sys.prefix,*sys.version_info))"`
-
-	[ ! -d "$PYTHON_INCLUDE" ] && echo "$PYTHON" development missing && exit 1
+	export PYTHON_INCLUDE="$("${PYTHON}-config" --includes)"
+	[ -z "$PYTHON_INCLUDE" ] && echo "$PYTHON" development missing && exit 1
 
 	ln -sf ../mlt.i
 


ddennedy:
I hope it works well enough for all.

sergiomb2:
Hi, 
I have another suggestion which I think is better final line [1] 

[1] 
```
export PYTHON_INCLUDE=`"$PYTHON" -c "from sysconfig import get_paths; print(get_paths()['include'])"`
```

ddennedy:
@sergiomb2 why? Is `python[23]-config --includes` broken on your distro? Can you cite official documentation to support your claim?

mikhailnov:
> Hi,
> I have another suggestion which I think is better final line [1]
>
> [1]
>
> |export PYTHON_INCLUDE=`"$PYTHON" -c "from sysconfig import get_paths; 
> print(get_paths()['include'])"` |
>
>
This works incorrectly on Ubuntu 19.04 and Python 2:

user@pay2:~$ python -c "from sysconfig import get_paths; 
print(get_paths()['include'])"
/usr/local/include/python2.7

There is nothing in /usr/local/include/python2.7, system-wide prefix is 
/usr, not /usr/local

This gives correct results on ROSA:

rosa-2016 ~ # python -c "from sysconfig import get_paths; 
print(get_paths()['include'])"
/usr/include/python2.7

But note that python[23]-configs gives ready to use CFALGS:

rosa-2016 ~ # python3-config --includes
-I/usr/include/python3.5m -I/usr/include/python3.5m
rosa-2016 ~ # python3-config --libs
-lpython3.5m -lpthread -ldl  -lutil -lm
rosa-2016 ~ #

Using your method will require constructing (guessing!) CFLAGS manually.

Output of python[23]-config may be different on different OSes, for 
eample note --ldflags on DragonFlyBSD and compare them with GNU/Linux 
ROSA and Ubuntu above:

[root@nas-dfly ~]# python3-config --includes
-I/usr/local/include/python3.6m -I/usr/local/include/python3.6m
[root@nas-dfly ~]# python3-config --ldflags
  -L/usr/local/lib -lpython3.6m -lintl -ldl -L/usr/local/lib -lintl 
-lutil -lm  -Wl,--export-dynamic
[root@nas-dfly ~]#

python-config has given us ready to use working CFLAGS, and we do not 
have to guess them.

sergiomb2:
here [1] is the problem when PYTHON_INCLUDE=-I/usr/local/include/python3.6m -I/usr/local/include/python3.6m .
Notes: ubuntu is bronken and I use "${PYTHON}-config" --ldflags .
But since we are talking I propose copy python to python3 and have 2 directories like I do in Fedora [2]

[1]
`-c -I../.. -I$PYTHON_INCLUDE`

[2] 
https://src.fedoraproject.org/rpms/mlt/blob/master/f/mlt.spec#_174

sergiomb2:
note that all swig/python/*.py have she bangs to one python version , for example [1]

[1]
https://github.com/mltframework/mlt/blob/master/src/swig/python/codecs.py#L1

mikhailnov:
07.11.2019 12:28, Sérgio Basto пишет:
    
    
      
      here [1] is the problem when
        PYTHON_INCLUDE=-I/usr/local/include/python3.6m
        -I/usr/local/include/python3.6m .
      
    
    I don't understand what the problem is. How do you get /usr/local
    when building a distro package in Fedora?!
    
      
        Notes: ubuntu is bronken and I use "${PYTHON}-config" --ldflags
        .
        But since we are talking I propose copy python to python3 and
        have 2 directories like I do in Fedora [2]
      [1]
        -c -I../.. -I$PYTHON_INCLUDE
      [2]
        https://src.fedoraproject.org/rpms/mlt/blob/master/f/mlt.spec#_174
        [
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/mltframework/mlt/pull/483?email_source=notifications\u0026email_token=ADYSBIEMOL3MWI3KT7RQRN3QSPNULA5CNFSM4I7SFE6KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEDLYYFY#issuecomment-550997015",
"url": "https://github.com/mltframework/mlt/pull/483?email_source=notifications\u0026email_token=ADYSBIEMOL3MWI3KT7RQRN3QSPNULA5CNFSM4I7SFE6KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEDLYYFY#issuecomment-550997015",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]
      
    
    That is a hack. ( cd src/swig/python3; export PYTHON=python2;
      ./build ) how I do in ROSA (https://abf.io/import/mlt/blob/rosa2019.1/mlt.spec#lc-211)
      is a much shorter and because of this more reliable solution than
      yours in Fedora, at least I believe so. Is there any reason to run
    %if %{with python3}
      cp -a src/swig/python src/swig/python3
      find src/swig/python3 -name '*.py' | xargs sed -i
      '1s|^#!/usr/bin/env python|#!%{__python3}|'
      sed -i -e 's|which python|which python3|' src/swig/python3/build
      sed -i -e 's|python -c|python3 -c|' src/swig/python3/build
      sed -i -e 's|python-config|python3-config|' src/swig/python3/build
      %endif
      %if %{with python2}
      find src/swig/python -name '*.py' | xargs sed -i
      '1s|^#!/usr/bin/env python|#!%{__python2}|'
      sed -i -e 's|which python|which python2|' src/swig/python/build
      sed -i -e 's|python -c|python2 -c|' src/swig/python/build
      sed -i -e 's|python-config|python2-config|' src/swig/python/build
      %else
      rm -rf src/swig/python
      %endif
    
    instead of
    cd src/swig/python3
      export PYTHON=python2
      ./build
    ?

mikhailnov:
07.11.2019 12:28, Sérgio Basto пишет:
>
> here [1] is the problem when PYTHON_INCLUDE=-I/usr/local/include/python3.6m -I/usr/local/include/python3.6m .

I don't understand what the problem is. How do you get /usr/local when building a distro package in Fedora?!

> Notes: ubuntu is bronken and I use "${PYTHON}-config" --ldflags .
> But since we are talking I propose copy python to python3 and have 2 directories like I do in Fedora [2]
>
> [1]
> -c -I../.. -I$PYTHON_INCLUDE
>
> [2]
> https://src.fedoraproject.org/rpms/mlt/blob/master/f/mlt.spec#_174

That is a hack. `( cd src/swig/python3; export PYTHON=python2; ./build )` how I do in ROSA (https://abf.io/import/mlt/blob/rosa2019.1/mlt.spec#lc-211) is a much shorter and because of this more reliable solution than yours in Fedora, at least I believe so. Is there any reason to run


> note that all swig/python/*.py have she bangs to one python version , for example [1]

Good catch... We should fix it

mikhailnov:
Shebang is not a problem because it does not exist in `mlt.py` which is the only `*.py` file being installed. But I had already written a Makefile capable of fixing shebangs before I noticed that shebang is not a problem, sent it in https://github.com/mltframework/mlt/pull/495

sergiomb2:
my propose  https://github.com/mltframework/mlt/pull/496

ddennedy:
> But since we are talking I propose copy python to python3 and have 2 directories like I do in Fedora [2]

No, python2 is no longer supported. Now, I regret accepting the change that restores support for it in MLT. if you need it in your distro, patch the source for your package and do not try to upstream it - at least for this package.

> note that all swig/python/*.py have she bangs to one python version
> Good catch... We should fix it

No, we should not. If you want to use the **example** python script with python 2.x, then invoke it with that interpreter or edit the example.

`-I$PYTHON_INCLUDE` is obviously a problem since the output of `python3-config --includes` includes the '-I` flags.

sergiomb2:
OK , no problem , we in Fedora + RPMFusion have one python-mlt client which is Flowblade .
Flowable will have an python3 version before end of the year [1] , and that is why I'd like have python2 and python3 version to migration be more smooth .

About `-I$PYTHON_INCLUDE` seems to me we just need remove -I and use just `$PYTHON_INCLUDE`

Thank you


[1]
https://github.com/jliljebl/flowblade/issues/597

ddennedy:
> About -I$PYTHON_INCLUDE seems to me we just need remove -I and use just $PYTHON_INCLUDE

Yes, I made that change.

PR #490, opened 2019-10-26
Add CMake build system

vpinon:
Hello again,

My initial goal was to get more easily windows devs on board, and so allow them to work on Kdenlive / MLT and all deps with tools they are used to: Visual Studio compilers, debuggers & profilers.

1st step is to have a more portable build system: CMake is what I know from KDE/Qt culture.
(note: having packaged Kdenlive / MLT in MXE and Craft, I know that MLT's custom "autotools" is very special and breaks common things like out of source tree building, special cross-compiling instructions...)

This has been (quickly) tested in Kdenlive on Windows with MinGW and Linux with GCC & Clang.

Notes:

- I haven't worked on deprecated modules, neither on the ones for which I was missing deps.
- Now MSVC porting needs work on the code; plus as every new compiler it also raises new interesting warnings

I would be happy to share this more publicly; it doesn't harm current process in place :)


a17r:
As a packager, having this not just for Windows would be much appreciated.

eszlari:
For consistency and simplicity, shotcut and webvfx could be ported to cmake too. Then shotcut adopt krita's [approach](https://invent.kde.org/kde/krita/tree/master/3rdparty) and utilize cmake's [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) functionality and get a rid of the build-shotcut.sh scripts.

Qt is moving to cmake for v6 too (https://lists.qt-project.org/pipermail/development/2018-October/034023.html) and they will be improving Qt Creator's cmake support even more.

ddennedy:
I agree to include this, but I do not yet volunteer to handle bugs related to it. The existing configure/Makefile system will continue as it currently offers some options not included here as far as I can tell. I will do my best to maintain it for files and modules that I add or remove directly or through pull requests.

PR #495, opened 2019-11-07
Implement standard Makefile for Python bindings and installation of Python bindings

mikhailnov:


ddennedy:
I have been through the topic of installing the python bindings already with others. There is no universal way to do it. I am not going to put that into MLT and deal with competing pull requests to correct it per distribution. You need to put that logic into your package build.

I also do not agree to extending python2 support by changing shebangs.

Finally, this change breaks integration with the top-level configure and Makefile not to mention there is a new CMake build system.

PR #496, opened 2019-11-07
Calculate the correct python includes dir

sergiomb2:


ddennedy:
This fails on Debian and Ubuntu systems. It also does not include all the directories `python3-config --includes` may contain.

PR #497, opened 2019-11-08
Correctly initialize avfilter default values for integer params

j-b-m:
Some avfilter parameters were not set to the correct default value, for example unsharp.
Or is there any reason to retrieve the double value for int params ?


bmatherly:
LGTM

PR #510, opened 2019-12-28
Fix chroma bleeding due to colorspace conversion.

bmatherly:
This is a proposal to help reduce the chroma bleeding when performing colorspace conversions.

**Colorbars Before:**
before

**Colorbars After:**
after

The vectorscope uses data that was converted from RGB->YUV420. The Zoom and Waveform scopes uses data that was converted in a round trip: RGB->YUV420->RGB.

I do not know the significance of the two removed flags.

Based on this comment, they might not be completely implemented?
https://ffmpeg.org/doxygen/trunk/swscale_8h_source.html#l00077

Also, this is a relevant change: https://github.com/mltframework/mlt/commit/adc5a2284b3a1073cb364c5f07d1d7c97e94c937
It seems that prior to that change, the flags were applied selectively based on the format.

For example, this was removed:
```
	// Set swscale flags to get good quality	
	switch ( *format )
	{
		case mlt_image_yuv422:
			interp |= SWS_FULL_CHR_H_INP;	
			break;	
		case mlt_image_rgb24:
			interp |= SWS_FULL_CHR_H_INT;	
			break;	
		case mlt_image_rgb24a:
		case mlt_image_opengl:
			interp |= SWS_FULL_CHR_H_INT;	
			break;
		default:
			// XXX: we only know how to rescale packed formats
			return 1;
	}
```

Maybe you have more history about the application of those flags.

### DIFF
--- a/src/modules/avformat/common.c
+++ b/src/modules/avformat/common.c
@@ -22,7 +22,7 @@
 #include <libavutil/channel_layout.h>
 #include <libavutil/samplefmt.h>
 
-int mlt_default_sws_flags = SWS_BICUBIC | SWS_FULL_CHR_H_INP | SWS_FULL_CHR_H_INT | SWS_ACCURATE_RND;
+int mlt_default_sws_flags = SWS_BICUBIC | SWS_ACCURATE_RND;
 
 int mlt_to_av_sample_format( mlt_audio_format format )
 {


ddennedy:
I do not think this is acceptable because it reduces the non-spatial accuracy of color conversion.
The change that you linked to references a [Shotcut bug](https://github.com/mltframework/shotcut/issues/674) with more information.

But also there were a couple threads in the forum around this time where color accuracy was criticized, tested, improved, and verified. Here is [one](https://forum.shotcut.org/t/shotcut-color-accuracy/8365/46) and [another](https://forum.shotcut.org/t/intermediate-files-for-editing/7856/43?u=shotcut)

As you can see a lot of work was put into this change, and I am not keen to regress in these test cases even though we do appears to lose some spatial accuracy with respect to chroma sub-sampling. I think chroma bleed will have to wait for the major pipeline change to support a higher bit depth without chroma sub-sampling (rgb24a unacceptable due to clamping).

bmatherly:
Thanks Dan. I guess there are difficult tradeoffs to make. It is disappointing. I would expect that color space conversion would be a solved problem by now. At least we have an intuitive sense of the tradeoffs and now we have some easier tools for seeing the effect of future changes. Feel fee to close this merge request if you do not think it will be useful.

ddennedy:
Please consider making a pull request that adds a runtime option to set the swscale flags. ffmpeg has `-sws_flags` and we could add a similar consumer property that gets passed upstream similar to other consumer properties. It does not necessarily need to support the same manner of string flag combinations like ffmpeg handles. It could be simply numeric, and one needs to figure out the correct number for their combination using the values in the header.

bmatherly:
> Please consider making a pull request that adds a runtime option to set the swscale flags.

I am willing to do that. Can you give more detail about who/what would be setting the flags? Are you thinking there would be an option exposed to the user in Shotcut? Or users would add flags in the "Other" tab under export?

Thinking about this some more, I expect that it is probably possible to determine the best use of the flags automatically (programatically). But I think that "mlt_default_sws_flags" would have to turn into a function with multiple parameters (in format, out format, scaling up or down or none, etc). I think that we could find the right combination if we were willing to go "one more round" on the deep color testing. But as you suggest, it might be best to save that effort for the next major refactor.

PR #511, opened 2019-12-31
Improve colorspace conversion.

bmatherly:
Here is another suggestion for colorspace conversion. I was able to reproduce the luma reduction reported by users and test for that. This proposal does not completely eliminate chroma bleeding. But it greatly reduces it and the scopes look very clean when color bars are played.

**Before:**
before
**After**
after2

One thing I like about this approach is that it provides a framework to further refine the flag decision down the road if further optimization is desired.

I understand that the timing is not good for this. Maybe it would be good to sit on this for a while and after a stable release it could be released to beta testers for further quality testing.

### DIFF
--- a/src/modules/avformat/common.c
+++ b/src/modules/avformat/common.c
@@ -20,9 +20,65 @@
 #include "common.h"
 
 #include <libavutil/channel_layout.h>
+#include <libavutil/pixdesc.h>
 #include <libavutil/samplefmt.h>
 
-int mlt_default_sws_flags = SWS_BICUBIC | SWS_FULL_CHR_H_INP | SWS_FULL_CHR_H_INT | SWS_ACCURATE_RND;
+int get_sws_flags(int srcwidth, int srcheight, int srcformat, int dstwidth, int dstheight, int dstformat)
+{
+	// Use default flags unless there is a reason to use something different.
+	int flags = SWS_BICUBIC | SWS_FULL_CHR_H_INP | SWS_FULL_CHR_H_INT | SWS_ACCURATE_RND;
+
+	if( srcwidth != dstwidth || srcheight != dstheight )
+	{
+		// Any resolution change should use default flags
+		return flags;
+	}
+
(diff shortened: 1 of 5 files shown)


src/modules/avformat/common.c
```diff
@@ -20,9 +20,65 @@
...
+#include <libavutil/pixdesc.h>
 #include <libavutil/samplefmt.h>
 
-int mlt_default_sws_flags = SWS_BICUBIC | SWS_FULL_CHR_H_INP | SWS_FULL_CHR_H_INT | SWS_ACCURATE_RND;
+int get_sws_flags(int srcwidth, int srcheight, int srcformat, int dstwidth, int dstheight, int dstformat)
+{
+	// Use default flags unless there is a reason to use something different.
+	int flags = SWS_BICUBIC | SWS_FULL_CHR_H_INP | SWS_FULL_CHR_H_INT | SWS_ACCURATE_RND;
+
+	if( srcwidth != dstwidth || srcheight != dstheight )
```
bmatherly: This function is not currently used for any resolution scaling. But I added this so that it could be easily ported for use in filter_swscale. Also, this avoids the possibility that someone might accidentally use it for resolution scaling in the future and get unexpected results.

src/modules/avformat/common.h
```diff
@@ -29,6 +29,6 @@ mlt_channel_layout av_channel_layout_to_mlt( int64_t layout );
 mlt_channel_layout mlt_get_channel_layout_or_default( const char* name, int channels );
 int mlt_set_luma_transfer( struct SwsContext *context, int src_colorspace,
 	int dst_colorspace, int src_full_range, int dst_full_range );
-extern int mlt_default_sws_flags;
+int get_sws_flags(int srcwidth, int srcheight, int srcformat, int dstwidth, int dstheight, int dstformat);
```
ddennedy: This should have a `mlt_` prefix.

PR #513, opened 2020-01-04
fix memory leak in MltRepository destructor

rodolforg:


ddennedy:
The current implementation is intentional. `mlt_repository` and `mlt_factory` work hand-in-hand. `mlt_factory_init()` and `mlt_factory_close()` take ownership of the `mlt_repository`. `Mlt::Repository` is simply a class to access the `mlt_repository` returned by `Mlt::Factory::init()`. Use `Mlt::Factory::close()`. I am sorry if you do not like that, but I am not going to change it now and break existing apps.

rodolforg:
Even if I call `Mlt::Factory::close()`, the `Mlt::Repository` returned by `Mlt::Factory::init()` is not deleted/deallocated. Just a few bytes leaked though.

ddennedy:
Of course, Mlt::Factory::close() does not delete Mlt:Repository*. Rather, it calls `mlt_factory_close()`, which calls `mlt_repository_close()` on the mlt_repository it manages. 
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_factory.c#L507

PR #515, opened 2020-01-09
Add package installation instructions

tacman:
Should make installing it for the first time faster.


ddennedy:
Those packages are insufficient and specific to Debian based Linux.

PR #516, opened 2020-01-26
Add rubberband module and rbpitch filter.

bmatherly:


ddennedy:
Please add CMake support.

bmatherly:
> Please add CMake support.

I am willing to do this if you require it. But it is inconvenient for me because:
1) I do not have any experience with CMake
2) My development environment is Ubuntu 16.04 which provides CMake 3.5.1 but the build system requires CMake 3.12 due to the use of add_compile_definitions()

So, before i go off to install a new build environment and learn CMake, can you help me understand the need? Why do we maintain two build systems? Who uses CMake? Why is it OK that there are other modules that do not use CMake (resample, ndi, etc)?

Maybe the people who use CMake can volunteer to maintain it?

Also (off topic now), if we support CMake, shall we add it to the weekly regression build test?

ddennedy:
I am going to wait a little while to see what happens with your rubberband pull requests. rubberband is already in Debian, Ubuntu, and Fedora. I do not yet want to include something into MLT based on a fork. If we do not get anywhere in 30 days we can fork it into the MLT as we have done with libebur128.

ddennedy:
Kdenlive is using cmake now for some of their builds. Missing modules are because they did not need them, and I was not strict about accepting it. However, I want everyone to try to maintain it at a basic level of file additions and removals. Maybe in the long run, we will switch over to it completely. I want to see how and where it goes. I have not had time to expand upon it and not really motivated either. I just wanted to accept it and see if Kdenlive and others expand it and maintain it. So far not really. :-( I will not require it of you. Rather, I will take the time when reviewing this to add it so I get more experience with it in MLT.

bmatherly:
> I am going to wait a little while

That is reasonable.

This one is merely a convenience to simplify the build_shotcut.sh script:
https://github.com/breakfastquay/rubberband/pull/23
It does not break any compatibility with currently installed versions.

This one is a programming convenience:
https://github.com/breakfastquay/rubberband/pull/24
If it is not well received, I will modify the code to make it unnecessary.

bmatherly:
> rubberband is already in Debian, Ubuntu, and Fedora

This comment inspired me. So I modified the patch to not depend on any rubberband API changes. It should be compatible with currently released versions.

ddennedy:
That’s great. I don’t mind at all to use your repo in the build scripts until you’re pull request is merged to remove the extra dependencies.

ddennedy:
I merged this (rather than rebase) in case you want to keep your rubberband branch to do more work in it.

PR #523, opened 2020-02-10
Add pitch compensation to timewarp producer

bmatherly:


src/modules/core/producer_timewarp.yml
```diff
@@ -52,6 +52,14 @@ parameters:
     required: yes
     mutable: no
 
+  - identifier: warp_pitch_compensate
```
bmatherly: I would be open to comment on this name. Feels verbose.
It has the "warp" prefix because the producer adopts all the properties of the encapsulated producer, and I do not want there to be a conflict.

src/modules/rubberband/filter_rbpitch.cpp
```diff
@@ -66,29 +66,62 @@ static int rbpitch_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *
...
+
+	// Detect the last frame
 	bool last_frame = false;
 	if ( mlt_filter_get_position( filter, frame ) + 1 == mlt_filter_get_length2( filter, frame ) )
 	{
 		// This is the last mlt frame. Drain the rest of the stretcher
 		last_frame = true;
 	}
 
+	// Sanity check parameters
```
bmatherly: This is my solution for graceful degradation of extreme parameters. These are the result of trial and error. They could be easily tweaked in the future.

src/modules/rubberband/filter_rbpitch.cpp
```diff
@@ -66,29 +66,62 @@ static int rbpitch_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *
...
+	// the future.
+	double pitchscale = mlt_properties_get_double( unique_properties, "pitchscale" );
+	if (pitchscale > 50)
+	{
+		pitchscale = 50;
+	}
+	else if ( pitchscale < 0.05 )
+	{
+		pitchscale = 0.05;
+	}
```
ddennedy: Please use the CLAMP macro that we have in mlt_types.h. Also, `50` is an integer literal; please change to `50.0` to make it a double literal.

src/modules/core/producer_timewarp.yml
ddennedy: I would be ok with simply `warp_pitch`

bmatherly: agree. Done.

src/modules/rubberband/filter_rbpitch.cpp
bmatherly: done

PR #542, opened 2020-04-06
(hopefully) make GPU effects (aka. movit) less crashy.

sandsmark:
movit assert()s on all errors, so make really sure we don't pass in invalid values.

Looking at a couple of bug reports, I think some of the problems are because of invalid values passed around and handed directly to movit, and movit just asserts instead of handling invalid values gracefully.

### DIFF
--- a/src/modules/opengl/filter_movit_blur.cpp
+++ b/src/modules/opengl/filter_movit_blur.cpp
@@ -28,6 +28,10 @@ using namespace movit;
 
 static int get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
 {
+	if (!width || !height || *width < 1 || *height < 1) {
+		return 1;
+	}
+
 	mlt_filter filter = (mlt_filter) mlt_frame_pop_service( frame );
 	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
 	GlslManager::get_instance()->lock_service( frame );
--- a/src/modules/opengl/filter_movit_blur.cpp
+++ b/src/modules/opengl/filter_movit_blur.cpp
@@ -28,10 +28,6 @@ using namespace movit;
 
 static int get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
 {
-	if (!width || !height || *width < 1 || *height < 1) {
-		return 1;
-	}
-
 	mlt_filter filter = (mlt_filter) mlt_frame_pop_service( frame );
 	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
 	GlslManager::get_instance()->lock_service( frame );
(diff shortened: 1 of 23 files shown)


src/modules/opengl/filter_movit_convert.cpp
```diff
@@ -465,6 +470,11 @@ static int movit_render( EffectChain *chain, mlt_frame frame, mlt_image_format *
 // Create an MltInput for an image with the given format and dimensions.
 static MltInput* create_input( mlt_properties properties, mlt_image_format format, int aspect_width, int aspect_height, int width, int height )
 {
+	if (width < 1 || height < 1) {
+		mlt_log_error( NULL, "Invalid frame size for create_input: %dx%d.\n", width, height );
+		return NULL;
```
ddennedy, line 475: You can use `nullptr` if you want in our C++ code since we now require C++11.

src/modules/opengl/mlt_movit_input.cpp
```diff
@@ -57,7 +73,11 @@ void MltInput::useYCbCrInput(const ImageFormat& image_format, const YCbCrFormat&
 
 void MltInput::set_pixel_data(const unsigned char* data)
 {
-	assert(input);
+    if (!input) {
```
ddennedy, line 76: fix indentation to use tabs

src/modules/opengl/mlt_movit_input.h
```diff
@@ -44,10 +44,10 @@ class MltInput
 
 private:
 	mlt_image_format m_format;
-	unsigned m_width, m_height;
+	unsigned m_width = 1, m_height = 1;
```
ddennedy, line 47: This valid but a minor consistency complaint: this change mixes initialization of members between declaration and constructor.

src/modules/opengl/filter_movit_blur.cpp
```diff
@@ -28,6 +28,10 @@ using namespace movit;
 
 static int get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
 {
+	if (!width || !height || *width < 1 || *height < 1) {
+		return 1;
```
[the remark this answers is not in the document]
sandsmark: > So, the width and height check belongs after that call. Also, it is preferable behavior to skip the filter (with a log message) instead of return an error, because an error return will cause a white image to be created. This comment applies to all of the other filters as well.

I assumed just returning `error` as the other checks does is the correct way to do this.

sandsmark:
> I usually make Shotcut release builds with NDEBUG defined to suppress asserts, but it is hard to reproduce runtime errors in debug builds.

Yeah, I actually started on replacing the check_error()s with another macro that just returned instead of assert()ing and crashing, but that's a bit bigger job. And the errors usually are valid anyways, so the application can end in a weird place and crash elsewhere.

A lot of the errors at least for kdenlive at least is probably because of the gl context stuff, it seems quite messy, but making the opengl code in mlt a bit more robust is nice anyways.

PR #546, opened 2020-04-08
Chain/Link Proof of Concept with Time Remapping example

bmatherly:
### Purpose
The purpose of this merge request is to create a design and code review for this concept. The code is not expected to be merged without a lot of discussion and probably changes.

### Concept
The concept is to add two new classes to the MLT framework: chain and link. A chain is a type of producer that can contain links. A link is a type of producer that can request frames from the next link that they are connected to. A link could request multiple source frames for one output frame. Or, it could request the same source frame multiple times. The advantage of this (unlilke filters) is that a link can perform temporal effects like speed conversions or motion effects.

This PoC also includes one link implementation: timeremap. The timeremap link allows the user to specify a sequence of keyframes that map input frame times to output frame times. This method allows arbitrary speed and direction changes in real time.

### Example:
First, download the source material.

`wget ftp://vqeg.its.bldrdoc.gov/HDTV/NTIA_source/TouchdownPass_8bit.avi`

This first command demonstrates:
* normal speed leading up to the snap
* smooth speed transition to 0.5x speed during the play
* 0.1x speed during the catch
* smooth transition to normal speed for the celebration
* freeze frame during the jump
* reverse back to the tackle

`melt -profile atsc_1080p_30 -chain TouchdownPass_8bit.avi out=570 -link timeremap map="00:00:00.000=0;00:00:05.000=5;00:00:10.000~=7.5;00:00:15.000~=8;00:00:20.000~=10;00:00:25.000~=15;00:00:30.000=15;;00:00:36.000~=9"`

This next command demonstrates 2x speed with frame blending:

`melt -profile atsc_1080p_30 -chain TouchdownPass_8bit.avi out=570 -link timeremap image_mode=blend map="00:00:00.000=0;00:00:10.000=20"`

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -576,3 +576,21 @@ MLT_6.22.0 {
     mlt_audio_channel_layout_channels;
     mlt_audio_channel_layout_default;
 } MLT_6.20.0;
+
+MLT_6.24.0 {
+  global:
+    mlt_factory_link;
+    mlt_chain_init;
+    mlt_chain_set_source;
+    mlt_chain_get_source;
+    mlt_chain_attach;
+    mlt_chain_detach;
+    mlt_chain_link_count;
+    mlt_chain_move_link;
+    mlt_chain_link;
+    mlt_chain_close;
+    mlt_link_init;
+    mlt_link_connect_next;
+    mlt_link_close;
+    mlt_repository_links;
+} MLT_6.22.0;
--- a/src/framework/mlt_chain.c
+++ b/src/framework/mlt_chain.c
@@ -1,0 +1,460 @@
+/**
+ * \file mlt_chain.c
+ * \brief link service class
+ * \see mlt_chain_s
+ *
+ * Copyright (C) 2020 Meltytech, LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include "mlt_chain.h"
+#include "mlt_factory.h"
+#include "mlt_frame.h"
+#include "mlt_log.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/** \brief private service definition */
+
+typedef struct
+{
+	int link_count;
+	int link_size;
+	mlt_link* links;
+	mlt_producer source;
+	mlt_profile source_profile;
+	mlt_properties source_parameters;
+	mlt_producer begin;
+}
+mlt_chain_base;
+
+/* Forward references to static methods.
+*/
+
+static int producer_get_frame( mlt_producer parent, mlt_frame_ptr frame, int index );
+static void relink_chain( mlt_chain self );
+static void chain_property_changed( mlt_service owner, mlt_chain self, char *name );
+static void source_property_changed( mlt_service owner, mlt_chain self, char *name );
+
+/** Construct a chain.
+ *
+ * \public \memberof mlt_chain_s
+ * \return the new chain
+ */
+
+mlt_chain mlt_chain_init( mlt_profile profile )
+{
+	mlt_chain self = calloc( 1, sizeof( struct mlt_chain_s ) );
+	if ( self != NULL )
+	{
+		mlt_producer producer = &self->parent;
+		if ( mlt_producer_init( producer, self ) == 0 )
+		{
+			mlt_properties properties = MLT_PRODUCER_PROPERTIES( producer );
+
+			mlt_properties_set( properties, "mlt_type", "chain" );
+			mlt_properties_clear( properties, "resource");
+			mlt_properties_clear( properties, "mlt_service");
+			mlt_properties_clear( properties, "in");
+			mlt_properties_clear( properties, "out");
+			mlt_properties_clear( properties, "length");
+
+			producer->get_frame = producer_get_frame;
+			producer->close = ( mlt_destructor )mlt_chain_close;
+			producer->close_object = self;
+
+			mlt_service_set_profile( MLT_CHAIN_SERVICE( self ), profile );
+
+			// Generate local space
+			self->local = calloc( 1, sizeof( mlt_chain_base ) );
+			mlt_chain_base* base = self->local;
+			base->source_profile = mlt_profile_init(NULL);
+
+			// Listen to property changes to pass along to the source
+			mlt_events_listen( MLT_CHAIN_PROPERTIES(self), self, "property-changed", ( mlt_listener )chain_property_changed );
+		}
+		else
+		{
+			free( self );
+			self = NULL;
+		}
+	}
+	return self;
+}
+
+/** Set the source producer.
+ *
+ * \public \memberof mlt_chain_s
+ * \param self a chain
+ * \param source the new source producer
+ */
+
+void mlt_chain_set_source( mlt_chain self, mlt_producer source )
+{
+	int error = self == NULL || source == NULL;
+	if ( error == 0 )
+	{
+		mlt_chain_base* base = self->local;
+		mlt_properties source_properties = MLT_PRODUCER_PROPERTIES( source );
+		int n = 0;
+		int i = 0;
+
+		// Clean up from previous source
+		mlt_producer_close( base->source );
+		mlt_properties_close( base->source_parameters );
+
+		// Save the source producer
+		base->source = source;
+		mlt_properties_inc_ref( source_properties );
+
+		// Create a list of all parameters used by the source producer so that
+		// they can be passed between the source producer and this chain.
+		base->source_parameters = mlt_properties_new();
+		mlt_repository repository = mlt_factory_repository();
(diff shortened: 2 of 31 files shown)


src/framework/mlt_chain.h
```diff
@@ -0,0 +1,56 @@
...
+};
+
+#define MLT_CHAIN_PRODUCER( chain )		( &( chain )->parent )
+#define MLT_CHAIN_SERVICE( chain )		MLT_PRODUCER_SERVICE( MLT_CHAIN_PRODUCER( chain ) )
+#define MLT_CHAIN_PROPERTIES( chain )	MLT_SERVICE_PROPERTIES( MLT_CHAIN_SERVICE( chain ) )
+
+extern mlt_chain mlt_chain_init( mlt_profile );
+extern void mlt_chain_set_source( mlt_chain self, mlt_producer source );
+extern mlt_producer mlt_chain_get_source( mlt_chain self );
+extern int mlt_chain_attach( mlt_chain self, mlt_link link );
```
bmatherly, line 49: I modeled the attach/detach/move functions after the same functions for service filters. My thinking for Shotcut is:
* In the clip properties, the user could selection an options to "allow re-timing" which would cause the clip to be placed in a chain
* If the filter dock detects a chain, it will add a new section for links above the filters (We could probably just call them "time" filters). So the filter types would be: audio, video, time (if retiming is enabled).

src/framework/mlt_frame.h
```diff
@@ -146,6 +146,11 @@ extern mlt_frame mlt_frame_clone( mlt_frame self, int is_deep );
 /* convenience functions */
 extern int mlt_sample_calculator( float fps, int frequency, int64_t position );
 extern int64_t mlt_sample_calculator_to_now( float fps, int frequency, int64_t position );
+extern void mlt_audio_copy( void* dst, void* src, mlt_audio_format format, int channels, int samples, int src_start, int dst_start, int src_max, int dst_max );
+extern int mlt_audio_plane_count( mlt_audio_format format, int channels );
+extern int mlt_audio_plane_size( mlt_audio_format format, int samples, int channels );
+extern void mlt_audio_collapse_channels( mlt_audio_format format, int channels, int allocated_samples, int used_samples, void* buffer );
+extern void mlt_audio_reverse( mlt_audio_format format, int channels, int samples, void* buffer );
```
bmatherly, line 153: After I copy/paste a function for the third time, I feel like it should be moved into the framework. These could all be moved into a new "audio" module if we think the frame file is too cluttered. I think it would be nice if we added an "audio" class to encapsulate format, channels, samples and buffer into a single struct with these (and other) convenience functions.

src/framework/mlt_link.c
```diff
@@ -0,0 +1,129 @@
...
+ * \param default_profile a profile to use if needed (some links derive their frame rate from the next producer)
+ * \return true on error
+ */
+
+int mlt_link_connect_next( mlt_link self, mlt_producer next, mlt_profile default_profile )
+{
+	self->next = next;
+	if ( self->configure )
+	{
+		self->configure( self, default_profile );
```
bmatherly, line 85: This is what triggers a link to check the framerate of downstream links to set up its own framerate. For example, a frame doubling deinterlacer would implement configure() and check the framerate of the "next" producer and them double its own framerate.

src/modules/core/link_timeremap.c
```diff
@@ -0,0 +1,417 @@
...
+
+	int result = 0;
+
+	// Create a  frame
+	*frame = mlt_frame_init( MLT_LINK_SERVICE(self) );
+	mlt_frame_set_position( *frame, mlt_producer_position( MLT_LINK_PRODUCER(self) ) );
+	mlt_properties unique_properties = mlt_frame_unique_properties( *frame, MLT_LINK_SERVICE(self) );
+
+	// Calculate the frames from the next link to be used
+	if ( !mlt_properties_exists( properties, "map" ) )
```
bmatherly, line 282: If the "map" property is not set, then this link is a basic frame rate converter.

src/framework/mlt_chain.c
```diff
@@ -0,0 +1,356 @@
...
+		mlt_chain_base* base = self->local;
+
+		// Save the source producer
+		mlt_producer_close( base->source );
+		base->source = source;
+		mlt_properties_inc_ref( MLT_PRODUCER_PROPERTIES( source ) );
+
+		// Set the producer to be in native frame rate
+		mlt_profile_from_producer( base->source_profile, base->source );
+		mlt_service_set_profile( MLT_PRODUCER_SERVICE(base->source), base->source_profile );
```
bmatherly, line 113: The real source producer for the entire chain is configured to be in a "native" profile so that the next link has access to every single frame. Some producer implementations may need to be updated to work better when the profile is changed. For example, producer_avformat works fine except that the length property is not updated after the profile framerate changes.

src/modules/core/link_timeremap.yml
```diff
@@ -0,0 +1,27 @@
...
+tags:
+  - Audio
+  - Video
+description: Remap frames in time.
+parameters:
+  - identifier: map
+    title: Map
+    type: float
+    description: >
+      A map of input frame times to output frame times.
```
ddennedy, line 18: Should specify the unit of measure for the map time value using the `unit` attribute.

src/framework/mlt_frame.h
ddennedy: Are you thinking about trying to replace the various audio properties in mlt_frame with a single struct? If so, that might mean a lot of changes and potentially with some apps. Otherwise, for now, I am content to have mlt_audio.[ch] and mlt_image.[ch] to move functions out of mlt_frame.[ch].

src/framework/mlt.vers
```diff
@@ -552,3 +552,26 @@ MLT_6.20.0 {
     mlt_profile_scale_height;
     mlt_properties_set_string;
 } MLT_6.18.0;
+
+MLT_6.21.0 {
```
ddennedy, line 556: I only use upcoming version in `mlt.vers` so that this does not need change upon release in a way that conflicts with a development version. For example, here `MLT_6.21.0` is based on `MLT_6.20.0` but then come release `MLT_6.22.0` would be based on `MLT_6.20.0` since there cannot be an empty version `MLT_6.22.0` based on `MLT_6.21.0` if you understand. Simply put, change `MLT_6.21.0` to `MLT_6.22.0` now.

src/framework/mlt_chain.c
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * \public \memberof mlt_chain_s
+ * \return the new link
+ */
+
+mlt_chain mlt_chain_init( mlt_profile profile )
+{
+	mlt_chain self = calloc( 1, sizeof( struct mlt_chain_s ) );
+	if ( self != NULL )
+	{
```
ddennedy, line 62: With regards to style, I am OK to make all new code not have spaces inside parentheses and opening braces on the same line as a statement. Also, I tend to not compare with 0 or NULL. For example,
```
mlt_chain self = calloc(1, sizeof(struct mlt_chain_s));
if (self) {
```
The choice is up to you; I am simply stating that I am trying to use a more normal style now because this old, unique style kind of bothers.

src/framework/mlt_properties.c
```diff
@@ -2246,6 +2246,19 @@ void mlt_properties_clear( mlt_properties self, const char *name )
 	mlt_events_fire( self, "property-changed", name, NULL );
 }
 
+/** Check if a property exists.
```
ddennedy, line 2249: Today, much of the code in MLT and elsewhere simply does `mlt_properties_get()` to check for existence, but that may invoke conversion or serialization of the property. Are you specifically trying to avoid those side effects or forgot the convention?

src/modules/core/producer_melt.c
```diff
@@ -418,6 +432,41 @@ mlt_producer producer_melt_init( mlt_profile profile, mlt_service_type type, con
 					mlt_properties_set_int( properties, "hide", 2 );
 			}
 		}
+		else if ( !strcmp( argv[ i ], "-chain" ) )
```
ddennedy, line 435: Need to update usage info in `melt.c` and `docs/melt.1` man page.

src/framework/mlt_frame.h
bmatherly: Yes. That is what I was thinking. If we make such a change, I would volunteer to change all the code. But I expect it would also break some applications or 3rd party modules. Perhaps there is a way to keep the current interface and add a new one. The current interface application could be a backwards compatible wrapper for the new interface. I would be willing to submit a proposal for this if you are interested.

I will move the audio/image functions into separate files. That would be a step in that direction.

src/framework/mlt_chain.c
bmatherly: Would you have any interest in a Make target or a script to run [Uncrustify ](https://github.com/uncrustify/uncrustify) to enforce a coding style? I have some experience with uncrustify.

Options would be:
1) "Fix" all source files
2) "Check" all source files and report an error if any do not comply

The check could be added to the weekly regression test (optional)

The fix could be used for:
1)  First "big bang" change to update all source files to match style
2) A quick run for developers before submitting code
3) A quick cleanup if the automated regression test reports non-compliant files (if we went that route) or occasional runs to catch up code that has deviated from the style

I think that we would not want to add it to the regression test right away.

src/framework/mlt_properties.c
bmatherly: mlt_properties_get() will break an animation and replace it with a string representation of the last value in the animation. I suppose this change is actually working around a bug. But I also think that it adds clarity to the code. I had been wishing for such a function in the past but never took the time.

src/framework/mlt_chain.c
ddennedy: I am aware of code formatters and used `astyle` in Kino. I am not inclined to enforce it but perhaps willing to accept a mass conversion.

src/framework/mlt_properties.c
ddennedy: I am OK to add `mlt_properties_exist()` but it should be added to Mlt::Properties as well. Due to heavy inheritance, the method on it should be named like `property_exists()` so it does not read like `filter.exists(..)`, which can be misleading .

bmatherly:
Closing this pull request. Will submit a new one since the implementation has evolved.

PR #548, opened 2020-04-11
Make avformat lower priority for image sequences

bmatherly:
The avformat producer does not always exhibit the same behavior as
qimage and pixbuf.

Proposed workaround for https://github.com/mltframework/mlt/issues/543


ddennedy:
They do not need to behave the same, avformat is faster, and avformat maintains a consistent frame rate independent of the consumer/profile frame rate.

bmatherly:
> They do not need to behave the same

Ok. Perhaps that was a bad assumption on my part. The advantage of the loader services having the same behavior is that users will receive a similar experience regardless of what service is available on their system. Without consistent behavior, the services are not really interchangeable.

ddennedy:
I agree with your point about consistency of interchangeable producers. But I would not be surprised if now someone is depending on the new behavior with an image sequence loaded with avformat and either default 25 fps or overridden with a `framerate` property. I do not care too strongly for Shotcut since we control it there (I might want to switch to avformat in the future).
However, I want to get @j-b-m or @vpinon opinion about this.

ddennedy:
@jliljebl how would this affect Flowblade?

bmatherly:
I agree we should check if anyone depends on the current behavior of avformat. If nobody is depending on it, I wonder if it could be changed to be similar to the others? Or, someone could possibly write a new image producer that is uses avformat under the hood in a way that is interchangeable. I might be available to do something like that if it would provide a performance improvement and would not hinder any applications depending on the current behavior.

ddennedy:
I do not agree avformat image sequence behavior in MLT should be changed and made the same as the others for the way it handles frame rate can be preferable. For this reason, I reject this pull request as if agreed by others, I would rather revert b4c3978b7b63a735d3bb82aa55bca2787edd9a27

ydesgagn:
Do you want to add a `-safe` command line option and read loader.safe instead of loader.dict in that case? That way both methods are preserved. On our side we can easily add that flag based on the scenario creation date.

ddennedy:
@ydesgagn  no thank you

ydesgagn:
Understood. Thanks for all this.

ddennedy:
reverted b995dd0f instead

jliljebl:
I just noticed this had a question for me. 

Reading this and the other thread I think I can't point to a behavioral difference that I would have noticed in this area, so reverting to earlier behavior is fine for me.

PR #559, opened 2020-05-10
Precharge the rbpitch filter with repeated samples.

bmatherly:
This avoids gaps in the audio output by always returning the number
of samples provided by the producer.


bmatherly:
Changed the log message to debug as suggested.

PR #562, opened 2020-05-17
Fixed "source setenv"

ulatekh:
While trying to track down some audio-corruption bugs (please see [here](https://bugs.kde.org/show_bug.cgi?id=410726)), I fixed "source setenv" (as described in docs/install.txt) by searching directories under src/modules for plugins (as it exists in the build tree).


ddennedy:
You have not described a problem this addresses. I have never had a problem with setenv.

ddennedy:
Are you using CMake? The configure and Makefile build system puts all built modules into `src/modules` and not into each subdirectory. I am not inclined to change this code to appease CMake behavior with respect to `setenv`. CMake is a newer and incomplete build system in MLT.

ulatekh:
The problem is that, when built with CMake, the compiled plugins don't get copied into the src/modules directory, but remain in their subdirectories. Changing the module search to one that searches subdirectories not only solves the problem, but is backwards-compatible with the non-CMake build and the installed version.

> CMake is a newer and incomplete build system in MLT.

Well, this will make it more complete. I have another change (not submitted yet) which improves the CMake build. I don't understand why changes that make the CMake build less incomplete are unwelcome.

vpinon:
Maybe Dan would prefer if we get CMake to output module libraries in the same folder, so no code change just for the build system...

ddennedy:
We should definitely explore a CMake option first. I do not think it good idea to change fundamental code in order to make a build process work, but I am not entirely opposed to it.

ulatekh:
I assume CMake can copy the build plugins to the parent directory locally, or export them there in the first place. I'll compare the CMake configuration with the existing makefiles and see what I can find.

ulatekh:
Please look at the change to src/modules/avformat/CMakeLists.txt, and tell me if you consider that method acceptable. If so, I'll revert the change to src/framework/mlt_repository.c, and modify the rest of the src/modules/*/CMakeLists.txt files.

ddennedy:
Does `set_target_properties(mltavformat PROPERTIES LIBRARY_OUTPUT_DIRECTORY ..)` break `make install`? I do not mind an intrusion on the cmake build at this early stage, and this change is much cleaner than that custom command. :-) I actually worry that custom command will stand the test of time and multiple platforms. Please help @vpinon

ulatekh:
On my system (Fedora Core 31), `set_target_properties(mltavformat PROPERTIES LIBRARY_OUTPUT_DIRECTORY ..)` doesn't break `make install`. I would imagine it's fine. It also matches better with what the non-CMake build does.

vpinon:
Hello,
Thanks for this change, I think this is the good approach.
(note: I don't play with melt inside build dir, I always "ninja install" before doing anything, so I didn't face the problem...)

ddennedy:
@ulatekh We would greatly appreciate if you can update all the modules' CMakeLists.txt with `set_target_properties(mltavformat PROPERTIES LIBRARY_OUTPUT_DIRECTORY ..)`. Thanks

ulatekh:
> We would greatly appreciate if you can update all the modules' CMakeLists.txt

Will do! Hopefully before the end of today.
I would greatly appreciate someone looking at [issue 566](https://github.com/mltframework/mlt/issues/566)...it's been a long time coming...

ulatekh:
OK, done.

I have other changes I'd like to do to the CMake build (e.g. strip all symbols except mlt_register() from the plugins), but I'll submit those as separate pull requests.

PR #569, opened 2020-05-22
Add export presets for screencasts and slide decks

austincbrooks:
Small file size while retaining perfect text clarity.

More information:
https://forum.shotcut.org/t/introducing-2-export-presets-for-video-conferencing-slide-decks-and-screencasts/18100

### DIFF
--- a/presets/consumer/avformat/Screencast-H264
+++ b/presets/consumer/avformat/Screencast-H264
@@ -18,5 +18,5 @@ f=mp4
 movflags=+faststart
 
 meta.preset.extension=mp4
-meta.preset.note=Small file size for screencasts, video conferences, and slide decks. Hardware encoding must be OFF.
+meta.preset.note=Small file size for screencasts, video conferences, and slide decks. Hardware encoding not recommended.
 meta.preset.name=Screencast (H.264)
--- a/presets/consumer/avformat/Screencast-H264
+++ b/presets/consumer/avformat/Screencast-H264
@@ -1,7 +1,6 @@
 pix_fmt=yuv420p
 vcodec=libx264
 vprofile=high
-level=51
 crf=27
 g=249
 bf=8
--- a/presets/consumer/avformat/Screencast-H264
+++ b/presets/consumer/avformat/Screencast-H264
@@ -1,0 +1,23 @@
+pix_fmt=yuv420p
+vcodec=libx264
+vprofile=high
+level=51
+crf=27
+g=249
+bf=8
+preset=veryfast
+b-pyramid=0
+refs=8
+weightb=1
+b_strategy=2
+rc-lookahead=25
+channels=1
+acodec=aac
+ar=32000
+ab=48k
+f=mp4
+movflags=+faststart
+
+meta.preset.extension=mp4
+meta.preset.note=Small file size for screencasts, video conferences, and slide decks. Hardware encoding must be OFF.
+meta.preset.name=Screencast (H.264)
--- a/presets/consumer/avformat/Screencast-HEVC
+++ b/presets/consumer/avformat/Screencast-HEVC
@@ -1,0 +1,17 @@
+pix_fmt=yuv420p
+vcodec=libx265
+crf=31
+g=249
+bf=8
+preset=slow
+x265-params=weightb=1:open-gop=0
+channels=1
+acodec=aac
+ar=32000
+ab=48k
+f=mp4
+movflags=+faststart
+
+meta.preset.extension=mp4
+meta.preset.note=Small file size for screencasts, video conferences, and slide decks. Hardware encoding must be OFF.
+meta.preset.name=Screencast (HEVC)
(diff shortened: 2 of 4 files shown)


presets/consumer/avformat/Screencast-H264
```diff
@@ -0,0 +1,23 @@
+pix_fmt=yuv420p
+vcodec=libx264
+vprofile=high
+level=51
```
ddennedy, line 4: level should be left out because it depends mainly on the resolution and frame rate.

presets/consumer/avformat/Screencast-H264
```diff
@@ -0,0 +1,23 @@
...
+crf=27
+g=249
+bf=8
+preset=veryfast
+b-pyramid=0
+refs=8
+weightb=1
+b_strategy=2
+rc-lookahead=25
+channels=1
```
ddennedy, line 14: I think it is bad to include channels because in Shotcut that Export parameter defaults from its Settings > Audio Channels. Choosing this preset could surprise people who want stereo.

presets/consumer/avformat/Screencast-H264
austincbrooks: Removed.

presets/consumer/avformat/Screencast-H264
austincbrooks: Hmm, I'm torn on this one.  These export presets are extremely specialized for media types that are 99% mono human speech.  Most video conferencing programs only capture mono audio.  Adding another (likely identical) channel of audio increases file size by 25% which defeats the project charter of "smallest file size possible" for the expected use cases.  I am not at all opposed to removing the `channels=1` flag if that's the convention, but I'm guessing more users could be surprised by larger file size than would be surprised by mono audio.  I defer to you.

ddennedy: A very large portion of Shotcut users are people recording games. I think a lot of them will see "Screencast" and think: oh, that is just like a game recording. These people will lose stereo output (not to mention poor audio quality due to the low bitrate).

presets/consumer/avformat/Screencast-H264
```diff
@@ -0,0 +1,22 @@
...
+acodec=aac
+ar=32000
+ab=48k
+f=mp4
+movflags=+faststart
+
+meta.preset.extension=mp4
+meta.preset.note=Small file size for screencasts, video conferences, and slide decks. Hardware encoding must be OFF.
```
ddennedy, line 21: I am not keen on the "Hardware encoding must be OFF." It is no longer true that all hardware encoding is inferior and produces large sizes. Sure, you can make that claim in response to a forum question about a particular encoder's large output but not sure I like it here.

presets/consumer/avformat/Screencast-H264
austincbrooks: Yeah, I can see the confusion.  The original name I gave these presets was "Slide Deck (H.264)" before switching to "Screencast" at the last minute.  Perhaps "Slide Deck" remains more representative.  Granted, these presets are a bit niche, so I'm okay with retracting them if they will cause more confusion than benefit.  I was specifically targeting video conference and distance education users with COVID stay-at-home arrangements, hence the low bitrate audio that's sufficient for human speech.

bmatherly: I like the idea of having these presets because I think the use case is common. Maybe the name needs some work. "Slide Deck" is not to bad. Other suggestions: "Presentation", "Slideshow", "High Compression".

presets/consumer/avformat/Screencast-H264
austincbrooks: Here's my dilemma... The H.264 version of the preset has customization that not even RTX can support.  The preset is heavily optimized for still images, giving a software encoder a clear advantage.  Non-RTX makes the preset almost pointless, as Quick Sync for example creates a file that is 5x as large with blurrier output at the same settings.  If somebody had hardware encoding on by default and used the preset with non-RTX, they would not be impressed with the results and would think the preset was ineffective.  Non-RTX still accounts for a lot of users today, especially those with laptop computers.

What about the phrase "Compare hardware encoding on and off to find the smallest file size."?

ddennedy: How about "Hardware encoding not recommended."?

presets/consumer/avformat/Screencast-H264
ddennedy: I agree to "Slide Deck" and then keep `channels=1`

presets/consumer/avformat/Screencast-H264
austincbrooks: I dig it.  Changes committed.

presets/consumer/avformat/Screencast-H264
austincbrooks: The name has been changed to "Slide Deck".  Thanks guys for your patience with me.

The output of these presets is around 10% the size of the YouTube preset.  For recordings of 2-hour class lectures and business meetings, the savings quickly become significant.

ddennedy:
Thank you for the contribution!

PR #573, opened 2020-05-30
Optimize Qimage producer

j-b-m:
Looking at the QImage producer, I found several bottlenecks that I could improve:

* Slightly faster check if image is readable on producer init (use QImageReader methods only to check image size and ensure it is readable instead of loading full image)

* When transferring the QImage data, use the QImage::Format_RGBA8888 or QImage::Format_RGB888 to copy the image in one pass instead of doing a pixel copy
* Disable caching for ttl=1 (image sequences with 1 image/frame). Caching scaled image does not make sense if we have an image sequence with 1 image/frame. Disable caching for those provides a noticeable speedup, in Kdenlive these sequences can now play at 20fps instead of 7fps on my machine.

Since this touches a rather important MLT component, comments are welcome.

### DIFF
--- a/src/modules/qt/qimage_wrapper.cpp
+++ b/src/modules/qt/qimage_wrapper.cpp
@@ -32,7 +32,6 @@
 #include <QtEndian>
 #include <QTemporaryFile>
 #include <QImageReader>
-#include <QDebug>
 
 #ifdef USE_EXIF
 #include <libexif/exif-data.h>
--- a/src/modules/qt/qimage_wrapper.cpp
+++ b/src/modules/qt/qimage_wrapper.cpp
@@ -32,6 +32,7 @@
 #include <QtEndian>
 #include <QTemporaryFile>
 #include <QImageReader>
+#include <QDebug>
 
 #ifdef USE_EXIF
 #include <libexif/exif-data.h>
@@ -51,11 +52,9 @@ extern "C" {
 static KComponentData *instance = 0L;
 #endif
 
-static void qimage_delete( void *data )
+static void qimage_delete( void *image )
 {
-	QImage *image = ( QImage * )data;
-	delete image;
-	image = NULL;
+	delete static_cast<QImage*>( image );
 #if defined(USE_KDE4)
 	delete instance;
 	instance = 0L;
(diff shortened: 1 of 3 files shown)


ddennedy:
I agree with the ideas but 2 things:

1. `ttl` defined does not imply it is an image sequence. `producer_qimage.count > 0` is a required check. In Shotcut, when you convert an image sequence back to still it does not clear `ttl`.

2. It is a heavy amount of change in some complex, stable code. Was it all necessary? How do we ensure there is no dangling pointer, bad pointer dereference, race condition, memory leak, etc.?

bmatherly:
I am interested in "Slightly faster check if image is readable on producer init" because the new slideshow feature in Shotcut will compel people to open many images at once.

It would be good if you could share some rough metrics of how much the performance is improved with these changes. For example, How much faster can 100 images be loaded with this change?

j-b-m:
@ddennedy I agree that this is a rather large patch which makes it difficult to track possible issues as it mixes some refactoring with other optimizations. I will maybe try to come up with several smaller patches that implement the various changes with only the necessary changes :

* One patch to disable caching on count > 0 and ttl == 1
* One patch to use the faster QImage::Format_RGBA8888 method to copy pixel data to MLT
* One patch to implement faster producer init (do not load image at init)
* Possibly a last patch to refactor some code (in current code we have several useless QImage * to MLT void* conversions).

@bmatherly In fact the speedup on producer init is very noticeable but it's because the image is not read anymore at that stage, only when requesting an image from it. I hope it's ok to proceed like this. Opening a project file with 75 images drops from 3 seconds to 0.3 seconds, but of course when displaying the images in monitor, they now need to be loaded. It also doesn't consume memory until the image is displayed.

sandsmark:
I might be missing something, but doesn't the overhead from flipping every pixel in the image to another format outweigh any cost of copying?

and don't you at least need to check the stride? I don't think it's safe to assume that the lines are aligned perfectly, I'd at least memcpy scanLine() by scanLine().

src/modules/qt/qimage_wrapper.cpp
```diff
@@ -180,98 +228,103 @@ int refresh_qimage( producer_qimage self, mlt_frame frame )
...
+			image_size = 3 * width * height;
+			self->format = mlt_image_rgb24;
+#if QT_VERSION >= 0x051300
+			scaled.convertTo( QImage::Format_RGB888 );
+#else
+			scaled = scaled.convertToFormat( QImage::Format_RGB888 );
+#endif
+		}
+		self->current_image = ( uint8_t * )mlt_pool_alloc( image_size );
+		memcpy( self->current_image, scaled.constBits(), image_size);
```
sandsmark, line 318: you should at least check that scaled.sizeInBytes() is same or bigger than image_size so you don't read out of bounds.

PR #582, opened 2020-06-06
QImage: speedup by disabling cache for 1 frame image sequence

j-b-m:
This is a cleanup of my previous #573 patch that contains only the necessary changes to disable caching on image sequences with ttl =1, and uses Qt's image format to transfer data to MLT frame instead of doing a per pixel copy.

Doing a quick comparative performance, I get these fps on a 1 frame image sequence in Kdenlive:
* without this patch: around 7fps
* disable caching only: around 15fps
* disabling caching and using Qt's image methods to copy data: 20fps

### DIFF
--- a/src/modules/qt/qimage_wrapper.cpp
+++ b/src/modules/qt/qimage_wrapper.cpp
@@ -286,11 +286,7 @@ void refresh_image( producer_qimage self, mlt_frame frame, mlt_image_format form
 		{
 			image_size = 3 * width * height;
 			self->format = mlt_image_rgb24;
-#if QT_VERSION >= 0x051300
-			scaled.convertTo( QImage::Format_RGB888 );
-#else
 			scaled = scaled.convertToFormat( QImage::Format_RGB888 );
-#endif
 		}
 		self->current_image = ( uint8_t * )mlt_pool_alloc( image_size );
 		memcpy( self->current_image, scaled.constBits(), image_size);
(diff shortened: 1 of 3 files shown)


src/modules/qt/qimage_wrapper.cpp
```diff
@@ -255,21 +265,42 @@ void refresh_image( producer_qimage self, mlt_frame frame, mlt_image_format form
...
+		}
+		self->current_image = ( uint8_t * )mlt_pool_alloc( image_size );
+		memcpy( self->current_image, scaled.constBits(), image_size);
+#else
+		// Convert scaled image to target format (it might be premultiplied after scaling).
+		scaled = scaled.convertToFormat( qimageFormat );
+		self->format = has_alpha ? mlt_image_rgb24a : mlt_image_rgb24;
+		image_size = mlt_image_format_size( self->format, self->current_width, self->current_height, NULL );
+		self->current_image = ( uint8_t * )mlt_pool_alloc( image_size );
+		image_size = mlt_image_format_size( self->format, self->current_width, self->current_height, NULL );
```
ddennedy, line 303: Line 303 can be removed.

src/modules/qt/qimage_wrapper.cpp
```diff
@@ -255,21 +265,42 @@ void refresh_image( producer_qimage self, mlt_frame frame, mlt_image_format form
...
+			image_size = 4 * width * height;
+			self->format = mlt_image_rgb24a;
+			scaled = scaled.convertToFormat( QImage::Format_RGBA8888 );
+		}
+		else
+		{
+			image_size = 3 * width * height;
+			self->format = mlt_image_rgb24;
+#if QT_VERSION >= 0x051300
+			scaled.convertTo( QImage::Format_RGB888 );
```
ddennedy, line 290: I realize the difference between `convertTo()` and `convertToFormat()`, but does it provide any advantage? Otherwise, it clutters the code, and the block just above this does not have the build option to use `convertTo()`.

ddennedy:
Other than these minor comments, this looks good to me. 👍

src/modules/qt/qimage_wrapper.cpp
vpinon: @j-b-m good you removed this check: 5.13 should have been 0x050D00 (or better QT_VERSION_CHECK(5,13,0)

PR #583, opened 2020-06-06
Add resizer filter using Qt's image methods

j-b-m:
This is an exact copy of the core/filter_resize.c normalizing filter, but copying the source image over a new blank image is done with Qt's QImage methods instead of pixel copy for rgb24a.
This brings a 40% speedup when compositing an image that is smaller than profile over a color or video clip

### DIFF
--- a/src/modules/qt/Makefile
+++ b/src/modules/qt/Makefile
@@ -16,7 +16,8 @@ CPPOBJS = common.o \
 	kdenlivetitle_wrapper.o \
 	producer_qtext.o \
 	transition_qtblend.o \
-	filter_qtblend.o
+	filter_qtblend.o \
+	filter_qresize.o
 
 ifdef USE_QT_OPENGL
 	CPPOBJS += consumer_qglsl.o
--- a/src/modules/qt/filter_qresize.cpp
+++ b/src/modules/qt/filter_qresize.cpp
@@ -1,0 +1,344 @@
+/*
+ * filter_resize.c -- resizing filter
+ * Copyright (C) 2003-2014 Meltytech, LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include "common.h"
+#include <framework/mlt.h>
+#include <framework/mlt_log.h>
+#include <QPainter>
+#include <QPainterPath>
+#include <QString>
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <math.h>
+
+static uint8_t *resize_alpha( uint8_t *input, int owidth, int oheight, int iwidth, int iheight, uint8_t alpha_value )
+{
+	uint8_t *output = NULL;
+
+	if ( input != NULL && ( iwidth != owidth || iheight != oheight ) && ( owidth > 6 && oheight > 6 ) )
+	{
+		uint8_t *out_line;
+		int offset_x = ( owidth - iwidth ) / 2;
+		int offset_y = ( oheight - iheight ) / 2;
+		int iused = iwidth;
+
+		output = (uint8_t*)mlt_pool_alloc( owidth * oheight );
+		memset( output, alpha_value, owidth * oheight );
+		offset_x -= offset_x % 2;
+
+		out_line = output + offset_y * owidth;
+		out_line += offset_x;
+
+		// Loop for the entirety of our output height.
+		while ( iheight -- )
+		{
+			// We're in the input range for this row.
+			memcpy( out_line, input, iused );
+
+			// Move to next input line
+			input += iwidth;
+
+			// Move to next output line
+			out_line += owidth;
+		}
+	}
+
+	return output;
+}
+
+static void resize_image( uint8_t *output, int owidth, int oheight, uint8_t *input, int iwidth, int iheight, int bpp, mlt_image_format format, uint8_t alpha_value )
+{
+	// Calculate strides
+	int istride = iwidth * bpp;
+	int ostride = owidth * bpp;
+	int offset_x = ( owidth - iwidth ) / 2 * bpp;
+	int offset_y = ( oheight - iheight ) / 2;
+	uint8_t *in_line = input;
+	uint8_t *out_line;
+	int size = owidth * oheight;
+	uint8_t *p = output;
+
+	// Optimisation point
+	if ( output == NULL || input == NULL || ( owidth <= 6 || oheight <= 6 || iwidth <= 6 || iheight <= 6 ) )
+	{
+		return;
+	}
+	else if ( iwidth == owidth && iheight == oheight )
+	{
+		memcpy( output, input, iheight * istride );
+		return;
+	}
+
+	if ( format == mlt_image_rgb24a )
+	{
+#if QT_VERSION >= 0x050200
+		QImage src(iwidth, owidth, QImage::Format_RGBA8888);
+		convert_mlt_to_qimage_rgba( input, &src, iwidth, iheight );
+		QImage dest( output, owidth, oheight, QImage::Format_RGBA8888 );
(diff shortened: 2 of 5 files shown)


j-b-m:
10 lines are added to the original code (line 92-102) for this.

ddennedy:
Hmm, before I get into a couple of minor cleanups I spotted, I wonder how this could negatively affect the GTK+ Flowblade application? Maybe it has worked around the qimage producer priority in `core/loader.dict` by explicitly selecting pixbuf, but it would not be able to do the same for a filter in `core/loader.ini`.
Also, I wonder what sort of speed would be gained by adding mlt_slices to the core/resize filter (could benefit all formats).

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+	else if ( iwidth == owidth && iheight == oheight )
+	{
+		memcpy( output, input, iheight * istride );
+		return;
+	}
+
+	if ( format == mlt_image_rgb24a )
+	{
+#if QT_VERSION >= 0x050200
+		QImage src(iwidth, owidth, QImage::Format_RGBA8888);
```
bmatherly, line 93: There is no need to use the the convert_* functions from common.cpp since you have already guarded with with a QT version check that removes the need for backwards compatibility. The way this is written, src will be initialized twice. Once on line 93, and then again in convert_mlt_to_qimage().

I would suggest something like:

```
QImage src(input, iwidth, iheight, QImage::Format_RGBA8888);
QImage dest(output, owidth, oheight, QImage::Format_RGBA8888 );
dest.fill(Qt::transparent);
QPainter painter(&dest);
painter.drawImage(QPoint((owidth - iwidth) / 2, (oheight - iheight) / 2), src);
painter.end();
```

src/modules/qt/Makefile
```diff
@@ -16,7 +16,8 @@ CPPOBJS = common.o \
 	kdenlivetitle_wrapper.o \
 	producer_qtext.o \
 	transition_qtblend.o \
-	filter_qtblend.o
+	filter_qtblend.o \
+	filter_qresize.o
```
bmatherly, line 20: I would suggest the name change to "qtresize" for consistency. I regret that I did not name "qtext" "qttext". I thought I was being clever by reusing the "t" at the time.

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+	// No change, return input
+	return input;
+}
+
+/** Do it :-).
+*/
+
+static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
+{
+	int error = 0;
```
bmatherly, line 197: Instead of duplicating so much code from filter_resize, could you detect in this function if QImage can be used based on the format? If QImage can be used, then use it. If not, then "fall back" to an encapsulated instance of filter_resize.

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+#include <QPainter>
+#include <QPainterPath>
+#include <QString>
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <math.h>
+
+static uint8_t *resize_alpha( uint8_t *input, int owidth, int oheight, int iwidth, int iheight, uint8_t alpha_value )
```
bmatherly, line 32: Could QImage::Format_Alpha8 or QImage::Format_Grayscale8 be used for this case?

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+	{
+		return;
+	}
+	else if ( iwidth == owidth && iheight == oheight )
+	{
+		memcpy( output, input, iheight * istride );
+		return;
+	}
+
+	if ( format == mlt_image_rgb24a )
```
bmatherly, line 90: I think it would also be trivial to handle mlt_image_rgb24 using QImage::Format_RGB888

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+extern "C" {
+
+/** Constructor for the filter.
+*/
+
+mlt_filter filter_qresize_init( mlt_profile profile, mlt_service_type type, const char *id, char *arg )
+{
+	mlt_filter filter = mlt_filter_new();
+	fprintf(stderr, "===== TRYING TO INIT QRESIZER\n");
```
bmatherly, line 329: Leftover debug

src/modules/qt/filter_qresize.cpp
```diff
@@ -0,0 +1,344 @@
...
+}
+
+extern "C" {
+
+/** Constructor for the filter.
+*/
+
+mlt_filter filter_qresize_init( mlt_profile profile, mlt_service_type type, const char *id, char *arg )
+{
+	mlt_filter filter = mlt_filter_new();
```
bmatherly, line 328: If this filter will be compiled for all versions of QT, then a version check here could return null unless QT_VERSION >= 0x050200.

bmatherly:
> I wonder how this could negatively affect the GTK+ Flowblade application?

@ddennedy can you explain this question better? Why would this filter not work for flowblade? If Qt is not present, won't the loader fall back to the next option?

ddennedy:
> can you explain this question better?

Flowblade is most often delivered through distributions that use a shared MLT build that includes the qt module, and these packages will pull in Qt as a dependency even if not always used. Then, the service's init function initializes a QApplication and _bleep_ hits the fan IIRC.

bmatherly:
> Then, the service's init function initializes a QApplication and _bleep_ hits the fan IIRC.

I have a couple of ideas for this:

1) Add a new function to mlt_repository to allow the application to blacklist certain modules. In this case, Flowblade could blacklist "qt" before the repository is initialized and then the qt services will never become available.

2) Add a new function to mlt_factory to allow the application to blacklist certain services. IN this case, Flowblade could blacklist all the services in the qt directory. At run-time, if a services is requested from mlt_factory that is in the blacklist, NULL will be returned instead.

Let me know if either of these ideas are interesting. I would be happy to mock something up.

ddennedy:
> allow the application to blacklist certain services

I seemed to recall there is a way to this and confirmed it - by pruning mlt_repository_producers() after factory init.

In any case, I think I am going to reject this pull request and ask that the simple code in the existing filter be improved instead of adding  more code. Besides mlt_slices there is also Duff's device, OpenMP, and SIMD asm.

bmatherly:
> I seemed to recall there is a way to this and confirmed it - by pruning mlt_repository_producers() after factory init.

Oh, that is good to know. So this would be a non-factor for Flowblade then.

> In any case, I think I am going to reject this pull request and ask that the simple code in the existing filter be improved instead of adding more code. Besides mlt_slices there is also Duff's device, OpenMP, and SIMD asm.

That is an understandable position, but unfortunate in some ways:
* The alternative suggestions you offer add complexity to the code and makes it harder to maintain. Of course, sometimes that complexity is worth the tradeoff.
* I do not know that the current contributors have a lot of experience with OpenMP and SMID optimization
* I believe that there is a lot of "low hanging" optimization waiting to be harvested from the existing optimized features in QT. I think we could easily reduce system lag in many areas by writing QT variants of many modules. We already have the framework to select the best service from a list of available services.

Did you see my suggestion to make the qtresize filter a wrapper around the current resize filter to avoid all the code duplication? That might be a good compromise?

jliljebl:
Dan remembers correctly, I had to add lines below to not get Qt stuff initialized but it works just fine now:

    repo.producers().set('qimage', None, 0)
    repo.producers().set('qtext', None, 0)
    repo.producers().set('kdenlivetitle', None, 0)

j-b-m:
I personnaly do not feel very confident to implement multithreading in the existing resizer.
However, I think some optimizations are really needed in MLT, and the normalizers are an interesting hotspot since they are used a lot. If the use of Qt is not wanted in core parts of the framework - which I can understand - I hope that someone else can jump in to improve the resizer filter because it would be a shame to stay with code that has a very poor performance - that hurts all users of MLT.

src/modules/qt/filter_qresize.cpp
ddennedy: Please test this code (whichever approach) with a variety of input and output sizes to avoid a regression, which I just found like this https://github.com/mltframework/mlt/commit/27361ebadedaa299a97b1a638fd7bc5675503a6c?diff=unified#r39869941

ddennedy:
I just pushed commit a955cd43 to dramatically increase the speed of core/resize without SIMD or multithread. Here was my test case with a debug build where test.png is a file that is 447x646 with alpha:

`time melt -verbose -profile atsc_1080p_24 test.png -consumer null mlt_image_format=rgb24a terminate_on_pause=1 out=1000`

before:  15.125s
after:    0.920s

I checked out this branch and tested with qresize: 6.294s

By varying `mlt_image_format` on the command line, I can test the others:
rgb24:
before: 9.018s
after:  0.913s

yuv422
before: 6.734s
after:  3.602s

PR #598, opened 2020-07-28
Add pillar echo filter

bmatherly:


ddennedy:
As I am testing this, every pillar box seems much more purple than I expect. Here are some examples:
image
image

I can accept the color for her sweatshirt, but why is her hair so purple? Why did the tower and ground become purple?

bmatherly:
Thanks for testing. I should have tested harder after a last minute cleanup. Should be fixed now.

PR #621, opened 2020-11-20
Hardware acceleration support for AVFormat producer

gabrieljablonski:
This makes use of Intel's video acceleration API (VA-API), which means the FFmpeg being used must have it available (`--enable-libva` flag for older versions, for newer versions simply having the libva installed in the system when compiling should be enough). In my tests between a couple different machines, I've noticed improvements anywhere from 15% to 30% in the CPU usage when decoding H.264 videos.

Most changes are dependent on the `USE_VAAPI` flag being set, but the most significant one overall, which does not depend on the flag being set, is the use of the `avcodec_send_packet()/avcodec_receive_frame()` flow instead of the deprecated `avcodec_decode_video2()`. An improvement to this would be to check the version of the Libav being used and switch seamlessly between the old and the new methods.

One last change that could be also added is a `./configure`  `--enable-vaapi` flag that defines the `USE_VAAPI` flag in the source, avoiding the need to change the source code manually.


ddennedy:
Thank you for the contribution and sorry the build mis-flagged as failed. That script needs a fix to suppress the errors when testing if this is mlt master branch to docker push.
I started working on some ffmpeg deprecations in a branch including `avcodec_decode_video2()` while working to integrate libdav1d for AV1. Of course, mine is different, and your change is not working for that (I think it goes into infinite loop).

I am using an Intel Core i7-9700K with integrated graphics. I tested with a 4K60 H.264 clip, and the performance was horrible. Even a 720p30 HEVC clip was a little worse. A 1080p24 H.264 clip was slightly better with VA-API except for seeking, which was very slow. Certainly there are some hardware decoders with resolution limits but also it can be worse below the limit probably due to RAM transfer speeds. How can we choose a threshold and disable using VA-API? Likewise, we need a way to opt out of using VA-API in a build at run time.

How can we deal with resource limits on the number of simultaneous decoders? MLT uses caching to hold open a limited number of files, libav contexts, and avformat producers to prevent thrashing. Thrashing here means continually closing, releasing, reopening, and reinitializing media files and decoder objects. For example, a person steps frames backwards and forwards between 2 clips. Another is transitions, multitrack blending, and combinations of them. For each frame going through those, multiple media files are used, and we do not want to continually close and reopen files/decoders as we process those. The default is max(4, 2 * #tracks). I expect there to be fairly low limits on these resources. Also, when testing with the 1080p24 clip above with multitrack blending across 4 tracks, I noticed the performance was much worse, probably due to throughput. So, like resolution, there is a probably a limit on number per frame for performance reasons too.

I suffered a crash or deadlock when trying to use MJPEG AVI. So, we need a way to block certain codecs the implementation accepts but which are problematic, even if that is due to the MLT integration. How can we do that?

Unfortunately, I suspect there is a lot of variability to these issues and limits per system and version.

gabrieljablonski:
I should've probably posted these: I've tested on an i7-6700T and an i7-6700 with Libva version 1.8.3 on both.

The implementation is probably not going to work properly on 8th gen and newer, given they don't support Libva 1.x, only 2.x (though I haven't tested it yet), so some other changes in the source might be needed to support 2.x, and switch between the two implementations.

As for your other concerns, such as thrashing, I'm not familiar enough with MLT's inner workings to think of a way to solve those issues. All I can say is that with the testing I've done so far, I haven't yet run into any performance issues.

If this is something that is not of interest right now, I don't mind the PR being closed and maybe revisited at a later date when there's demand for it. 

If it is of interest, I suggest changing the PR from "Open" to "Draft" at least until some more testing is done, either by me (I intend to start working with several other CPU gens in the near future) or by some other member of the community.

ddennedy:
We are not disinterested, but it must be something someone chooses to use explicitly at run time considering the hard and soft limitations. More work is really needed probably against libva directly to dig out the device's hard limits and integrate them with the framework's multitrack and cache classes in order to enforce the limits. That is not easy, and vaapi only serves a very niche user base (Linux users primarily using only distro packages since we learned from experience that vaapi in portable binaries provides horrible compatibility due to version mismatches). Combine these limits with the fact that many hwaccel decode implementations are not using unified memory and require expensive memory transfer. So, real benefits come when combined with GPU video processing via OpenGL/Movit using textures and avoiding memory transfers. Now, how to do that with all the different APIs for the different operating systems? And what if not using Movit and adopting some other GPGPU and multicore framework? I know that I am extrapolating very much. However, I am trying to give you some insight into my thinking around this whole area and with respect to priorities especially when we have highly-optimized, multi-threaded CPU decoders that are super reliable.

I do not understand how the libva version can affect the integration of FFmpeg vaapi hwaccel into MLT, and I do not see anything in FFmpeg's `./configure --help` about libva versions, but I have not looked into this very deeply.

We can integrate this for specialized applications who want to use it explicitly now. You can do that using an environment variable if a flag property does not work. However, I think the change to `avcodec_send_packet()/avcodec_receive_frame()` needs more work. I can push my branch to address deprecations and get dav1d working for you to review and integrate.

src/framework/mlt.h
```diff
@@ -60,6 +60,8 @@ extern "C"
 #include "mlt_version.h"
 #include "mlt_slices.h"
 
+#define USE_VAAPI 0
```
ddennedy, line 63: This does not belong here. Move it into producer_avformat.c.

src/modules/avformat/producer_avformat.c
```diff
@@ -138,9 +143,19 @@ struct producer_avformat_s
...
+	AVBufferRef* hw_device_ctx;
+	AVFrame* sw_video_frame;
+#endif
 };
 typedef struct producer_avformat_s *producer_avformat;
 
+#ifdef USE_VAAPI
+#define HW_PIX_FMT AV_PIX_FMT_VAAPI
+#define HW_DEVICE_TYPE AV_HWDEVICE_TYPE_VAAPI
+#define HW_DEVICE "/dev/dri/renderD128"
```
ddennedy, line 156: Can these be properties that match with ffmpeg usage: `hwaccel`, `hwaccel_device`, and `hwaccel_output_format` perhaps with some defaults when unspecified per the `hwaccel`? If these would need to be supplied during producer initialization, they can be supplied as query string arguments on the resource property. `hwaccel` could be the way to explicitly choose to use this.

src/modules/avformat/producer_avformat.c
```diff
@@ -623,6 +638,10 @@ static enum AVPixelFormat pick_pix_fmt( enum AVPixelFormat pix_fmt )
 #if defined(FFUDIV)
 	case AV_PIX_FMT_BAYER_RGGB16LE:
 		return AV_PIX_FMT_RGB24;
+#endif
+#ifdef USE_VAAPI
+	case AV_PIX_FMT_VAAPI:
```
ddennedy, line 643: Can this be expanded for hwaccel support beyond just VAAPI? If so, then we change the define macro name to `USE_HWACCEL`

src/modules/avformat/producer_avformat.c
```diff
@@ -1833,7 +1857,35 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
...
+						{
+							// av_strerror( ret, errstr, 1000 );
+							// mlt_log_error( NULL, "avcodec_send_packet() failed %s\n", errstr );
+						}
+						if ( ( ret = avcodec_receive_frame( codec_context, self->video_frame ) ) < 0 ) 
+						{
+							// av_strerror( ret, errstr, 1000 );
+							// mlt_log_error( NULL, "avcodec_receive_frame() failed %s\n", errstr );
+						}
+					} while ( ret == AVERROR( EAGAIN ) );
```
ddennedy, line 1887: Some decoders need multiple packets before they will return a frame. I will share my branch with you where I have adopted this API.

ddennedy (changes requested):

ddennedy:
Here is the branch I mentioned: https://github.com/mltframework/mlt/compare/ffmpeg_deprecations

gabrieljablonski:
I very much understand and share your concerns with pushing such a niche use case into production without being 100% sure all bases are covered.

Regarding the usage of `avcodec_send_packet()/avcodec_receive_frame()`, I admit my implementation ended up not being too elegant (I used some FFmpeg code as reference and it wasn't that similar to MLT's), but I ended up settling to leave it as is since it seemed to be working without issue, though I will surely integrate your modifications.

Overall, given your great understanding of the codebase, I trust your decision in how to proceed with this going forward.
I'm working on a fork with some other more specific use cases, and I intend to put forward any generic solutions that I find might be interesting for the community.

Thanks for the feedback!

gabrieljablonski:
> Here is the branch I mentioned: https://github.com/mltframework/mlt/compare/ffmpeg_deprecations

I've noticed a very significant memory leak regarding [this frame allocation](https://github.com/mltframework/mlt/blob/bfa121263e135b4597b1a3b8e34353c63b3608d6/src/modules/avformat/producer_avformat.c#L1715). It's very noticeable when playing a short video on loop or when loading and playing a new video multiple times.

It also happened in my implementation, and unrefing the frame under `exit_get_image:` seems to fix it.

ddennedy:
> I've noticed a very significant memory leak regarding this frame allocation. It's very noticeable when playing a short video on loop or when loading and playing a new video multiple times.

I am a little surprised considering the old code did not need it and based on the decode_video.c example. Nevertheless, I reproduced it, confirmed the master branch is not affected, and verified adding it keeps memory flat. Thanks.

Will you be re-basing your branch upon mine and addressing the review comments? If so, I can test it on other hwaccel including other operating systems. Though I do not how much special code is needed to support different ones, and how much time to give to that. At the very least we can make it more agnostic as there is very little VAAPI-specific in your change. Then, I or others expand upon it over time.

gabrieljablonski:
I've applied your changes. The only thing missing is setting the hardware parameters via query string ([lines 845-850](https://github.com/mltframework/mlt/pull/621/files#diff-dd4dd7f8b6c28984ac8e670241eb3a0e5ddc65efa5c2784a7267a018d6ff6811R845-R850)), which I'm not sure which is the proper way to do, so if you can point me to an example that does something similar I'll fix it.

gabrieljablonski:
> Here is the branch I mentioned: https://github.com/mltframework/mlt/compare/ffmpeg_deprecations

By an absolutely amazing stroke of (bad?) luck, I've managed to identify another bug, but this one seems a bit peskier and I haven't found the solution, though I imagine it won't be too hard to fix.

To reproduce it you'll need a video in which the video stream index is 1 (maybe anything other than 0?). Seeking to about the end of the video (100~200 frames to end) and playing makes the producer enter an infinite loop under the [`av_read_frame()` while loop](https://github.com/mltframework/mlt/blob/42622ceccab0fa60fc48170a44b706be3733e4fc/src/modules/avformat/producer_avformat.c#L1718).

You should be able to reproduce it by remapping an already existing sample in the "wrong" order (audio track first):

`ffmpeg -i sample.mp4 -vcodec copy -acodec copy -map 0:a:0 -map 0:v:0 sample-remapped.mp4`

The video seems to always enter the infinite loop right about the time the correctly mapped sample would log `avcodec_send_packet failed with -541478725` (maybe some EOF related error? didn't manage to find which error this code is supposed to be referencing).

ddennedy:
> The only thing missing is setting the hardware parameters via query string

See the function `parse_url()` in `producer_avformat.c`. More specifically, find the comment `// generic demux/device option support`. That is where query string parameters are copied to an AVDictionary that is then supplied to the AVFormatContext on `avformat_open_input()`.

gabrieljablonski:
I don't quite understand what [this `AVFormat`](https://github.com/mltframework/mlt/pull/621/commits/85bf9f21c9ed32f2d098b26b89bb92c9b2754831#diff-dd4dd7f8b6c28984ac8e670241eb3a0e5ddc65efa5c2784a7267a018d6ff6811R551) is being used for, as it returns as NULL when using the `file://` protocol for the resource being loaded, so I ended up leaving it commented out to test out the query string. Maybe I'm misunderstanding what `protocol` means in this context.

When that's figured out, the hardware acceleration should be properly being set by loading a resource as such:
```
<protocol>://<resource-uri>?hwaccel=<hwaccel-type>
```

The configuration for other hardware types can be set [here](https://github.com/mltframework/mlt/pull/621/commits/85bf9f21c9ed32f2d098b26b89bb92c9b2754831#diff-dd4dd7f8b6c28984ac8e670241eb3a0e5ddc65efa5c2784a7267a018d6ff6811R853-R856). I currently only have access to VA-API devices, so I can't test out the other types.

ddennedy:
Yes, `parse_url()` does not process a query string for a file resource as it stands. It is intended more for device and network protocols. Trying to make file resources use it could easily break on file names that contain a question mark, and I do not want to require them to be escaped (would not be backwards compatible). Maybe we should require our own special protocol `hwaccel` that we can ignore (does not call `av_find_input_format()` on it) but still allows query string processing (and requires escaping '?' in the file name).

ddennedy:
> Seeking to about the end of the video (100~200 frames to end) and playing makes the producer enter an infinite loop under the av_read_frame() while loop.

This is fixed in my branch. It was a simple EOF handling problem with the new state machine.

gabrieljablonski:
> This is fixed in my branch. It was a simple EOF handling problem with the new state machine. Thanks for finding it!

Good to know. 

Regarding the query string stuff, do you think simply using the `file://` prefix would be more straight forward (possibly even useful for other things that might use query strings in the future) than using a special one such as `hwaccel://`?

I won't be able to do any testing and make sure it works properly until Monday, but f37b7c3 has the rough idea of how it would work. Functionality remains absolutely the same for other protocols, but when using the `file://` prefix, it handles escaped question marks when searching for the query string. 

`file://<file-path>?hwaccel=<hwaccel-type>` would be the syntax.

ddennedy:
Changing the `file` protocol creates a backwards compatibility problem. Today, avformat producer can accept a `file://` prefix including an unescaped question mark, and this change unfortunately breaks it. However, I am on board with the idea of creating a special protocol that is generic and highly unlikely to ever be in libavdevice or libavformat!

Today, I found another problem with the new requirements of `avcodec_send_packet()/avcodec_receive_frame()`. It is incompatible with the frame threading in mlt_consumer. Frame threading occurs when `abs(real_time) > 1`. I get warnings `[swscaler @ 0x7fd1d003e1c0] bad src image pointers` when I use it. I have not fully diagnosed all the problems, but I know it may cause calls to `producer_get_image` out-of-order, and this will cause havoc with the state machine around using these new APIs and the new field `producer_avformat->video_send_result`. There is a `video_mutex` to protect some things, but it does not guarantee sequential access. There is nothing in MLT that can do that. In theory, I could cache frames decoded before reaching the target frame, but that is additional caching on top of existing pixel format and colorspace converted images. And things are already memory hungry enough in MLT. I am working away from using frame threading and more into slice processing and OpenMP when it comes to CPU. However, I am not sure I am ready to disable frame threading yet, and that may delay merging all of this. I would have to convert many more things to slice threading, and make performance comparisons, etc. All of this will not merge before the next release anyways, which is sometime in the next week or two.

ddennedy:
> I am on board with the idea of creating a special protocol that is generic

filesystem?
fs?

ddennedy:
Sorry, clicked wrong button!

gabrieljablonski:
> Frame threading occurs when `abs(property("real_time")) > 1`. I get warnings `[swscaler @ 0x7fd1d003e1c0] bad src image pointers` when I use it.

I don't know if it's related, but this information might help in figuring something out. I also had this warning show up consistently when using VA-API. I'm not sure how or why it worked, but the solution was to use the `format` from `self->video_frame` instead of from `codec_params` when calling `convert_image()` [here](https://github.com/gabrieljablonski/mlt/blob/f37b7c3bf1cff0fb3176e9c7a8f023f4ee7c5d27/src/modules/avformat/producer_avformat.c#L1749) and [here](https://github.com/gabrieljablonski/mlt/blob/f37b7c3bf1cff0fb3176e9c7a8f023f4ee7c5d27/src/modules/avformat/producer_avformat.c#L1964).

gabrieljablonski:
With 580a501, loading a file with `fs://<path-to-file>?hwaccel=<hwaccel-type>` with a valid type will enable hardware acceleration.
When using `fs://` (which can be easily changed to anything else), it is assumed special characters (such as `?` are escaped with backslashes). `parse_url()` will return the filename without any backslashes (maybe change it so only backslashes followed by `?` 
 are removed?).

I believe that after solving the issue with the `bad src image pointers` warning, everything else should be working fine, and we'll only be missing testing with other hwaccel types (and more testing with VA-API).

ddennedy:
>  I also had this warning show up consistently when using VA-API. I'm not sure how or why it worked, but the solution was to use the format from self->video_frame instead

I tried this, but it did not help. In addition to the warning there are image artifacts.

ddennedy:
I installed an NVIDIA card in Linux box and got it working with NVDEC API. The performance is better than my CPU (not VA-API) with the 4K60 H.264 test file. I also added support for hwaccel_device as a query string parameter. Commits for these are pushed to https://github.com/mltframework/mlt/commits/producer-avformat-hwaccel

Hmm, I see in your "reapply" commit that you reformatted a bunch of my code changes by adding spaces inside parenthesis. The code style is intentionally gradually changing. I am not too concerned about which, but extra spaces and braces on new lines are no longer required FYI.

gabrieljablonski:
> I installed an NVIDIA card in Linux box and got it working with NVDEC API.

Good to know it only took changing the parameters and nothing else for it to work. As I said, sometime in the near future I'll be doing some testing with different CPUs and maybe some dedicated GPUs to try and track down any compatibility issues that may arise.

> The code style is intentionally gradually changing.

Thanks for the info. When opening the PR I thought it would be a good idea to check it there were any style guides being followed, and I came across [this page](https://mltframework.org/docs/codingstyleguide/), so I ended up just following it, but I will keep this in mind now.

gabrieljablonski:
Ran into another bug (reproducible on [`ffmpeg_deprecations`](https://github.com/mltframework/mlt/compare/ffmpeg_deprecations)). It seems that whenever `avcodec_send_packet()` returns `EOF` (which happens a bit before the last frame, is this really expected behavior?), the frames either stop being decoded or they stop being sent to the consumer (not sure which). To verify this, I played a video frame by frame and the last 5 frames (approximately) were the fifth-to-last frame  (approximately) being repeated. 

Although this seems related to the infinite loop bug, reverting the fix with the mutex unlock didn't seem to work.

gabrieljablonski:
Found a [related issue](https://github.com/microsoft/FFmpegInterop/issues/217) on a Microsoft repo. 
The solution might be related to "draining mode" as reported in the [FFmpeg docs](https://www.ffmpeg.org/doxygen/3.4/group__lavc__encdec.html#MSearchField:~:text=End%20of%20stream%20situations.%20These%20require,has%20to%20be%20reset%20with%20avcodec_flush_buffers().):

> End of stream situations. These require "flushing" (aka draining) the codec, as the codec might buffer multiple frames or packets internally for performance or out of necessity (consider B-frames). This is handled as follows:
> 
> Instead of valid input, send NULL to the avcodec_send_packet() (decoding) or avcodec_send_frame() (encoding) functions. This will enter draining mode.
> 
> Call avcodec_receive_frame() (decoding) or avcodec_receive_packet() (encoding) in a loop until AVERROR_EOF is returned. The functions will not return AVERROR(EAGAIN), unless you forgot to enter draining mode.
> 
> Before decoding can be resumed again, the codec has to be reset with avcodec_flush_buffers().

ddennedy:
> The solution might be related to "draining mode",

The latter part is there with calling `avcodec_receive_frame()`, but we need to figure out when to call `avcodec_send_packet()` with NULL instead of skipping the call.

Please be prepared that this might need to sit in a branch for a long time because the incompatibility with frame threading is currently a deal-breaker. There is much performance-related work needed elsewhere including external things like frei0r or its replacement before I can abandon that. The user experience for the primary applications of Kdenlive, Flowblade, and Shotcut are highest priority. I will likely create a new branch to work on some ffmpeg deprecations that does not include the decode APIs.

gabrieljablonski:
> Please be prepared that this might need to sit in a branch for a long time because the incompatibility with frame threading is currently a deal-breaker.

Not a problem. Aside from this last bug, the things I'm working on don't require much more optimization.

> we need to figure out when to call `avcodec_send_packet()` with NULL instead of skipping the call.

I'll try to figure this out and keep you posted.

vpinon:
> incompatibility with frame threading is currently a deal-breaker [...] for the primary applications of Kdenlive, [...]

Hello,

You are coming to a topic I would want to raise for months (or more): reliability of multithreading in MLT processing.

In Kdenlive we have always marked the multithreading option in the settings UI as experimental, and in the last versions we also moved the export (render) option to the hidden-by-default panel. This is because, even if working well most of the time, we regularly have reports about video artifacts, very long render times or even crashes, that are solved when moving back to 1 processing thread. As you said, I suspect most of the time it comes from plugin effects (frei0r?).

We never took the time to properly track what was precisely at fault. Actually, it was my proposal for 2020 Kdenlive Sprint (which couldn't happen): have a day dedicated to identifying multithreading errors. And work with you on that. Parallel programing is a difficult topic, at least we are (I am) not very good at this.

Efficient multithreading is a performance boost probably easier to get than having a whole pipeline on GPU.
But this is a different topic, sorry for hijacking this thread.

In short: maybe breaking current multithreading would not be so disastrous on Kdenlive side ;)

ddennedy:
I will reply in more detail to @vpinon but I want to add that I am not totally happy with frame-threading either, and there are plans to move away and expand usage of slice-threading and possibly OpenMP. When I wrote "long time" I mean basically not within several weeks, but it is hard to say when. What I can tell you is that frame-threading today is often a big boost if not always reliable, and I am not willing to accept the guaranteed harsh criticism of taking that away without improvement elsewhere to compensate.

bmatherly:
> What I can tell you is that frame-threading today is often a big boost if not always reliable, and I am not willing to accept the guaranteed harsh criticism of taking that away without improvement elsewhere to compensate.

I am glad to hear you say this. I have had great success with frame threading in other video products. I volunteer to help debug frame threading issues if they are reproducible. I think that a final solution will include a mix of frame threading and slice threading.

gabrieljablonski:
bfa9cac fixes the usage of the draining mode (most of it was indeed already implemented). Unfortunately, I was unable to make it work with the infinite loop fix (missing mutex unlock). 

Since the infinite loop can be avoided (seemingly) by simply not playing media in which the video track is not 0, and dropping the last frames is a much bigger issue for me right now, I chose to just leave it commented out until we can figure out the proper way to do it after these changes (should not be that hard).

ddennedy:
> the incompatibility with frame threading is currently a deal-breaker

I will make an experiment to convert caching in this producer to the unconverted decoded frames - including ones intermediate to the requested position - to see if it resolves this frame-threading problem.

ddennedy:
I pushed a commit to my branch to fix the "bad src image pointers" I was seeing with frame threading. I was working on AVFrame caching, and that alone did not fix this. Then, after I made the fix and did more testing, I found I no longer need the AVFrame caching, which was currently ugly and having some other problems. The existing caching was working and just as fast. So, that is unblocked until further notice.

src/modules/avformat/producer_avformat.c
```diff
@@ -1825,8 +1825,8 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 						{
 							// Try to reconnect to live sources by closing context and codecs,
 							// and letting next call to get_frame() reopen.
+							mlt_service_unlock( MLT_PRODUCER_SERVICE( producer ) );
```
gabrieljablonski, line 1828: Since `mlt_service_unlock()` is already called under `exit_get_image:` (L2082), would calling it here be problematic?

src/modules/avformat/producer_avformat.c
```diff
@@ -1907,6 +1907,7 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 								if( transfer_data_result < 0 ) 
 								{
 									mlt_log_error( MLT_PRODUCER_SERVICE( producer ), "av_hwframe_transfer_data() failed %d\n", transfer_data_result );
+									mlt_service_unlock(MLT_PRODUCER_SERVICE(producer));
```
gabrieljablonski, line 1910: Same as above.

src/modules/avformat/producer_avformat.c
ddennedy: `prepare_reopen()` immediately calls `mlt_service_lock()` but now I see it also calls `mlt_service_unlock()` at the end. So, I need to relock it here. There is another call to `prepare_reopen()`  inside `producer_get_audio()` which is why that function is not changed. I see I also erroneously removed the unlock on the packets_mutex.

src/modules/avformat/producer_avformat.c
ddennedy: Yes, thanks

gabrieljablonski: > Wait, my branch of this says `return -1` after this line instead of `goto exit_get_image`

Forgot to mention I changed this as well. `goto` is probably the proper way to go about it here, since simply returning would leave a bunch of stuff in an inconsistent state.

ddennedy:
Instead of your commit bfa9cac531024b5ba08b62f9c758b10871349a5c to fix draining on EOF, please test the following which solves it for me as well as avoid the infinite loop:
```
--- src/modules/avformat/producer_avformat.c
+++ src/modules/avformat/producer_avformat.c
@@ -1793,9 +1793,9 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 			self->sw_video_frame = av_frame_alloc();
 #endif
 
-		while ( ( self->video_send_result >= 0 || self->video_send_result == AVERROR( EAGAIN ) ) && !got_picture )
+		while (!got_picture && (self->video_send_result >= 0 || self->video_send_result == AVERROR(EAGAIN) || self->video_send_result == AVERROR_EOF))
 		{
-			if ( self->video_send_result != AVERROR( EAGAIN ) ) 
+			if ( self->video_send_result != AVERROR( EAGAIN ) )
 			{
 				// Read a packet
 				if ( self->pkt.stream_index == self->video_index )
@@ -1819,8 +1819,7 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 					{
 						if ( ret == AVERROR_EOF ) 
 						{
-							pthread_mutex_unlock( &self->packets_mutex );
-							break;
+							self->pkt.stream_index = self->video_index;
 						} else 
 						{
 							mlt_log_verbose( MLT_PRODUCER_SERVICE( producer ), "av_read_frame returned error %d inside get_image\n", ret );
@@ -1889,7 +1888,7 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 					self->video_send_result = avcodec_send_packet( codec_context, &self->pkt );
 					mlt_log_debug( MLT_PRODUCER_SERVICE( producer ), "decoded packet with size %d => %d\n", self->pkt.size, self->video_send_result );
 					// Note: decode may fail at the beginning of MPEGfile (B-frames referencing before first I-frame), so allow a few errors.
-					if (self->video_send_result < 0 && self->video_send_result != AVERROR(EAGAIN))
+					if (self->video_send_result < 0 && self->video_send_result != AVERROR(EAGAIN) && self->video_send_result != AVERROR_EOF)
 					{
 						mlt_log_warning( MLT_PRODUCER_SERVICE( producer ), "avcodec_send_packet failed with %d\n", self->video_send_result );
 					}
```

ddennedy:
I pushed commits to the mlt repo branch of this to support videotoolbox, d3d11va, and dxva2 hwaccels for macOS and Windows. So far, this is working good for me, and I will merge it soon. Also, soon I am going to make a branch for a major new version with API/ABI-breaking changes where I hope to drop the "fs:" protocol and require paths with a '?' to be escaped or encoded.

ddennedy:
I pushed another commit in my branch to replace deprecated `avcodec_decode_audio4()`. I know it is not needed for hwaccel and av1, but it is the final thing I wanted to get done regarding deprecations in the producer for now, and these branches are based on my ffmpeg_deprecations branch. Other deprecations are not so easy and require more study. I could use your help testing that change to finally merge.

gabrieljablonski:
I unfortunately won't have access to the machines I use for testing until next week.

So whether you decide to include anymore changes or not, I'll be able to apply and test them probably on Monday.

ddennedy:
> When the MLT profile does not match the audio/video file then A/V sync is completely broken and video rate is erratic.

This was a regression in the memory leak fix, and I pushed a fix to my branch.

src/modules/avformat/producer_avformat.c
```diff
@@ -2092,10 +2097,7 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 				(codec_params->field_order == AV_FIELD_PROGRESSIVE ||
 				 codec_params->field_order == AV_FIELD_UNKNOWN) );
 	}
-	av_frame_unref( self->video_frame );
```
gabrieljablonski: Was leaving this out intentional, or is it just missing above? Without this, the memory leak addressed [here](https://github.com/mltframework/mlt/pull/621#issuecomment-732147656) comes back.

ddennedy: intentional, read above. video_frame must be reusable. We need to find a fix for the memory leak some other way.

src/modules/avformat/producer_avformat.c
```diff
@@ -1199,7 +1199,7 @@ static int seek_video( producer_avformat self, mlt_position position,
 			// Remove the cached info relating to the previous position
 			self->current_position = POSITION_INVALID;
 			self->last_position = POSITION_INVALID;
-			av_freep( &self->video_frame );
+			av_frame_unref(self->video_frame);
```
gabrieljablonski, line 1202: If this doesn't turn out to have any unintended side effects, I believe it properly fixes the memory leak.

src/modules/avformat/producer_avformat.c
```diff
@@ -132,11 +132,12 @@ struct producer_avformat_s
...
-	int hw_device_type;
-	char hw_device[128];
-	AVBufferRef* hw_device_ctx;
-	AVFrame* sw_video_frame;
+	struct {
+		int pix_fmt;
+		int device_type;
+		char device[128];
+		AVBufferRef* device_ctx;
+	} hwaccel;
```
gabrieljablonski, line 140: This change is purely cosmetic. Although I personally thinks this looks nicer since it better isolates the hwaccel context, changing it back is not a problem.

src/modules/avformat/producer_avformat.c
```diff
@@ -1934,20 +1930,22 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 						else
 						{
 #if USE_HWACCEL
-							if ( self->hw_device_ctx && self->video_frame->format == self->hw_pix_fmt )
+							if (self->hwaccel.device_ctx && self->video_frame->format == self->hwaccel.pix_fmt)
 							{
-								int transfer_data_result = av_hwframe_transfer_data( self->sw_video_frame, self->video_frame, 0 );
-								if( transfer_data_result < 0 ) 
+								AVFrame *sw_video_frame = av_frame_alloc();
+								int transfer_data_result = av_hwframe_transfer_data(sw_video_frame, self->video_frame, 0);
```
gabrieljablonski, line 1936: `sw_video_frame` doesn't have any use outside of this code section, and I don't imagine this would change. Removing it from producer context seems like a good idea.

src/modules/avformat/producer_avformat.c
ddennedy: Yes, seems to be good here without side effect. But also, I think we should do
```
--- src/modules/avformat/producer_avformat.c
+++ src/modules/avformat/producer_avformat.c
@@ -3129,3 +3129,5 @@ static void producer_avformat_close( producer_avformat self )
 	av_frame_unref( self->video_frame );
+	av_frame_free( &self->video_frame );
 	av_frame_unref( self->audio_frame );
+	av_frame_free( &self->audio_frame );
 
@@ -3183,3 +3185,3 @@ static void producer_avformat_close( producer_avformat self )
 			av_packet_unref( pkt );
-			free( pkt );
+			av_packet_free( &pkt );
 		}
@@ -3193,3 +3195,3 @@ static void producer_avformat_close( producer_avformat self )
 			av_packet_unref( pkt );
-			free( pkt );
+			av_packet_free( &pkt );
 		}
```

src/modules/avformat/producer_avformat.c
ddennedy: OK, but like above, I think we should instead use `av_frame_free()` instead of`av_frame_unref()`.

ddennedy:
You can fix the merge conflict by merging from master and edit the `before_install:` block to read:
```
        - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
        - docker pull ubuntu:20.04
```
(We need to use credentials to pull from docker hub since they enforce limits now, and it would be the anonymous Travis account otherwise, which is always maxed out. And our changes require FFmpeg v4 that is in Ubuntu 20.04.)

gabrieljablonski:
All good! The CI build failed but it seems to be some minor stuff.

gabrieljablonski:
Actually we still have a problem. Although the one memory leak that happened when loading a video is seemingly fixed, stopping the producer and playing it again is also leaking memory. Will try to track it down.

gabrieljablonski:
Interestingly enough, I had actually solved this problem on another branch I had been working on without even noticing. [Here's the file](https://github.com/gabrieljablonski/mlt/blob/525439012080887807c59808f35fb5b521abfaa6/src/modules/decklink/consumer_decklink.cpp).

In my implementation I'm using VANC data, which is not supported in 8-bit frames (which MLT is currently using for the decklink consumer), so I had to setup a conversion from 8 to 10-bit frames, since I was not confident in overhauling the consumer to simply just use 10 bit frames. 

For that to work, I had to create a separate video frame queue which saved the 8-bit base frame after `ScheduleVideoFrame()`, and then popped it back to the main video frame queue under `ScheduledFrameCompleted()`, effectively discarding the 10-bit frame returned in the callback.

I find this to be a very inelegant solution, but it worked for what I was trying to accomplish, and somehow solves the memory leak issue. Hopefully this helps reveal the source of the leak and it can be fixed in a more elegant manner (maybe in another PR?).

Here's the relevant changes:

```
--- a/src/modules/decklink/consumer_decklink.cpp
+++ b/src/modules/decklink/consumer_decklink.cpp
@@ -85,6 +85,7 @@ private:
        mlt_deque                   m_aqueue;
        pthread_mutex_t             m_aqueue_lock;
        mlt_deque                   m_frames;
+       mlt_deque                   m_frames_interim;
 
        pthread_mutex_t             m_op_lock;
        pthread_mutex_t             m_op_arg_mutex;
@@ -141,6 +142,7 @@ public:
                m_deckLink = NULL;
                m_aqueue = mlt_deque_init();
                m_frames = mlt_deque_init();
+               m_frames_interim = mlt_deque_init();
                m_buffer = NULL;
 
                // operation locks
@@ -167,6 +169,7 @@ public:
 
                mlt_deque_close( m_aqueue );
                mlt_deque_close( m_frames );
+               mlt_deque_close( m_frames_interim );
 
                op(OP_EXIT, 0);
                mlt_log_debug( getConsumer(), "%s: waiting for op thread\n", __FUNCTION__ );
@@ -484,7 +487,10 @@ protected:
                pthread_mutex_unlock( &m_aqueue_lock );
 
                m_buffer = NULL;
-               while ( IDeckLinkMutableVideoFrame* frame = (IDeckLinkMutableVideoFrame*) mlt_deque_pop_back( m_frames ) )
+               IDeckLinkMutableVideoFrame* frame;
+               while ( frame = (IDeckLinkMutableVideoFrame*) mlt_deque_pop_back( m_frames ) )
+                       SAFE_RELEASE( frame );
+               while ( frame = (IDeckLinkMutableVideoFrame*) mlt_deque_pop_back( m_frames_interim ) )
                        SAFE_RELEASE( frame );
 
                // set running state is 0
@@ -618,6 +624,7 @@ protected:
                                mlt_log_error( getConsumer(), "%s:%d: ScheduleVideoFrame failed, hr=%.8X \n", __FUNCTION__, __LINE__, unsigned(hr) );
                        else
                                mlt_log_debug( getConsumer(), "%s: ScheduleVideoFrame SUCCESS\n", __FUNCTION__ );
+                       mlt_deque_push_back( m_frames_interim, decklinkFrame );
                }
        }
 
@@ -793,7 +800,9 @@ protected:
        {
                mlt_log_debug( getConsumer(), "%s: ENTERING\n", __FUNCTION__ );
 
-               mlt_deque_push_back( m_frames, completedFrame );
+               // mlt_deque_push_back( m_frames, completedFrame );
+               if (void *frame = mlt_deque_pop_front(m_frames_interim))
+                       mlt_deque_push_back(m_frames, frame);
 
                //  change priority of video callback thread
                reprio( 1 );
```

ddennedy:
I just pushed a change on master d769db242 that changes the semantics for a file URL with a query string. Now, the special "fs:" protocol is not used or supported. An unescaped `?` is considered a part of the file name, and the query string delimiter is `\?`. However, when using the melt command line, a string with `=` and no `:` before the `=` is interpreted as a property. In order to force to interpret the string as a producer URL you must add a prefix with either a protocol ("file:" works) or the "avformat:" service name.
Example:
`melt -verbose 'file:/home/ddennedy/net/media/unsorted/videos/gopro/2021-02/GX010217.MP4\?hwaccel=vaapi'`

Outside of melt, such as in XML or API, a protocol or service name prefix is not required.

PR #625, opened 2020-11-29
Fix crash with files having more than 32 streams

j-b-m:
When using a file with more than 32 streams (a user had a file with many subtitle streams), MLT crashes when setting audio_index=-1. This is reproducible with the file attached to the Kdenlive bug (426640) report:
https://bugs.kde.org/attachment.cgi?id=133701&action=edit

SImply do:
melt out.mkv audio_index=-1
will crash. Patch adds a check to prevent accessing an invalid array index.


ddennedy:
Thanks!

PR #626, opened 2020-11-30
Add strobing filter

sandsmark:
Extremely simple filter, just alternates between clearing and forwarding the normal frame.

Could probably be more efficient, but I'm not that familiar with the APIs.

Tested it briefly in kdenlive, and it seems to work fine, except when there's any filters after it that overwrite the alpha (like the transform filter), not sure if that can easily be fixed (short of converting the entire frame to something with an alpha channel and overwriting that, which kills performance pretty badly).

### DIFF
--- a/src/modules/plus/filter_strobe.c
+++ b/src/modules/plus/filter_strobe.c
@@ -1,0 +1,118 @@
+/*
+ * filter_strobe.c -- simple strobing filter
+ * Copyright (C) 2020 Martin Sandsmark <martin.sandsmark@kde.org>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include <framework/mlt_filter.h>
+#include <framework/mlt_frame.h>
+#include <framework/mlt_producer.h>
+#include <framework/mlt_service.h>
+#include <framework/mlt_factory.h>
+#include <framework/mlt_property.h>
+
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
+{
+	mlt_filter filter = mlt_frame_pop_service( frame );
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+
+	int error = mlt_frame_get_image( frame, image, format, width, height, 1 );
+	if ( error )
+	{
+		return error;
+	}
+
+	int invert = mlt_properties_get_int( properties, "strobe_invert" );
+	int interval = mlt_properties_get_int( properties, "interval" );
+	mlt_position currentpos = mlt_filter_get_position( filter, frame );
+
+	int do_strobe = ( currentpos % ( interval + 1 ) ) > interval / 2;
+
+	if ( invert )
+	{
+		do_strobe = !do_strobe;
+	}
+
+	if ( do_strobe != 1 )
+	{
+		return 0;
+	}
+
+	assert( *width >= 0 );
+	assert( *height >= 0 );
+	size_t byteCount = *width * *height;
+
+	// We always clear the alpha mask, in case there's some optimizations
+	// that can be applied or other filters modyify the image contents.
+	uint8_t *alpha_buffer = mlt_frame_get_alpha_mask( frame );
+	assert( alpha_buffer != NULL );
+
+	// We assert because the API contract guarantees that we always get a
+	// buffer, but don't want to crash in release build if it is broken.
+	if ( alpha_buffer )
+	{
+		memset( alpha_buffer, 0, byteCount );
+	}
+
+	if ( *format == mlt_image_rgb24a )
+	{
+		for ( size_t i=3; i<byteCount * 4; i+=4 )
+		{
+			image[i] = 0;
+		}
+	}
+
+	return 0;
+
+}
+
+/** Filter processing.
+*/
+
+static mlt_frame filter_process( mlt_filter filter, mlt_frame frame )
+{
+
+	// Push the filter on to the stack
+	mlt_frame_push_service( frame, filter );
+
+	// Push the frame filter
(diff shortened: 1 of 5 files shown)


sandsmark:
chaining avfilter.photosensitivity after this is fun.

src/modules/plus/filter_strobe.yml
```diff
@@ -0,0 +1,31 @@
...
+language: en
+description: Strobes the alpha channel to 0. Many other filters overwrite the alpha channel, in that case this needs to be last.
+tags:
+  - Video
+parameters:
+  - identifier: strobe_invert
+    title: Invert
+    type: boolean
+    description: Whether to invert which frames are on and which is off
+    default: false
```
ddennedy, line 18: We do not use `true` or `false` as values for booleans in the service metadata. You can see the `metaschema.yml` for reference. Please change this to 0.

src/modules/plus/filter_strobe.c
```diff
@@ -0,0 +1,91 @@
...
+	mlt_filter filter = mlt_frame_pop_service( frame );
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+
+	int error = mlt_frame_get_image( frame, image, format, width, height, 1 );
+	if (error) {
+		return error;
+	}
+
+	int invert = mlt_properties_get_int( properties, "strobe_invert" );
+	int interval = mlt_properties_get_int( properties, "interval" );
```
ddennedy, line 42: Do you want to make the interval keyframable? It is fairly easy but not required.

src/modules/plus/filter_strobe.c
```diff
@@ -0,0 +1,91 @@
+/*
+ * filter_strobe.c -- simple strobing filter
+ * Copyright (C) 2007 Martin Sandsmark <martin.sandsmark@kde.org>
```
ddennedy, line 3: You should make the copyright year 2020 even though we all hate that year now.

ddennedy (changes requested):

ddennedy:
Thanks for the contribution. Your filter cannot control what happens downstream, but you can add the filter to the master output track.

src/modules/plus/filter_strobe.c
```diff
@@ -0,0 +1,91 @@
...
+	mlt_position currentpos = mlt_filter_get_position( filter, frame );
+
+	int do_strobe = (currentpos % (interval + 1)) > interval/2;
+	if (invert) {
+		do_strobe = !do_strobe;
+	}
+
+	if (do_strobe == 1) {
+		int alphasize = *width * *height;
+		uint8_t *alpha_buffer = mlt_frame_get_alpha_mask( frame );
```
ddennedy, line 52: If the `format` is `mlt_image_rgb24a` then you need to clear that alpha channel as well because if a downstream component requests rgb24a and your service returns there is no conversion that will happen in `mlt_frame->convert_image()`.

src/modules/plus/filter_strobe.c
sandsmark: I thought about it, but didn't really see the point for myself.

src/modules/plus/filter_strobe.c
sandsmark: haha, but fixed.

src/modules/plus/filter_strobe.c
sandsmark: this might explain some of the issues I had while testing.

sandsmark: Is there a preferred method of doing this, or should I just do it the old school way?

sandsmark: just did it the oldschool way

src/modules/plus/filter_strobe.c
sandsmark: Do I just need to set `mutable: yes` and switch to `mlt_properties_anim_get_int`?

ddennedy: You may not see the point, but this will get exposed to users probably through Kdenlive who will then ask for it repeatedly.

`mutable: yes` is not for animation; it simply means when the app changes the value the filter sees the new value instead of only the initial. We do not yet have a flag in service metadata for animation keyframes.

`mlt_properties_anim_get_int` mostly, yes, but that function needs a position and duration:
```
	mlt_position position = mlt_filter_get_position( filter, frame );
	mlt_position length = mlt_filter_get_length2( filter, frame );
```

src/modules/plus/filter_strobe.c
```diff
@@ -0,0 +1,109 @@
...
+	if ( do_strobe != 1 )
+	{
+		return 0;
+	}
+
+	assert( *width >= 0 );
+	assert( *height >= 0 );
+	size_t alphasize = *width * *height;
+	uint8_t *alpha_buffer = mlt_frame_get_alpha_mask( frame );
+	memset( alpha_buffer, 0, alphasize );
```
bmatherly, line 63: I think you could put this in an else statement below since rgb24a will not need an alpha mask.

Also good to check the pointer for null before accessing it (although failure is unlikely).

src/modules/plus/filter_strobe.c
sandsmark: > You may not see the point, but this will get exposed to users probably through Kdenlive who will then ask for it repeatedly.

Yup, that's why I want to implement it, was just explaining why it wasn't done initially.

src/modules/plus/filter_strobe.c
sandsmark: > I think you could put this in an else statement below since rgb24a will not need an alpha mask.

I'm not sure if that's true for all filters? I. e. if some filters modify the image we still want to completely blank the frame.

I plan to go through the other filters and fix the issue with them ignoring the alpha mask eventually, though, since it's not intuitive that this filter is overridden if it's not last in the chain.


> Also good to check the pointer for null before accessing it (although failure is unlikely).

the API contract (or whatever, the comments at least) states that mlt_frame_get_alpha_mask() will allocate if missing, but I can add a check anyways.

ddennedy: >> I think you could put this in an else statement below since rgb24a will not need an alpha mask.
> I'm not sure if that's true for all filters? I. e. if some filters modify the image we still want to completely blank the frame.

Image conversion in MLT takes care of merging and extracting the alpha channel when converting to and from rgba. It is not the intention that alpha channel needs to be redundant on every frame. I am sorry I gave that impression. If a filter is operating in rgba, it can assume the embedded alpha channel is fine. Otherwise, it can use `mlt_frame_get_alpha()` or `mlt_frame_get_alpha_mask()`. Brian is correct that if else can work here.

> I plan to go through the other filters and fix the issue with them ignoring the alpha mask eventually,

That is not desired. I think the problem you are seeing is that some services accidentally overwrite the alpha channel instead of passing it through. Some frei0r filters have an alpha operation parameter, which is nice.

ddennedy: One thing that makes alpha channel handling messy in MLT is the `get_alpha_mask` function pointer in `mlt_frame` as used inside `mlt_frame_get_alpha_mask()`. I would like to get rid of that in v7. There is only one thing that uses it (`transition_region.c`), and it can be interesting, but it makes things complicated: if a filter is operating in rgba but the frame has this function pointer set, what should it do? Many things (e.g. frei0r, Qt) cannot handle two sources of alpha channel.

ddennedy (approved):

PR #631, opened 2020-12-13
Add 2020 color space metadata

bmatherly:


src/modules/avformat/producer_avformat.c
```diff
@@ -442,6 +442,13 @@ static mlt_properties find_default_streams( producer_avformat self )
 				case AVCOL_SPC_BT709:
 					mlt_properties_set_int( meta_media, key, 709 );
 					break;
+				case AVCOL_SPC_BT2020_NCL:
+				case AVCOL_SPC_BT2020_CL:
+					mlt_properties_set_int( meta_media, key, 2020 );
+					break;
```
bmatherly, line 448: This is not very scalable. I think that eventually we should convert this to use the actual value from ffmpeg (which is based on a standard)

ddennedy: I did not realize there is a standard for enum values! I see it is now withdrawn: https://www.iso.org/standard/62088.html
In any case, I am not certain scalability here is much needed as MLT does not aim for the comprehensiveness of FFmpeg or gstreamer. The current approach was made for user friendliness (easy to specify, easy to see in property values in files and log messages, etc.). In MLT 7 we can certainly look to use something like `mlt_image` with name/id functions like we have done elsewhere. and use ISO/IEC 23001-8 values.

ddennedy:
As you can see, this breaks the build on the current supported versions of FFmpeg. This can wait for one of the other branches dependent upon a newer version of FFmpeg.

ddennedy (changes requested):
Build error `producer_avformat.c:449:10: error: 'AVCOL_SPC_SMPTE2085' undeclared (first use in this function)`
Drop the seldom-used 2085, and I can merge this sooner than later.

src/modules/avformat/producer_avformat.c
bmatherly: See the updated commit for a new proposal. Essentially: 240, 601 and 709 mean the same thing they always have. Continue to assume 601 and 709 for SD and HD respectively. Pass the FFmpeg value for 2020 and 4k resolutions. This can get it through a transitional period so that going forward all new color spaces can use the FFMpeg values.

Thoughts?

If we don't like this, I can easily change it back to the 2020 value.

ddennedy: I am not keen on this change. There is a lot of 4K video devices and tools are making without an explicit colorspace. I checked several of my samples, and it was definitely hit-or-miss. You would think a camera would have it, but I checked my Sony F55 samples, and it did not! Also, an older GoPro Hero 4 and some stuff downloaded from YouTube did not. This is more than informational, it is supplied as an input to swscale, and I am not sure what it does with AVCOL_SPC_UNSPECIFIED. 4K is far more likely to be 709 than anything else by a wide margin.

ddennedy: I am fine to change the values for BT.2020 to the enum values, but we should also change `filter_avformat.c`, which is using 2020 and 2021. We can keep those values in there for backwards compatibility, but it is not like they were really being used currently.

ddennedy (approved):

bmatherly:
89a5c94 is my favorite one so far. It works for all future types that may be added while keeping legacy values unchanged.

PR #633, opened 2020-12-15
fix build without swig

sandsmark:
without this it fails very silently and weird (you get a "permission denied" error when building).


ddennedy:
thanks

PR #634, opened 2020-12-19
avformat producer: 10- and 16-bit color support

vivid-synth:
This is not ready to merge; I've added support but could use help with testing and review!


ddennedy:
What is your use case for this? `mlt_image_yuv422p16` is not the direction planned to support image processing > 8 bits. I do not yet know what it is, but it is unlikely to be sub-sampled, non-linear-color, 16-bit YUV. `mlt_image_yuv422p16` was added for a contributor's specific use case and not general usage. Currently, with Movit, > 8-bit is half float linear RGB.

bmatherly:
I would also be interested to know the OP's use case.

10 and 12 bit 4:2:2 using 709 colorspace has been around in the broadcast world long before anyone was talking about HDR. I had always expected that we would need support for it. I envisioned  mlt_image_yuv422p16 as our "bridge" to allow round trip processing of these >8 bit broadcast formats.

Even if we choose to recommend movit for >8 bit, we still need a way to get the image data in/out of libav without bit depth loss.

ddennedy:
My point is not about HDR but general usage. Does accepting open the door to further expectations about processing in that image format? This could certainly help with edits that do not require video processing. Also `mlt_image_yuv422p16` can be added to the avfilter filter for use with avfilters that can handle that. I admit I have a Shotcut- and Kdenlive-centric perspective to MLT.

bmatherly:
I would add, too that Qt has QImage::Format_RGBA64 now. I think it would be trivial to add support for 16bit yuv and RGB. I do not know if those formats should be used in Shotcut/Kdenlive due to risk of poor user experience. But for other use cases, it could be a good addition.

vivid-synth:
The use case here is I've got a camera that produces 10-bit color video (4:2:0 or 4:2:2) and I'd like to be able to edit and render in 10-bit without losing information.

My best-case scenario is to have full integration -- at least a few transitions, filters, etc. working in 10bit color. In the best case, this functionality would head downstream and be supported directly in Shotcut, Kdenlive, etc.

However, my minimum-necessary scenario is simply being able to make an Edit Decision List-style MLT XML file which simply has avformat producers and an avformat consumer, all in 10-bit color. A possible workflow I can imagine is to use Shotcut/Kdenlive even without 8-bit color support, but then go in and tweak the MLT XML file by hand for the final render. At the moment, I use off-the-shelf Kdenlive with my 10-bit color video and it handles them fine - it just downgrades to 8-bit.

I am totally happy to re-work this PR to use e.g. half-float linear RGB. The only reason I chose `mlt_image_yuv422p16` was that it seemed to be a dual of the avformat consumer. If using half-float linear RGB allowed me to use Movit services with 10-bit color, I'd be even more inclined to go in that direction.

src/modules/avformat/producer_avformat.c
```diff
@@ -1450,6 +1461,21 @@ static int convert_image( producer_avformat self, AVFrame *frame, uint8_t *buffe
...
+	else if ( *format == mlt_image_yuv422p16 )
+	{
+		int flags = mlt_get_sws_flags(width, height, src_pix_fmt, width, height, AV_PIX_FMT_RGBA);
+		struct SwsContext *context = sws_getContext( width, height, src_pix_fmt,
+			width, height, AV_PIX_FMT_YUV422P16LE, flags, NULL, NULL, NULL);
+		uint8_t *out_data[8];
+		int out_stride[8];
+		av_image_fill_arrays(out_data, out_stride, buffer, AV_PIX_FMT_YUV422P16LE, width, height, IMAGE_ALIGN);
+		// libswscale wants the RGB colorspace to be SWS_CS_DEFAULT, which is = SWS_CS_ITU601.
+		mlt_set_luma_transfer( context, self->yuv_colorspace, 2100, self->full_luma, 0 );
```
vivid-synth, line 1473: I should also point out here that I optimistically chose 2100 here even though I don't know if MLT supports it.

faridosc:
> What is your use case for this? `mlt_image_yuv422p16` is not the direction planned to support image processing > 8 bits. I do not yet know what it is, but it is unlikely to be sub-sampled, non-linear-color, 16-bit YUV. `mlt_image_yuv422p16` was added for a contributor's specific use case and not general usage. Currently, with Movit, > 8-bit is half float linear RGB.

Would you consider using [OpenColorIO](https://github.com/AcademySoftwareFoundation/OpenColorIO) for color management? This would help us integrate our work in a pipeline with [other programs](https://opencolorio.org/#supported_apps) (floss or proprietary) by using default industry standard format. 

I don't know how complex of a task this is but as member of the Kdenlive community I am willing to help out by setting a campaign to crowdfund it (or any other non-technical support required) to see it happen. If you think it would be feasible of course.

ddennedy (changes requested):
I plan to treat this like the way we handle `mlt_image_yuv420p` today - limited and integrated with:

- `producer_avformat.c` (this PR)
- `filter_avcolour_space.c` (for including 8-bit sources in a project including RGB, done)
- `consumer_avformat.c` (done)
- `filter_avfilter.c` (should be easy)
- `movit` (see `filter_movit_convert.cpp` in MLT and `ycbcr_input.h` in movit)

That would allow the use cases of end-to-end processing using either avfilter or movit for video processing. That is not a full solution for the MLT NLEs, of course, but I think it is fair based on what we have today. I will not require the movit integration to merge this, but I would like to the avfilter one.

ddennedy:
> Would you consider using OpenColorIO for color management?

I can add it to the [Road Map](https://github.com/mltframework/mlt_web/blob/master/changes/todo.md). However, one of Brian's pull requests is ready and necessitates breaking ABI. Thus, it is time to make MLT v7, and we need to make a number of planned changes so as to not require another API breakage soon after. As a result, Brian and I will be busy working on those things for the next few months. Crowdfunding will not help with that, but perhaps another qualified developer would. 

My knowledge of OpenColorIO is limited, but based on what I understand simply dropping OpenColorIO in place of libswscale will be a performance killer. Trying to route 4K60 uncompressed (non-subsampled?) video to and from GPU memory is slow, and my understanding is that its CPU-based code is very slow.

sobotka:
Hay @vivid-synth — instead of arguing about the glaringly obvious need for OCIO and float image processing here, why not head over to Olive and ply your code there? Float offline reference working space, OCIO is already integrated including V2, and a full offline / online rendering chain.

elsandosgrande:
To add to Troy's comment above, what's stopping you from going full-GPU like Olive did (and don't mention frei0r, it's not color managed)? Sure, it would require a lot of changes, which would require quite a bit of work and necessitate a whole new major version of MLT, but we should both be able to agree that leveraging Vulkan, or even OpenGL, would yield greater performance on a wider range of hardware, not to mention that it would mitigate the possible performance issues of a CPU-GPU round trip when using OpenColorIO. 

I understand that the Movit library already tries to levarage the GPU, but my experience with it and KDEnlive was not all that great. Sure, nothing crashed, but I could never get a single video to render when using GPU effects, like the GPU vignette effect. That _might_ have changed in the latest releases, but I sadly don't have the time to test right now. I think that it would be better to just fully switch to the GPU and forget about CPU rendering, that is integrate GPU rendering (**not** encoding) into MLT, therefore forgetting about Movit.

Also, do you consult with people who know color science very well? I'm still just a beginner, but people like Troy would probably be nice to consult ([he's well-versed in color science](https://hg2dc.com) and he's already been helping the Olive project out a lot, the rewrite was a result of his guidance). I don't know if you're already consulting somebody, but, if you're not, I think that the project could benefit a lot from it.

Have a nice day!

faridosc:
> The use case here is I've got a camera that produces 10-bit color video (4:2:0 or 4:2:2) and I'd like to be able to edit and render in 10-bit without losing information.
> 
> My best-case scenario is to have full integration -- at least a few transitions, filters, etc. working in 10bit color. In the best case, this functionality would head downstream and be supported directly in Shotcut, Kdenlive, etc.
> 
> However, my minimum-necessary scenario is simply being able to make an Edit Decision List-style MLT XML file which simply has avformat producers and an avformat consumer, all in 10-bit color. A possible workflow I can imagine is to use Shotcut/Kdenlive even without 8-bit color support, but then go in and tweak the MLT XML file by hand for the final render. At the moment, I use off-the-shelf Kdenlive with my 10-bit color video and it handles them fine - it just downgrades to 8-bit.
> 
> I am totally happy to re-work this PR to use e.g. half-float linear RGB. The only reason I chose `mlt_image_yuv422p16` was that it seemed to be a dual of the avformat consumer. If using half-float linear RGB allowed me to use Movit services with 10-bit color, I'd be even more inclined to go in that direction.

Our studio and other independent videomakers using floss need this step forward. I hope you manage to commit this. Thanks! :)

faridosc:
> Crowdfunding will not help with that, but perhaps another qualified developer would.

If you know anyone qualified for this task let us know.

faridosc:
> > Would you consider using OpenColorIO for color management?
> 
> I will add it to the [Road Map](https://github.com/mltframework/mlt_web/blob/master/changes/todo.md). However, one of Brian's pull requests is ready and necessitates breaking ABI. Thus, it is time to make MLT v7, and we need to make a number of planned changes so as to not require another API breakage soon after. As a result, Brian and I will be busy working on those things for the next few months. Crowdfunding will not help with that, but perhaps another qualified developer would.

Wow, looking forward for MLT 7, lot's of cool stuff. If you would consider putting OCIO, OTIO ad LV2 support up the list would be fantastic. 

If you know anyone qualified for this task let us know!

> 
> My knowledge of OpenColorIO is limited, but based on what I understand simply dropping OpenColorIO in place of libswscale will be a performance killer. Trying to route 4K60 uncompressed (non-subsampled?) video to and from GPU memory is slow, and my understanding is that its CPU-based code is very slow.

I hope we can find a way to deal with this issue.

bmatherly:
>> My knowledge of OpenColorIO is limited, but based on what I understand simply dropping OpenColorIO in place of libswscale will be a performance killer. Trying to route 4K60 uncompressed (non-subsampled?) video to and from GPU memory is slow, and my understanding is that its CPU-based code is very slow.

> I hope we can find a way to deal with this issue.

I expect that a full GPU pipeline will be the solution for this. This will be a long road, and I dread the user support from compatibility issues.

PR #639, opened 2021-01-09
Typewriter effect for kdenlive titler

rlalik:
Effect is based on:
 https://github.com/rlalik/mlt_extra_modules@v0.3.3

For description of usage see:
  https://github.com/rlalik/mlt_extra_modules

Here is the demo video:
  https://www.youtube.com/watch?v=9_ouF3NLTxA


src/modules/kdenlive/filter_typewriter.cpp
```diff
@@ -0,0 +1,287 @@
+/*
+ * filter_oldfilm.c -- oldfilm filter
```
bmatherly, line 2: This header is incorrect.

src/modules/kdenlive/CMakeLists.txt
```diff
@@ -2,9 +2,16 @@ if(NOT MOD_KDENLIVE)
     return()
 endif()
 
+find_package(Qt5 COMPONENTS Core Xml)
```
bmatherly, line 5: Can you explain why the changes in this file are required for the typewriter module? I do not see where the typewriter code uses Qt.

src/modules/kdenlive/configure
```diff
@@ -0,0 +1,130 @@
+#!/bin/sh
```
bmatherly, line 1: I also wonder why this file is needed. I do not know if MLT can still compile with Qt 4 - so some of the checks may be unnecessary.

src/modules/kdenlive/filter_typewriter.cpp
rlalik: I'll get it corrected

src/modules/kdenlive/CMakeLists.txt
rlalik: The `filter_typewriter.cpp` does almost not (except lines 228-229), but there is whole c++ class in `typewriter.cpp` which renders the effects frames and it uses Qt5::Xml for parsing xml from kdenlive to extract text content.

src/modules/kdenlive/configure
rlalik: I reused the checks from `src/modules/qt/configure`, in this way I am at least compatible and not limited to the rest of mlt. If Qt4 support is dropped, then it could be removed from the whole project.

bmatherly: My question about why it is needed for typewriter. QDomElement is already used in kdenlivetitle_wrapper.cpp. So it would not be required to add a check for Qt. 

I think that maybe this commit includes two unrelated changes:
1) Add typewriter
2) Add configure script and cmake check for Qt detection?

That might be OK. It is just good for the reviewer to know.

rlalik: Yes, agree, might be redundant, but question appears - whether I should be depent on other module which I am not related to? What is one day the whole `src/modules/qt` is gone, or the checks in qt module will change and I lose my checks I trusted on? And if the qt checks are required by more modules, maybe one should move it higher in the checks tree.
And does it harm to have it repeated? I am able to remove it if reviewer ask for.

eszlari: I opened a PR (https://github.com/mltframework/mlt/pull/641) that moves all dependency checks to the top level CMakeLists.txt.

rlalik: I can wait when PR #641 is accepted then update my code accordingly and commit new changes.

ddennedy: That PR does not address the configure and makefile build system to which this belongs. In any case, this is the wrong folder/module for this service. It belongs in `qt` since it depends on a) Qt and b) kdenlivetitle.

ddennedy:
Please change the name of this to `kdenlive_typewriter` to make it more clear that it is based on kdenlivetitle and to avoid future conflict with a potential Qt- or Pango-based typewriter effect.

src/modules/kdenlive/configure
rlalik: It has been moved to qt module.

rlalik:
> Please change the name of this to `kdenlive_typewriter` to make it more clear that it is based on kdenlivetitle and to avoid future conflict with a potential Qt- or Pango-based typewriter effect.

Though it was written for kdenlive, it is based on universal typewriter rendered which just happen to work with kdenlive in this effect, but the same engine can be extended to any other kind of producer. In fact it is kind of producer independent, it works like a man-in-the-middle:
* it captures the producer data (the text on which the effect should be applied),
* modifies it for each frame,
* and a such modified text is forwarded to the meant producer.

So if in future any other producer want to use it, it is just a matter of implementing proper handling of the given producer. Thus the name should not need to reflect that it is kdenlive effect, it is much more generic.

ddennedy:
I do not know what “universal typewriter rendered” means. It parses XML. What XML? Where is the DTD or schema?

rlalik:
It parses XML because kdenlive_titler uses XML. If you had another producer which accepts json, qml, or whatever you want, just capture it, parse properly, run tw renderer, and update your format: json, qml, whatever, and rediret back to the producer.

ddennedy:
Got it. Thanks

ddennedy:
Thank you for your contribution; this is very interesting. I have done a full review of it yet, and that may take some time especially as I do not regularly use kdenlive. Please be patient. I am very open for others to test it and review, and that will help move things along.

smartfuturecode:
I tested this effect with kdenlive and after @rlalik fixed some issues it works fine and stable.
However I have to admit, that I haven’t tested to render stuff with the effect because of some problems (but these problems are not connected to this effect!!!)

src/modules/qt/filter_typewriter.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+        if (cont->is_template)
+            d = mlt_properties_get( producer_properties, "_xmldata");
+        else
+            d = mlt_properties_get( producer_properties, "xmldata" );
+
+        if (d == nullptr)
+            return 0;
+
+        step_length = atoi(mlt_properties_get(filter_p, "step_length"));
```
ddennedy, line 125: This is the wrong way to convert a property to integer or float. These will fail on null pointers when these properties have not been set or cleared. See `mlt_properties.h` for the ways to get these as integer and double. You must use our functions instead of converting string to numeric like this.

src/modules/qt/filter_typewriter.yml
```diff
@@ -0,0 +1,55 @@
...
+    Original development: https://github.com/rlalik/mlt_extra_modules
+
+parameters:
+  - identifier: step_length
+    title: Distance between basic steps
+    type: integer
+    description: Defines how many frames it takes to display next character
+    readonly: no
+    mutable: yes
+    default: 25
```
ddennedy, line 26: This parameter default value is not initialized in `filter_typewriter.cpp` and must be. It is not enough to rely on kdenlive to set these. This must also be a well-behaved member of MLT. Same goes for additional values below that are not 0.

src/modules/qt/filter_typewriter.yml
```diff
@@ -0,0 +1,55 @@
...
+    mutable: yes
+    default: 0
+
+  - identifier: macro_type
+    title: Macro expansion type
+    type: integer
+    description: |
+        Defines type of macro expansion:
+        0 - custom macro,
+        1 - expansion char-by-chat
```
ddennedy, line 50: "chat" is a misspelling of "char"

ddennedy (changes requested):
I test this today, and unfortunately it fails immediately with the command line
`melt -profile atsc_1080p_24 ~/src/kdenlive/data/titles/simple.kdenlivetitle -attach typewriter step_length=10`

It must be usable and testable in melt as I will not be able to test it in kdenlive.

src/modules/qt/filter_typewriter.yml
```diff
@@ -0,0 +1,55 @@
...
+parameters:
+  - identifier: step_length
+    title: Distance between basic steps
+    type: integer
+    description: Defines how many frames it takes to display next character
+    readonly: no
+    mutable: yes
+    default: 25
+
+  - identifier: sigma_step
```
ddennedy, line 28: It is not `sigma_step`; it is `step_sigma`.

ddennedy:
When I use the following command line where I give all parameters default values explicitly to avoid crash, I do not see any typewriter effect. I only see the full text immediately:

`melt -profile atsc_1080p_25 ~/src/kdenlive/data/titles/simple.kdenlivetitle -attach typewriter step_length=25 step_sigma=0 random_seed=0 macro_type=1`

I tried variations on this with no improvement. For example

`melt profile atsc_1080p_25 kdenlivetitle:$HOME/src/kdenlive/data/titles/simple.kdenlivetitle resource=$HOME/src/kdenlive/data/titles/simple.kdenlivetitle out=300 -attach typewriter step_length=25 step_sigma=0 random_seed=0 macro_type=1 out=300`

src/modules/qt/filter_typewriter.cpp
rlalik: Fixed in the new commit.

src/modules/qt/filter_typewriter.yml
rlalik: The init function was updated to load defaults.

src/modules/qt/filter_typewriter.yml
rlalik: Corrected.

src/modules/qt/filter_typewriter.yml
rlalik: Corrected.

rlalik:
> I test this today, and unfortunately it fails immediately with the command line
> 
> `melt -profile atsc_1080p_24 ~/src/kdenlive/data/titles/simple.kdenlivetitle -attach typewriter step_length=10`
> 
> It must be usable and testable in melt as I will not be able to test it in kdenlive.

Without the last commit it indeed didn't work. With the new fixes, it works for me. Also manipulating the parameters works now.

src/modules/qt/filter_typewriter.cpp
```diff
@@ -0,0 +1,313 @@
...
+        }
+        cont->step_length = step_length;
+        cont->sigma = sigma;
+        cont->seed = seed;
+    }
+
+    return 1;
+}
+
+static int update_producer(mlt_frame frame, mlt_properties frame_p, FilterContainer * cont, int restore)
```
ddennedy, line 206: Parameter `frame_p` is not used and can be removed. Also, restore can be a `bool`.

ddennedy (changes requested):
OK, I got it to work now, but it will not work correctly if I use consumer property `real_time=-2`, which enables frame threading. You probably do not know about that, but basically that makes `filter_get_image()` concurrent, and it is not thread-safe. It is far from easy to make this thread-safe; so, you need to add locks like this:
```
--- src/modules/qt/filter_typewriter.cpp
+++ src/modules/qt/filter_typewriter.cpp
@@ -263,4 +263,6 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
     FilterContainer * cont = (FilterContainer*) filter->child;
 
+    mlt_service_lock(MLT_FILTER_SERVICE(filter));
+
     int res = get_producer_data(properties, frame_properties, cont);
     if (res == 0)
@@ -273,4 +275,6 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
     update_producer(frame, frame_properties, cont, 1);
 
+    mlt_service_unlock(MLT_FILTER_SERVICE(filter));
+
      return error;
 }
```
These last set of changes will complete this.

src/modules/qt/typewriter.cpp
```diff
@@ -0,0 +1,736 @@
...
+    ++i;
+    return i;
+}
+
+/*======================================================================*
+ XML Parser related stuff
+*======================================================================*/
+
+char* clone_string(const char* string) {
+    char* new_string = new char[strlen(string)+1];
```
ddennedy, line 670: You should `#include <cstring>` for `strlen()` and `strcpy()`

rlalik: Missing include add.

src/modules/qt/filter_typewriter.cpp
rlalik: This and others unused arguments has been hide (commented out).

PR #641, opened 2021-01-12
cmake: check deps at top level

eszlari:


ddennedy:
Why? I prefer to have each module's dependencies expressed in the module-local CMakeLists.txt. That also provides a more obvious approach for external modules.

eszlari:
> Why? I prefer to have each module's dependencies expressed in the module-local CMakeLists.txt.

* See #639: If a second module needs a dependency that is already declared, you have to move it to a higher level anyway, leading to needless code churn.
* Having all code relating to dependencies in one central place, is imho better for code discovery / readability. It makes it easier for packagers to find out, all the dependencies need to be added for a build with maximum functionality.

> That also provides a more obvious approach for external modules.

What do you mean specifically?

eszlari:
Another thing: When/if moving from `pkg_check_modules()` to `find_package()` based dependency discovery, it will be necessary to add some `findXXX.cmake` modules for dependencies that don't provide their own `XXXConfig.cmake`. And it's common to put them all in a top level `/cmake` directory and add it to cmake's `CMAKE_MODULE_PATH`.

eszlari:
@ddennedy
If you prefer a module-local approach, I guess you would be ok with moving `option(MOD_* ...)` to the module subdirs too?

vpinon:
> Another thing: When/if moving from `pkg_check_modules()` to `find_package()` based dependency discovery, it will be necessary to add some `findXXX.cmake` modules for dependencies that don't provide their own `XXXConfig.cmake`. And it's common to put them all in a top level `/cmake` directory and add it to cmake's `CMAKE_MODULE_PATH`.

I agree it is the standard way to do.
When I started writing the CMake files, I was not sure what we would need and was a bit lazy, but now I think it's a good thing to sort everything as usual for packagers.
It would be also interesting to contribute the FindXXX.cmake upstream...

eszlari:
After thinking about this a little bit more, I keep coming to the same conclusion: a module-local approach leads either to

* inconsistency: some dependencies are checked at the top level, others at lower levels (see `sdl2`, `fftw`)
* or redundancy: some dependencies are checked multiple times (see `libexif`)

But since I wouldn't like to turn this into a bike-shedding issue and instead focus on completing the TODO list (https://github.com/mltframework/mlt/issues/645), I would just be happy if a decision could be made.

ddennedy:
@eszlari You have convinced me now, and I thank you for your devotion to improving our build system. I want @vpinon's approval to merge this since he contributed it.

vpinon:
I'm OK with this PR, and trust @eszlari to write CMake probably better than me :smile:

ddennedy:
Please resolve and I will merge

eszlari:
> Please resolve and I will merge

done

PR #653, opened 2021-02-06
cmake: add kwalify tests

eszlari:


ddennedy:
FYI, this change has a build failure when merged to the v7 branch:
https://github.com/mltframework/mlt/runs/1846516483?check_suite_focus=true

PR #654, opened 2021-02-07
v7: fix cmake

eszlari:
After file globbing was removed in https://github.com/mltframework/mlt/commit/69521d53974f6738adc62799c30c618a946f0db6, new files now need to be added to the build system explicitly.


ddennedy:
thanks!

Issue #655, opened 2021-02-10
MLT 7.0 Task List


bmatherly:
### Definite
- [x] link and chain for timing effects
- [x] remove mlt_geometry (convert anything affected to mlt_rect)
- [x] rename mlt_service_type enums to be like mlt_service_...
- [x] rename mlt_image_rgb24a as mlt_image_rgba
- [x] rename mlt_image_rgb24 as mlt_image_rgb
- [x] remove mlt_image_opengl
- [x] rename mlt_image_glsl as mlt_image_movit
- [x] rename mlt_image_glsl_texture as mlt_image_opengl_texture
- [x] remove the `height += 1` in mlt_image_format_size()
- [x] remove `mlt_transmitter` and variable argument lists in `mlt_events`
- [x] remove `mlt_frame.get_alpha_mask` function pointer
- [x] remove deprecated modules (dv, feeds, gtk2, kino, linsys, lumas, swfdec, videostab)
- [x] remove motion_est module
- [x] remove the core/data_feed and data_show filters
- [x] remove the core/region filter and and transition
- [x] merge vmfx module into plus
- [x] rename opengl module to movit
- [x] allow building the jackrack module in a LADSPA-only mode
- [x] add a mlt_image class
- [x] convert build completely to CMake

### Under Consideration
- [ ] add const and cleanup constructors in C++ API
- [ ] add option to build without locale
- [ ] mlt_profile -> mlt_context derived from mlt_properties, namely for reference-counting but also to replace mlt_environment()
- [ ] add color_trc, color_range, color_primaries, interpolation, and audio params (channels, frequency, channel_layout) to mlt_context
- [ ] tighten the property namespace (inconsistent service name prefixing, hiding, and putting things on mlt_frame)
- [ ] add mlt_rational type and property functions, and convert things like aspect ratios and frame rates to use it
- [ ] add 16-byte alignment for FFmpeg including aligned image plane and audio channel buffers
- [ ] change mlt_frame_get_image() to take mlt_image instead of multiple parameters (as well as get_image in all services)
- [ ] change mlt_frame_get_audio() to take mlt_audio instead of multiple parameters (as well as get_audio in all services)
- [ ] make more structures in framework more private

ddennedy:
[...]

PR #658, opened 2021-02-14
Image class

bmatherly:
For review and comment.


src/framework/mlt_image.c
```diff
@@ -0,0 +1,374 @@
...
+/** Calculate the number of bytes needed for the Image data.
+ *
+ * \public \memberof mlt_image_s
+ * \param self the Image object
+ * \return the number of bytes
+ */
+
+int mlt_image_calculate_size( mlt_image self )
+{
+	int height = self->height + 1; // Legacy bug workaround
```
ddennedy, line 141: A couple of weeks ago, I added the removal of this `+ 1` to the v7 list. It causes too many bugs on code contributions and changes (most recent with RGBA in qimage). This is not the only occurrence.

src/framework/mlt_image.c
```diff
@@ -0,0 +1,374 @@
...
+  * \public \memberof mlt_image_s
+  * \param format the image format
+  * \param width width of the image in pixels
+  * \param height height of the image in pixels
+  * \param[out] bpp the number of bytes per pixel (optional)
+  * \return the number of bytes
+  */
+int mlt_image_format_size( mlt_image_format format, int width, int height, int *bpp )
+{
+	height += 1;
```
ddennedy, line 289: Here another height + 1

ddennedy:
This is a nice start by simply refactoring. Let's outline planned additions:
1. An opaque pointer for a AVFrame, AVPicture, GL texture, other unknown mappings for example something in ArrayFire or OpenCV. Could we need more than 1?
2. functions for converting planes and rows between unaligned and some #bytes aligned

Do you want to simply complete the refactoring in this branch and then merge it before adding things? I think we should.

src/framework/mlt_image.c
bmatherly: Yes! I am eager to do that, too. But how do we know what bugs this avoided so that we can fix them? Do we just remove the +1 and wait for bugs to present themselves?

bmatherly:
In addition to your points, I also had in mind:

1. Add C++ wrappers and unit tests
2. Move all alpha support into image class
3. Add other image attributes like colorspace and transfer coeffs

I had planned to add the C++ wrappers and simple unit tests to this branch before it is merged, but I do not feel strongly about that.

Can you explain what you mean by "Do you want to simply complete the refactoring..."?

src/framework/mlt_image.c
ddennedy: I did a git log message search on "height" and I did not find the reason for it, but I did find several crash fixes to workaround its unexpected result. I also searched my gmail, and I only found my vague answer to someone's question about it:

> Some image processing operations that are data-parallel (multiple byte) may overshoot the logical end of the image buffer.

It might be my old scaler in the gdk module or some of the things that use "Duff's device." I do not recall because it goes back so far! It seems to cause more harm than good, and there might be regressions, but we need to make this change some time, and it seems better to do it now than in a supposed stable branch.

I finally traced it back to my commit from 2004
https://github.com/mltframework/mlt/commit/74268cd86c30a0af9f03981c799aa4bba253c6e9

In this early development we made rather large commits without a lot of details in commit messages or comments while just cranking out the code. Now, I recall there was something related to field order conversion that involved shifting the image or something. Check this out, in particular "// Offset the image pointer by one line":

https://github.com/mltframework/mlt/commit/74268cd86c30a0af9f03981c799aa4bba253c6e9#diff-4ef9f584ebce981fe664a34a8222b8c6464718df6ed21aa2acce0a7ce5c6b0e7L41

At this time we were also trying to optimize things to meet our goals for the customer with a tight deadline and no certainty we would be doing anything further with MLT afterwards. This code is long since removed and not how we do field order correction today.

ddennedy:
> Can you explain what you mean by "Do you want to simply complete the refactoring..."?

I meant converting other services to use this like you did spot remover.

bmatherly:
>> Can you explain what you mean by "Do you want to simply complete the refactoring..."?

> I meant converting other services to use this like you did spot remover.

I converted spot remover as an example to work from. I am willing to convert one or two more services to give more/better examples if you have some to suggest. But I would prefer to wait to convert all services until we can do the image and audio classes at the same time. Also, I would like the unit tests to be done before converting a bunch of classes. 

Services are not required to be converted until we change mlt_frame_get_image() and mlt_frame_get_audio() to use the structures. So that could all be done at once.

src/framework/mlt_image.c
bmatherly: I support this. I added it to the task list: https://github.com/mltframework/mlt/issues/655
I agree that we may just have to wait for the regressions to be presented to us. But 7.0 is the time to break it.

PR #662, opened 2021-02-21
mlt_event_data

ddennedy:
For review. Some comments and observations:

1. `mlt_transmitter` was basically useless. I guess it could be used for some injection purpose, but after at least 15 years of being nothing but an intermediary and unnecessary complication, I am killing it.
2. 90% of existing events have 0 or 1 event argument. 80% of events with 1 arg are either integer, string, or `mlt_frame`. These had a big impact on the design of the new API.
4. custom structs are provided for events with multiple args. The framework defines 2: `mlt_event_data_thread` and `mlt_factory_event` (I think the latter should be renamed `mlt_event_data_factory` now that I write this.)
5. Even though the private struct currently only has a single union member, I intentionally did not expose `mlt_event_data` as a pointer to union to make it more flexible to later changes and keep things hidden.
5. I did brief testing with SWIG and ruby, but the current callback into Ruby causes a stack overflow even on master branch. I was able to verify some functionality with `printf` up to the point of callback from C++ to Ruby. Maybe something changed in the Ruby C API over the years; this is seldom used or tested by me.
6. The design is more convenient for SWIG because that works well when functions return specific types instead of defining a lot of custom code per language to convert each type of event to something for each language. `Mlt::Event::get_int/string/frame()` provides 90% coverage, and the rest probably not needed. For example, the Ruby code in `mlt.i` might be reduced to a single generic class if the ruby-wrapped `Mlt::EventData` can be passed to the ruby Proc.
6. Tested with Shotcut as well - minimal changes in a `mlt_event_data` branch.
7. I need to add more Doxygen comments.


src/framework/mlt_events.c
```diff
@@ -59,11 +59,20 @@ typedef struct mlt_events_struct *mlt_events;
...
-	mlt_events owner;
+	mlt_events parent;
 	int ref_count;
 	int block_count;
 	mlt_listener listener;
-	void *service;
+	void *listener_data;
+};
+
+struct mlt_event_data_s {
```
bmatherly, line 69: One suggestion to avoid the heap allocation and free() function: We could obfuscate this type and put it in the header. Then, perform the reinterpretation in the get/set functions by casting. For example:
```
struct mlt_event_data {
    union {
        int i;
        void* p;
    } u;
};
```
The size will be either 32 or 64 bits depending on the platform and could be passed by value instead of as a pointer.

The resulting functions would become:

```
mlt_event_data_set_int(mlt_event_data event_data, int value)
{
    event_data.i = int;
}

int mlt_event_data_get_int(mlt_event_data event_data)
{
    return event_data.i;
}

mlt_event_data_set_string(mlt_event_data event_data, const char *value)
{
    event_data.p = (void*)value;
}

const char *mlt_event_data_get_string(mlt_event_data event_data)
{
    return (const char*)event_data.p;
}
```


If we like this, maybe it should be a generic type defined in mlt_types named "mlt_opaque_data".

bmatherly (commented):
This is a great improvement. I see many advantages. My two concerns are:
1) Heap allocation (I have made enough comments about that)
2) Documentation: There is not a good way to tie the event name to the data type it provides. When I receive an event, how shall I know to call get_int() or get_other()? And if get_other(), what do I cast that to? The only way to know is through code inspection. Maybe we could add some text to the doxygen comments somehow or create an event naming idiom that hints at the type. I would be interested in your thoughts on this.

src/framework/mlt_events.c
ddennedy: I agree with the concerns about the heap allocations and will factor in your ideas except `mlt_opaque_data`.

src/framework/mlt_factory.h
```diff
@@ -60,4 +60,10 @@ extern void mlt_factory_register_for_clean_up( void *ptr, mlt_destructor destruc
 extern void mlt_factory_close( );
 extern mlt_properties mlt_global_properties( );
 
+typedef struct {
+	const char *name;
+	const void *input;
+	void *service;
+} mlt_factory_event;
```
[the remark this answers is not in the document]
ddennedy: Renaming this to `mlt_factory_event_data`. mlt_factory is very specific and constrained. I do not anticipate the need for something else, and if it ever comes up the new will be made unique.

src/framework/mlt_consumer.h
```diff
@@ -145,4 +145,13 @@ extern void mlt_consumer_stopped( mlt_consumer self );
...
 extern mlt_position mlt_consumer_position( mlt_consumer );
 
+typedef void* ( *mlt_thread_function_t )( void* );
+
+typedef struct {
+	void **thread;
+	int *priority;
+	mlt_thread_function_t function;
+	void *data;
+} mlt_event_data_thread;
```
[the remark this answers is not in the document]
ddennedy: Moving this to `mlt_events.h` because this could possibly be used by mlt_slices some day, and moving `mlt_thread_function_t` to `mlt_types.h`.

src/framework/mlt_events.h
```diff
@@ -58,5 +51,15 @@ extern void mlt_event_block( mlt_event self );
 extern void mlt_event_unblock( mlt_event self );
 extern void mlt_event_close( mlt_event self );
 
+extern mlt_event_data mlt_event_data_set_int(int value);
```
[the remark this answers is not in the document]
ddennedy: Had considered using "init" & "get" similar to other classes in the framework. However, going to use "from" and "to" as this area is different and reads nicely.

src/framework/mlt_events.h
```diff
@@ -58,5 +51,15 @@ extern void mlt_event_block( mlt_event self );
...
+extern int mlt_event_data_get_int(mlt_event_data);
+extern mlt_event_data mlt_event_data_set_string(const char *value);
+extern const char* mlt_event_data_get_string(mlt_event_data);
+extern mlt_event_data mlt_event_data_set_frame(mlt_frame);
+extern mlt_frame mlt_event_data_get_frame(mlt_event_data);
+extern mlt_event_data mlt_event_data_set_other(void*);
+extern void* mlt_event_data_get_other(mlt_event_data);
+extern void mlt_event_data_free(mlt_event_data);
```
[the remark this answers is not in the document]
ddennedy: accepted

src/framework/mlt_events.c
```diff
@@ -59,11 +59,20 @@ typedef struct mlt_events_struct *mlt_events;
...
+	union {
+		int integer;
+		const char *string;
+		mlt_frame frame;
+		void *other;
```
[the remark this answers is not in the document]
ddennedy: Going to use `mlt_event_data_from_object()` and `mlt_event_data_to_object()`.

src/framework/mlt_events.c
```diff
@@ -176,47 +174,33 @@ int mlt_events_register( mlt_properties self, const char *id, mlt_transmitter tr
...
-
  * \public \memberof mlt_events_struct
  * \param self a properties list
  * \param id the name of an event
+ * \param data an opaque pointer
  * \return the number of listeners
  */
 
-int mlt_events_fire( mlt_properties self, const char *id, ... )
+int mlt_events_fire(mlt_properties self, const char *id, mlt_event_data *event_data)
```
bmatherly, line 184: Suggestion: pass event_data by value

`int mlt_events_fire(mlt_properties self, const char *id, mlt_event_data event_data)`

src/framework/mlt_factory.c
```diff
@@ -354,13 +324,20 @@ mlt_producer mlt_factory_producer( mlt_profile profile, const char *service, con
 		service = mlt_environment( "MLT_PRODUCER" );
 
 	// Offer the application the chance to 'create'
-	mlt_events_fire( event_object, "producer-create-request", service, resource, &obj, NULL );
+	mlt_factory_event_data data = {
```
bmatherly, line 327: I do not  feel too strongly about these comments. But if you accept my suggestion to pass event_data by value, then these lines:

```
	mlt_factory_event_data data = {
	    .name = service,
	    .input = resource,
	    .service = &obj
	};
	mlt_event_data event_data;
	mlt_event_data_from_object(&event_data, &data);
	mlt_events_fire( event_object, "producer-create-request", &event_data );
```

Collapse to:

```
	mlt_factory_event_data data = {
	    .name = service,
	    .input = resource,
	    .service = &obj
	};
	mlt_events_fire( event_object, "producer-create-request", mlt_event_data_from_object(data) );
```

src/tests/test_events/test_events.cpp
```diff
@@ -44,10 +44,10 @@ class TestEvents : public QObject
         QVERIFY(owner == m_properties);
     }
 
-    static void onPropertyChanged(mlt_properties owner, TestEvents* self, char* name)
+    static void onPropertyChanged(mlt_properties owner, TestEvents* self, mlt_event_data* data)
     {
         QVERIFY(self != nullptr);
-        QCOMPARE(name, "foo");
+		QCOMPARE(Mlt::EventData(data).to_string(), "foo");
```
bmatherly, line 50: Did the whitespace get messed up here (and below)?

bmatherly (commented):
This looks fantastic. The documentation is great and satisfies my need. Here are a few minor comments which you are free to ignore.

src/framework/mlt_events.c
ddennedy: There are a lot of events with no data and currently pass NULL for this parameter. These would all have to be changed to pass basically a mlt_event_data somehow initialized with a NULL like
`mlt_event_data _from_object(NULL)`
or a new `mlt_event_data_null()`?

Does this change then motivate changing every `mlt_listener` to pass by value?
`typedef void ( *mlt_listener )( mlt_properties, void*, mlt_event_data );`
And then changing every all `mlt_event_data_to_...()` to pass by value? Finally, making all the changes everywhere these are used?

It is possible; I am in the middle of doing all that. It does make it easier and less verbose to use at the expense of more being more verbose when event data is null.

src/tests/test_events/test_events.cpp
ddennedy: yes, fixed locally

src/framework/mlt_factory.c
ddennedy: I made all of these changes, and it is working. Overall I think it is cleaner, less pointer checks, Mlt::EventData no longer needs an `is_valid()`. Still reviewing all my changes before pushing it.

src/framework/mlt_events.c
ddennedy: I made all of these changes, and it is working. Overall I think it is cleaner, less pointer checks, Mlt::EventData no longer needs an `is_valid()`. Still reviewing all my changes before pushing it.

bmatherly: Sounds promising.

For the "no data" cases, we could make a convenience function for that to set the value to 0:

```
mlt_event_data mlt_event_data_none()
{
    mlt_event_data event_data;
    event_data.u.i = 0;
    return event_data;
}
```

Depending on the patterns you see, you could also decide to specialize the fire function for each type:

```
mlt_events_fire(mlt_properties self, const char *id)
mlt_events_fire_data(mlt_properties self, const char *id, mlt_event_data *event_data)
mlt_events_fire_int(mlt_properties self, const char *id, int value)
mlt_events_fire_frame(mlt_properties self, const char *id, mlt_frame frame)
```

Based on your comments, it sounds like we may have already found the sweet spot.

PR #664, opened 2021-02-21
Delete the animation string after parsing it

bmatherly:
In some cases, the string would be re-parsed and changes to the animation would be lost.

While testing Shotcut from the MLTv7 branch, I see a problem when when I open a .mlt file with filters having animations - when trying to manipulate those keyframes they will stubbornly return to the initial state from when the file was opened. Debugging revealed that the initial string persists and parsing it would overwrite any changes made to the animation.

I do not know if some recent changes may have exposed this. But intuitively, the change makes sense to me. Maybe you know of a reason that the original string should be kept after it has been parsed.

### DIFF
--- a/src/framework/mlt_property.c
+++ b/src/framework/mlt_property.c
@@ -1322,6 +1322,11 @@ static void refresh_animation( mlt_property self, double fps, locale_t locale, i
 		self->animation = mlt_animation_new();
 		self->serialiser = (mlt_serialiser) mlt_animation_serialize_tf;
 		mlt_animation_parse( self->animation, self->prop_string, length, fps, locale );
+		// Delete the string and interact with the animation from now on.
+		self->types &= !mlt_prop_string;
+		if ( self->prop_string )
+			free( self->prop_string );
+		self->prop_string = NULL;
 	}
 	else if ( ( self->types & mlt_prop_string ) && self->prop_string )
 	{
--- a/src/framework/mlt_property.c
+++ b/src/framework/mlt_property.c
@@ -496,6 +496,8 @@ int mlt_property_get_int( mlt_property self, double fps, locale_t locale )
 		result = ( int ) ( (mlt_rect*) self->data )->x;
 	else if ( ( self->types & mlt_prop_string ) && self->prop_string )
 		result = mlt_property_atoi( self, fps, locale );
+	else if ( self->animation )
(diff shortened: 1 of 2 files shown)


src/framework/mlt_property.c
```diff
@@ -1322,6 +1328,11 @@ static void refresh_animation( mlt_property self, double fps, locale_t locale, i
 		self->animation = mlt_animation_new();
 		self->serialiser = (mlt_serialiser) mlt_animation_serialize_tf;
 		mlt_animation_parse( self->animation, self->prop_string, length, fps, locale );
+		// Delete the string and interact with the animation from now on.
+		self->types &= !mlt_prop_string;
```
ddennedy, line 1332: I think this needs to be bitwise not operator instead of logical not.
https://en.wikipedia.org/wiki/Bitwise_operations_in_C#Bitwise_NOT_~_/_ones'_complement_(unary)

ddennedy:
I think I was trying to maintain the design principle of mlt_property that it does not clear a stored string when getting it as an int or double, and instead we convert the string. And maybe also optimizing away the need to reserialize the animation simply because some filters were making anim_get calls. The only downside I can think of is that a user-supplied string is destroyed simply because a filter makes an anim_get call and no longer passes-through to XML serialization as supplied. But is that really an issue?

src/framework/mlt_property.c
bmatherly: Good catch. Will fix if we go with this method.

bmatherly:
> I think I was trying to maintain the design principle of mlt_property that it does not clear a stored string when getting it as an int or double, and instead we convert the string. And maybe also optimizing away the need to reserialize the animation simply because some filters were making anim_get calls.

Another option I thought of: I could add a "dirty bit" to the animation. Any call on the animation to modify it will set the dirty bit. When any call is made to read the property, the property would first check the dirty bit and re-serialize the string if the animation had changed. I am willing to implement this if you are interested.

> The only downside I can think of is that a user-supplied string is destroyed simply because a filter makes an anim_get call and no longer passes-through to XML serialization as supplied. But is that really an issue?

Maybe if the string fails to parse it would be desirable to keep it un changed? I think the dirty bit idea could support this since the string would only be reparsed if the animation was modified.

ddennedy:
Yes, let's do the dirty bit. Thanks

PR #665, opened 2021-02-22
Builtin typewriter effect for kdenlive titler

rlalik:
Following request and discussion of kdenlive community and devs: https://invent.kde.org/multimedia/kdenlive/-/issues/789#note_190007
I also created TW as a built in feature of the titler. It allows to apply the effect to a single Text item in the titler (the effect applies to all).

### DIFF
--- a/src/modules/qt/filter_typewriter.cpp
+++ b/src/modules/qt/filter_typewriter.cpp
@@ -28,23 +28,9 @@
 #include <vector>
 
 #include "typewriter.h"
+#include "kdenlivetitle_wrapper.h"
 
-struct producer_ktitle_s
-{
-    struct mlt_producer_s parent;
-    uint8_t *rgba_image;
-    uint8_t *current_image;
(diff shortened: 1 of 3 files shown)


evertvorster:
Hi there. 
This commit re-produces behavior that was lost in Kdenlive since a re-write some years ago. It has been one of the most-requested features for the longest time. 
I have tested this patch personally, and it works for me. For whatever that is worth. :)

ddennedy:
Compile error in automated build based on Qt 5.12 in Ubuntu 20.04 LTS, which must be supported:
```
kdenlivetitle_wrapper.cpp: In member function ‘void PlainTextItem::updateText(QString)’:
kdenlivetitle_wrapper.cpp:170:16: error: ‘class QPainterPath’ has no member named ‘clear’
```

This method was added in 5.13: https://doc.qt.io/qt-5/qpainterpath.html#clear

PR #667, opened 2021-02-25
Clear animation cache

bmatherly:
This is an alternative proposal to solve https://github.com/mltframework/mlt/pull/664

In this proposal, the initial string is preserved unless the animation is changed. If the animation is changed, the string is destroyed and recreated if/when needed.

All the proposed tests in this patch failed before the changes.

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -594,4 +594,5 @@ MLT_7.0.0 {
     mlt_link_close;
     mlt_repository_links;
     mlt_animation_shift_frames;
+    mlt_animation_get_string;
 } MLT_6.22.0;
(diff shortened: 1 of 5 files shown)


ddennedy:
Sorry, but I do not like a public function `mlt_animation_get_cache()`

bmatherly:
I'm not in love with it either. I like that I can detect the strings have become out of sync without adding an additional member to the animation.

How about if I change it to:
`mlt_animation_get_string()`
or
`mlt_animation_get_serial_string()`
?

Else, I will add a member variable and a function named mlt_animation_is_modified(). But I am not in love with that either because it does not indicate since when it was modified.

ddennedy:
Let's go with `mlt_animation_get_string()`

bmatherly:
> Let's go with mlt_animation_get_string()

Done

PR #668, opened 2021-03-02
Remove data_feed data_show and feeds

bmatherly:
I am unsure if these are still used or if the comment should be changed to not reference "data_feed":
```
 * \properties \em global_feed a flag to indicate whether this tractor feeds to the consumer or stops here
 * \properties \em global_queue is something for the data_feed functionality in the core module
 * \properties \em data_queue is something for the data_feed functionality in the core module
```
https://github.com/mltframework/mlt/blob/83b6b806c41af779cba5c6763b9bda1cfb5f51cb/src/framework/mlt_tractor.h#L37


ddennedy:
"global_feed", "global_queue", and "data_queue" are all related to the removed data_feed filter. Basically lines 567 - 586 in `mlt_tractor.c` can be removed.

bmatherly:
> "global_feed", "global_queue", and "data_queue" are all related to the removed data_feed filter. Basically lines 567 - 586 in mlt_tractor.c can be removed.

Done. This even cleans up the XML a little bit.

src/modules/xml/mlt-xml.dtd
```diff
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 
-<!-- MLT XML DTD v0.2.0 -->
```
ddennedy, line 3: I am thinking perhaps we should change the DTD version to a major and minor only that reflects the MLT version. We do not need to increment it with each minor version of MLT, but anytime we break the schema by changing it in a non-backwards compatible way, it should coincide with a major new version of MLT. And anytime we add something in a backwards compatible way or fix the schema bump the minor version. The XML interface is nearly as important as the ABI. See how in Shotcut I am now using the MLT major version reflected in the MLT XML to check compatibility.

bmatherly: I also need to add chain and link types to the dtd before the official release.

Shall I change the DTD version to 7.0.0?

ddennedy: yes, change it to 7.0.0

bmatherly: Done

PR #673, opened 2021-03-11
Fix #663

### LINKED ISSUE #663: Image sequence does not seem to properly handle images of different sizes
I reported this issue to kdenlive here: https://bugs.kde.org/show_bug.cgi?id=431376. When creating a sequence of images, it seems that an image is initially shown with the same size of the previous one, despite the fact that the size is different.

This is a test set I prepared to reproduce the issue: https://app.box.com/s/df55s9ldew4sj4nyk6t9h7k7u96fjtuw.
This is the video prepared by kdenlive:
* Video without dissolve transition: https://app.box.com/s/rcgtxwojmjxnleava475f7eehkogjagy
* Video without any animation: https://app.box.com/s/uaqlftg0sb6xwhc7meiqdu8boi0l123z

As you can see, I was told that the melt command also shows the same behavior. Is this the proper place to report the issue?
Thanks!

carlonluca:
TTL should be searched for in producer props instead of frame props.

### DIFF
--- a/src/modules/qt/producer_qimage.c
+++ b/src/modules/qt/producer_qimage.c
@@ -337,7 +337,7 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
 		mlt_frame_set_position( *frame, mlt_producer_position( producer ) );
 
 		// Refresh the image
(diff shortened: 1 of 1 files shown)


ddennedy:
Indeed! it was a regression in 27361eba. Thank you

PR #687, opened 2021-03-23
V7/cmake: jack fix, .def, feature_summary

eszlari:


src/modules/jackrack/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
-add_library(mltjack MODULE consumer_jack.c factory.c)
+add_library(mltjack MODULE factory.c)
 
 target_compile_options(mltjack PRIVATE ${MLT_COMPILE_OPTIONS})
 
-target_link_libraries(mltjack PRIVATE mlt Threads::Threads PkgConfig::jack)
+target_link_libraries(mltjack PRIVATE mlt Threads::Threads)
+
+if(TARGET PkgConfig::jack)
```
ddennedy, line 7: How does one explicitly disable this dependency without removing the jack.pc file?

eszlari: I could add a option, but in what case is this useful?

ddennedy: It is mainly useful for testing the build process in different configurations but also curious in case it is ever needed. Let's wait and see.

eszlari:
@ddennedy

`find_package()` is now used to search for JACK

You can use `CMAKE_DISABLE_FIND_PACKAGE_<PackageName>=ON` to disable each dependency.

PR #690, opened 2021-03-28
[WIP] Frame image

bmatherly:
For discussion: not ready for commit.

This change is a proof of concept. It would be a necessary step if eventually we want mlt_frame_get_image() to use mlt_image (same for audio).

Some questions for this review:
* Are we interested in this approach overall?
* Is MLT7 good timing for a change this large (perhaps the window has closed). I would like to change mlt_frame_get_*() to use mlt_image and mlt_audio for MLT7. 

Notice that this concept provides a backwards compatible form of mlt_frame_get_image() so that every single service does not have to be changed (just the ones that modify properties directly).

### DIFF
--- a/src/framework/mlt_frame.c
+++ b/src/framework/mlt_frame.c
@@ -53,12 +53,9 @@ mlt_frame mlt_frame_init( mlt_service service )
 
 		// Set default properties on the frame
 		mlt_properties_set_position( properties, "_position", 0.0 );
-		mlt_properties_set_data( properties, "image", NULL, 0, NULL, NULL );
-		mlt_properties_set_int( properties, "width", profile? profile->width : 720 );
-		mlt_properties_set_int( properties, "height", profile? profile->height : 576 );
+		self->image.width = profile ? profile->width : 720;
+		self->image.height = profile? profile->height : 576;
 		mlt_properties_set_double( properties, "aspect_ratio", mlt_profile_sar( NULL ) );
-		mlt_properties_set_data( properties, "audio", NULL, 0, NULL, NULL );
-		mlt_properties_set_data( properties, "alpha", NULL, 0, NULL, NULL );
 
 		// Construct stacks for frames and methods
 		self->stack_image = mlt_deque_init( );
@@ -91,8 +88,7 @@ mlt_properties mlt_frame_properties( mlt_frame self )
 int mlt_frame_is_test_card( mlt_frame self )
 {
 	mlt_properties properties = MLT_FRAME_PROPERTIES( self );
-	return ( mlt_deque_count( self->stack_image ) == 0
-			 && !mlt_properties_get_data( properties, "image", NULL ) )
+	return ( mlt_deque_count( self->stack_image ) == 0 && !self->image.data )
 			|| mlt_properties_get_int( properties, "test_image" );
 }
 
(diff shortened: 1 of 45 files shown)


src/framework/mlt_frame.c
```diff
@@ -334,7 +330,9 @@ mlt_deque mlt_frame_service_stack( mlt_frame self )
 
 int mlt_frame_set_image( mlt_frame self, uint8_t *image, int size, mlt_destructor destroy )
```
bmatherly, line 331: This demonstrates how we can deprecate the old function to keep existing code mostly working until it is eventually converted.

src/framework/mlt_frame.c
```diff
@@ -887,29 +882,7 @@ mlt_frame mlt_frame_clone( mlt_frame self, int is_deep )
...
-			if ( data )
-			{
-				if ( ! size )
-					size = width * height;
-				copy = mlt_pool_alloc( size );
-				memcpy( copy, data, size );
-				mlt_properties_set_data( new_props, "alpha", copy, size, mlt_pool_release, NULL );
-			};
-		}
+		mlt_image_copy_deep( &self->image, &new_frame->image );
```
bmatherly, line 885: This is a good demonstration of how a lot of duplicated code can be replaced with single function calls.

src/framework/mlt_frame.h
```diff
@@ -97,6 +97,8 @@ struct mlt_frame_s
 	mlt_deque stack_audio;   /**< \private the audio processing stack of operations and data */
 	mlt_deque stack_service; /**< \private a general purpose data stack */
 	int is_processing;       /**< \private indicates if a frame is or was processed by the parallel consumer */
+	struct mlt_image_s image;        /**< \private the frame image */
```
bmatherly, line 100: This is the heart of the entire change set. I added mlt_image to the public interface. I see here that I labeled it private. But I access it directly in many modules. Maybe there should be a get function instead of direct access.

I also added mlt_audio to the frame for illustration.

bmatherly (commented):
I did test this with Shotcut. There were a few tweaks to make it work. There are some bugs I would have to work out if we want to move forward with this in the short term.

src/framework/mlt_frame.h
ddennedy: We should make a private member of the mlt_frame struct like we have done elsewhere and put things into it including:
- stack_image
- stack_audio
- stack_service
- is_processing
- image
- audio

ddennedy:
> I did test this with Shotcut. There were a few tweaks to make it work. There are some bugs I would have to work out

This statement and its size and scope of the change concerns me. Can we separate this into one change that affects the v7 API in a manner we can live with (making things private to help), and then 2 others that implement and integrate changes?

bmatherly:
> Can we separate this into one change that affects the v7 API in a manner we can live with (making things private to help), and then 2 others that implement and integrate changes?

That is good insight. I think we can reduce the risk by making the changes more incrementally.

Some increments would be (one commit each):
1. Do not access "image", "alpha", and "audio" properties directly. Always access them with the existing get_image/audio() functions. This is a large part of the changes in this commit. It could be broken into one commit per module.
2. Make a private struct for the frame
3. Move the stacks into the private struct
4. Add new get/set image() and audio() functions to return image and audio structures that are populated on demand from the existing "image" and "audio" properties.
5. Deprecate the old get/set functions and accessing width/height/samples properties directly
6. Release MLT 7.0 - at this point, all the current methods still work but are deprecated
7. Modify MLT services one at a time to use the new get/set functions and stop accessing properties directly
8. When all services have been converted, add image/audio structs to the private frame structure and remove the deprecated functions
9. Release MLT 7.1

PR #702, opened 2021-04-14
master/6.26 cmake changes to build Kdenlive on Craft/mingw

vpinon:
Hello,
I had to apply these changes to build our Kdenlive windows release with 6.26.1 cmake.
- we don't build gtk2, and the check was missing
- some include were missing (already in v7 branch IIRC)
- the lib renaming with SO version is messing Kdenlive's cmake FindMLT.cmake => we will have to evolve this for v7, but maybe not yet for v6 ?


ddennedy:
I merged this into a new v6 branch. Very soon I am soon going to merge v7 into master, and I do not want to avoid conflicts.

PR #706, opened 2021-04-26
Add speed parameter to timer filter

emlun:
Hi! I'm currently using Shotcut to edit a timelapse video and wanted the ability to add a sped-up "real-time" timer to the timelapse. I traced that back to here and was able to successfully plumb this through the Shotcut UI and implement the feature. Do you want it?

This is my first contact with this library, so please let me know if there's something more I should do for this (like add tests, or translation keys?).

### DIFF
--- a/src/modules/plus/filter_timer.c
+++ b/src/modules/plus/filter_timer.c
@@ -38,7 +38,7 @@ double time_to_seconds( char* time )
 static void get_timer_str( mlt_filter filter, mlt_frame frame, char* text )
 {
 	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
-	mlt_position current_frame = mlt_filter_get_position( filter, frame );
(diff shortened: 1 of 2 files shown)


emlun:
Parallel pull request in Shotcut: https://github.com/mltframework/shotcut/pull/1080

emlun:
Ah, by the way, the reason this branches from tag `v6.24.0` instead of master is that I couldn't get Shotcut to compile against it if either strayed too far from the versions currently in Arch linux repositories...

ddennedy:
> Ah, by the way, the reason this branches from tag `v6.24.0` instead of master is that I couldn't get Shotcut to compile against it if either strayed too far from the versions currently in Arch linux repositories...

It is OK. It is based on master (v7) branch, and the build and unit tests passed.

PR #707, opened 2021-05-03
Use correct python installation path for binaries

lgbaldoni:


ddennedy:
How do we know this is correct?

lgbaldoni:
> How do we know this is correct?

From [cmake documentation](https://cmake.org/cmake/help/latest/module/FindPython3.html):
```
Python3_SITELIB
    Third-party platform independent installation directory.
Python3_SITEARCH
    Third-party platform dependent installation directory.
```

Since swig creates creates a `.so` as part of the python plugin, I would say it's platform dependent.

ddennedy:
Thanks! That makes sense as there can pure python architecture-independent modules.

PR #709, opened 2021-05-05
Honor DESTDIR when creating the melt symlink

antonio-rojas:
Otherwise it will try to create it on the host build system


ddennedy:
@eszlari  Can you please review this one?

PR #719, opened 2021-06-20
Fix clang LTO error

schw0reismus:
Based on https://stackoverflow.com/questions/14506151/invalid-symbol-redefinition-in-inline-asm-on-llvm and https://github.com/anrieff/libcpuid/issues/69.


ddennedy (approved):
I confirmed it compiles without error on gcc and still works after the change.

Issue #717, opened 2021-06-18
Problem with playing some 4K Clips (results in white image)


jlskuz:
A Kdenlive user reported a problem with the playback of some 4K clips ([here](https://invent.kde.org/multimedia/kdenlive/-/issues/1110) is the original report). I am reporting this here because also happens with the `melt` cmd (6.26.1 as well as 7.1.0). There is no specific pattern recognized as it occurs only with some clips from the same camera while others play as expected. Here is a file to reproduce: https://www.dropbox.com/s/x5c86t36g8jkfyq/NORM0012.MP4?dl=0
The cmd gives this error:
```
[producer avformat] /path/to/file/NORM0012.MP4
    video avcodec_send_packet failed with -22
```

bmatherly:
I posted one potential fix. It makes images appear with the provided clip. I do not know the side effects of this change. It may require a lot of testing.

bmatherly:
The previous commit was reverted due to causing an infinite loop in some cases.

Here is a narrower version which might avoid the infinite loop problem: https://github.com/bmatherly/mlt/commit/a8bd82a9f72161518a5695d2f09cef8521409e70

ddennedy:
FYI the loop/hang problem I previously reported was not due to this change, but I believe you if you find another.

bmatherly:
> FYI the loop/hang problem I previously reported was not due to this change, but I believe you if you find another.

I did not find a loop problem. I misunderstood your message and offered this alternate patch to see if it works for you.

If this patch does not cause a problem for you, then why did you revert it? https://github.com/mltframework/mlt/commit/320c5fc4d965960cfb6ddfcf6f255d4d07455ab8

ddennedy:
OK, sorry for the confusion. It appears that revert I had made locally I forgot to undo before making other changes and then having it get push inadvertently. I just reverted the revert for being simpler than the alternative.

jlskuz:
Thanks @bmatherly ! I asked the user who reported the problem to test, I will reopen in case the problem persists, for now I am closing this issue.

PR #721, opened 2021-06-23
Attempt to decode frame when EINVAL is received.

bmatherly:
This is a potential fix for https://github.com/mltframework/mlt/issues/717


ddennedy:
It makes sense that it is another invalid condition alongside the others.

ddennedy:
A problem that I found with this is that it can go into an infinite loop. I was testing in Shotcut, and it was working until I tried to add a filter.

PR #723, opened 2021-07-08
Fix timer filter speed interaction with start delay

emlun:
See: https://forum.shotcut.org/t/please-test-the-version-21-06-release-candidate/28284/27

The speed parameter added in PR #706 did not interact well with the start parameter, nor with duration set to zero. This PR makes the parameters interact sensibly, although this choice is up for debate as it is somewhat arbitrary:

- Start parameter is in playback time and is not affected by speed.
- Duration is in timer time; playback duration _is_ scaled by speed but the max timer value is _not_ affected by speed.
- Offset is in timer time and is not affected by speed.

So, for example, with start 5s, duration 30s, offset 7s and speed 10.0:

- The timer will start frozen with value 00:00:07.
- At playback time 00:00:05.000, the timer starts counting 10 seconds per 1 second of playback.
- At playback time 00:00:08.000, the timer stops and remains frozen with value 00:00:37.

I've also confirmed that this behaviour also works for timer direction "down".

### DIFF
--- a/src/modules/plus/filter_timer.c
+++ b/src/modules/plus/filter_timer.c
@@ -38,7 +38,8 @@ double time_to_seconds( char* time )
 static void get_timer_str( mlt_filter filter, mlt_frame frame, char* text )
 {
 	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
-	mlt_position current_frame = mlt_filter_get_position( filter, frame ) * mlt_properties_get_double( properties, "speed" );
+	double filter_speed = mlt_properties_get_double( properties, "speed" );
+	mlt_position current_frame = mlt_filter_get_position( filter, frame ) * filter_speed;
 	char* direction = mlt_properties_get( properties, "direction" );
 	double timer_start = time_to_seconds( mlt_properties_get( properties, "start" ) );
 	double timer_duration = time_to_seconds( mlt_properties_get( properties, "duration" ) );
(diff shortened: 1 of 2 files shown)


bmatherly:
LGTM

ddennedy:
One problem with this is that it changes behavior and will affect existing projects where someone has manually compensated for this. Is there a way to do this outside MLT so that the Shotcut UI factors speed into the properties it sets on this filter? If not, then I think we need to a boolean parameter (int property) to indicate whether the speed factors in these other expressions. Then, the Shotcut UI could set this property 1 only for newly added filters.

emlun:
Hm... will it though? The affected feature was added in commit 58971d02 which comes after the latest release tag `v7.0.1`. So it seems like the feature has not yet been officially released, right?

ddennedy:
That is correct, but newest Shotcut version already includes it. Overall it is very limited and probably not worth addressing.

emlun:
Alright :slightly_smiling_face: But just for clarity: "not worth addressing" in what sense? Since this was merged I assume you mean "the change of behaviour should have such limited impact we won't worry about backwards compatibility for this"?

bmatherly:
> Alright 🙂 But just for clarity: "not worth addressing" in what sense? Since this was merged I assume you mean "the change of behaviour should have such limited impact we won't worry about backwards compatibility for this"?

I assume that, too. And I agree. Backwards compatibility is not critical in this small case.

PR #728, opened 2021-07-26
Obscure filter: replace geometry by rect

jlskuz:
This makes the filter keyframable.

...but it also brakes backward compatibility. I don't know what the best solution is or how it is usually done in such cases? Should this be moved to a new filter or should the filter support both geometry and rect?

### DIFF
--- a/src/modules/core/filter_obscure.c
+++ b/src/modules/core/filter_obscure.c
@@ -20,115 +20,67 @@
 #include <framework/mlt_filter.h>
 #include <framework/mlt_frame.h>
 #include <framework/mlt_profile.h>
+#include <framework/mlt_log.h>
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


ddennedy:
`obscure` is a legacy filter. It is not yet deprecated, but I am not certain it should be extended. Certainly we should not break compatibility. I would say that if anyone wants animation here they should instead use the masking filters in conjunction with other filters:

1. `mask_start` with `frei0r.alphaspot`
2. other video filters such as `frei0r.pixeliz0r` or a blur
3. `mask_apply`

This supports keyframes and is more generic and extensible (anything can shape the alpha channel). If you really want to add this to obscure, however, I will allow it by keeping the old properties.

If you want to see how this works in Shotcut, use in order:

- Mask: Simple Shape
- Mosaic
- Mask: Apply

PR #730, opened 2021-08-22
Fix data race in time remapping

### LINKED ISSUE #729: Data Race in Time Remap around `mlt_frame_get_image`
I ran into a problem where output frames would be corrupted when using time remapping on some 4K footage. Specifically, the source was a MP4 file, 3840x2160 at 29.8471 Hz, and the rendered output being 30 Hz. The `*.mlt` I was running came out of a `kdenlive` project. What I observed was the following:

- *Nearest* mode: Single frames would be corrupted (mostly greenish, arbitrary patterns, looked like something partially decoded)
- *Blend* mode: Looked like corrupted frames would be interpolated/blended with non-corrupted frames

This made me think that it's probably a problem of the data flowing into the time remapping getting corrupted, rather than anything after the remapping. Also, the problem with *nearest* seemed to mainly appear at the boundary from a sped-up section to a 100% section, and never within the sped-up section. With *blend* mode the corruption would appear in many more places.

Digging through the code and running a debug build of `melt` and specially `libmltcore` with my own tweaks/hacks, I realized the following: Locking and unlocking a `pthread_mutex` in `link_timeremap.c` before `link_get_image_blend`/`link_get_image_nearest` would call `mlt_frame_get_image` until after the image data has been copied/processed would completely get rid of the problem. This suggests that there might be a data race on the calls to `mlt_frame_get_image`, where the returned image buffer might be corrupted by a second call to the function in a concurrent thread.

I'm not familiar enough with the code base and how it uses threading to know where to look to fix this more properly than my ugly static mutex in the critical parts of the code. Are there any considerations related to this in the design of MLT, or does the time remapping hit a new corner case because it requires more random access to input frames? Would this require adding the ability to lock/unlock the returned image data somehow?

fabianschuiki:
~This is the hack referred to in #729 that got me around the frame corruption problem. Just for reference; not an actual fix :wink:.~

Fix an issue in `link_timeremap.c`, where multiple concurrent threads calling `mlt_frame_get_image` would cause the returned image data to be corrupted, or to be consumed incorrectly. As a result, the produced frames would be seemingly arbitrarily corrupted, more so in "blend" mode than in "nearest" mode, due to the increased number of frames fetched.

Add a `mlt_service_lock` and `mlt_service_unlock` call around the critical function calls to ensure data returned is intact.

Fixes #729.

### DIFF
--- a/src/modules/core/link_timeremap.c
+++ b/src/modules/core/link_timeremap.c
@@ -268,7 +268,10 @@ static int link_get_image_blend( mlt_frame frame, uint8_t** image, mlt_image_for
 		{
 			break;
 		}
-		if ( mlt_frame_get_image( src_frame, &images[image_count], format, &requested_width, &requested_height, 0 ) != 0 )
+		mlt_service_lock( MLT_LINK_SERVICE(self) );
+		int result = mlt_frame_get_image( src_frame, &images[image_count], format, &requested_width, &requested_height, 0 );
+		mlt_service_unlock( MLT_LINK_SERVICE(self) );
+		if ( result != 0 )
 		{
 			mlt_log_error( MLT_LINK_SERVICE(self), "Failed to get image %s\n", key );
 			break;
@@ -333,7 +336,10 @@ static int link_get_image_nearest( mlt_frame frame, uint8_t** image, mlt_image_f
 	if ( src_frame )
 	{
 		uint8_t* in_image;
-		if ( !mlt_frame_get_image( src_frame, &in_image, format, width, height, 0 ) )
+		mlt_service_lock( MLT_LINK_SERVICE(self) );
+		int result = mlt_frame_get_image( src_frame, &in_image, format, width, height, 0 );
+		mlt_service_unlock( MLT_LINK_SERVICE(self) );
+		if ( !result )
 		{
 			int size = mlt_image_format_size( *format, *width, *height, NULL );
 			*image = mlt_pool_alloc( size );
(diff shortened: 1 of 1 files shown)


bmatherly:
Thank you for the excellent analysis and deduction! You may be able to help us solve a nagging issue that I have not been able to reproduce:
https://forum.shotcut.org/t/export-issue-with-21-05-01-version-and-time-remap/27382

We prefer to allow as much parallel processing as possible. As an experiment, could you try only putting the lock around the call to mlt_frame_get_image() and releasing immediately after?

Also, to make it not be a "hack", you can call mlt_service_lock() and mlt_service_unlock(). Those functions are intended to synchronize critical sections in MLT services - situations exactly like this one.
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_service.h#L81

fabianschuiki:
Happy to help! You were right, the problem was indeed not the data behind the image pointer being corrupted, but more narrowly the call to `mlt_frame_get_image` itself. Replaced the pthread hack with `mlt_service_[un]lock` as you suggested, which should make this a more or less proper fix :smiley:.

src/modules/core/link_timeremap.c
```diff
@@ -333,7 +336,10 @@ static int link_get_image_nearest( mlt_frame frame, uint8_t** image, mlt_image_f
 	if ( src_frame )
 	{
 		uint8_t* in_image;
-		if ( !mlt_frame_get_image( src_frame, &in_image, format, width, height, 0 ) )
+		mlt_service_lock( MLT_LINK_SERVICE(self) );
+		int result = mlt_frame_get_image( src_frame, &in_image, format, width, height, 0 );
+		mlt_service_unlock( MLT_LINK_SERVICE(self) );
+		if ( !result )
```
bmatherly, line 342: Sorry for the nit, but could you make the "if" statements match? I do not have a preference, just choose "!result" or "result != 0" so they match.

Then we can get this merged in just in time for the next release!

bmatherly (changes requested):

src/modules/core/link_timeremap.c
fabianschuiki: Done 👍

bmatherly (approved):

bmatherly:
Thanks again for this valuable contribution. I know some other users will be very happy this is resolved.

PR #731, opened 2021-09-03
Update filter_brightness.c

j-b-m:
Investigating an alpha channel issue with the brightness when processing a non rgba transition, I think there is a mistake here. 
The alpha channel for non rgba images was set to a size of 0, preventing it to be correctly processed. This patch fixes the problem


j-b-m:
Hmm, I should probably use proc_image.width and proc_image.height instead of *width and *height

PR #732, opened 2021-09-03
Restore simplified alpha transparency syntax

j-b-m:
Prior to MLT 7, the brightness filter could be used to create fade from/to alpha using a simplified syntax, like :
melt color:red -track myclip.mp4 -attach brightness out=100 start=0 end=1 alpha=-1 -transition frei0r.cairoblend
This creates a simple fade from transparent to opaque for the effect duration, without needing to introduce animated parameters (alpha follows the opacity defined by start/end).
This patch re-introduces the feature. Would you be ok for it ? I know that I can also use the full syntax by setting alpha="0=0;100=1" but the simplified syntax was much easier to handle... Anyways I would understand if you don't want to reintroduce this, just let me know.


ddennedy:
I am not OK with it. You can use -1 as the frame number to refer to the end.

j-b-m:
ok no problem.

PR #733, opened 2021-09-06
Fix implicite delaration of isspace()

jlskuz:
In `gps_parser.c` the compiler throws a implicit declaration warning for `isspace`. Depending on your compiler settings this can lead to an error. Including [`ctype.h`](https://www.cplusplus.com/reference/cctype/) fixes it.


vpinon:
I'm merging this simple patch, as it blocks our build...

ddennedy:
Sure, feel free to push simple obvious fixes.

PR #735, opened 2021-09-23
WIP: Nested properties

bmatherly:
This is a proposal to support the future markers feature in Shotcut. This feature adds the ability to nest properties inside properties. This allows to use the native MLT  APIs to store hierarchical data in the MLT XML.

Here is an example of what the marker storage would look like in the MLT file using this proposed implementation:

```
<?xml version="1.0" standalone="no"?>
<mlt LC_NUMERIC="C" version="7.1.0" title="Shotcut version 21.06.12" producer="main_bin">
  <profile description="HD 1080p 25 fps" width="1920" height="1080" progressive="1" sample_aspect_num="1" sample_aspect_den="1" display_aspect_num="16" display_aspect_den="9" frame_rate_num="24" frame_rate_den="1" colorspace="709"/>

... snip producers, playlists, etc...

  <tractor id="tractor0" title="Shotcut version 21.06.12" in="00:00:00.000" out="00:00:29.417">
    <property name="shotcut">1</property>
    <property name="shotcut:projectAudioChannels">2</property>
    <property name="shotcut:projectFolder">0</property>
    <properties name="shotcut:markers">
      <properties name="1">
        <property name="text">Marker 1</property>
        <property name="start">100</property>
        <property name="end">150</property>
        <property name="color">65280</property>
      </properties>
      <properties name="2">
        <property name="text">Marker 2</property>
        <property name="start">200</property>
        <property name="end">250</property>
        <property name="color">255</property>
      </properties>
      <properties name="3">
        <property name="text">Marker 3</property>
        <property name="start">300</property>
        <property name="end">350</property>
        <property name="color">16711680</property>
      </properties>
    </properties>
    <track producer="background"/>
    <track producer="playlist0"/>
    <transition id="transition0">
      <property name="a_track">0</property>
      <property name="b_track">1</property>
      <property name="mlt_service">mix</property>
      <property name="always_active">1</property>
      <property name="sum">1</property>
    </transition>
    <transition id="transition1">
      <property name="a_track">0</property>
      <property name="b_track">1</property>
      <property name="version">0.9</property>
      <property name="mlt_service">frei0r.cairoblend</property>
      <property name="threads">0</property>
      <property name="disable">1</property>
    </transition>
  </tractor>
</mlt>
```

I would like your comments on this general approach. And, if the approach is favorable, feel free to also comment on the code. Else, I can take a different approach.


src/mlt++/MltProperties.cpp
```diff
@@ -431,3 +431,18 @@ Animation *Properties::get_anim(const char *name)
 {
 	return new Animation( mlt_properties_get_animation( get_properties(), name ) );
 }
+
+int Properties::set( const char *name, Properties& properties )
+{
+	return mlt_properties_set_properties( get_properties(), name, properties.get_properties() );
+}
+
+Properties* Properties::get_props( const char *name )
```
bmatherly, line 440: I use "get_props" here because there is a name collision when Mlt::Service inherits Mlt::Properties. Mlt::Service already has a method named "get_properties". I'm open to suggestions here.

ddennedy:
Did you consider using a property name convention instead?

bmatherly:
> Did you consider using a property name convention instead?
> shotcut:marker:1:text
> Then, in the app you can simply split and join the string on :

If I understand your suggestion, the example above would become:

<property name="shotcut:marker:1:text">Marker 1</property>
<property name="shotcut:marker:1:start">100</property>
<property name="shotcut:marker:1:end">150</property>
<property name="shotcut:marker:1:color">65280</property>
<property name="shotcut:marker:2:text">Marker 2</property>
<property name="shotcut:marker:2:start">200</property>
<property name="shotcut:marker:2:end">250</property>
<property name="shotcut:marker:2:color">255</property>
<property name="shotcut:marker:3:text">Marker 3</property>
<property name="shotcut:marker:3:start">300</property>
<property name="shotcut:marker:3:end">350</property>
<property name="shotcut:marker:3:color">16711680</property>

I think that could work. I think there is an existing convention to use "." - for example "meta.media.width".

Advantage: no need to add this new nesting feature (and the complication)

Disadvantage: It will be more cumbersome in the code to discover, remove and insert entries. But perhaps that could be mitigated by moving the logic into helper functions. In general, I think it would result in more looping and string parsing.

Disadvantage: The XML structure does not reflect the structure of the data.

Do you have a preference for using a name convention to indicate the hierarchy? If so, I can try to adapt by Shotcut working copy to use that instead.

ddennedy:
Yes. Also, there is a convention to use ':' as the namespace delimiter and '.' as the delimiter for hierarchy. So, it could be like instead:
`<property name="shotcut:marker.1.text">Marker 1</property>`
etc.
I am a little hesitant to change the framework to accommodate app data. For example, one could also simply store JSON in a single string property. However, there could be a general usefulness for the ability to access and (de)serialize properties objects as you have implemented. So, I am still thinking about it.

bmatherly:
> one could also simply store JSON in a single string property

I think that would be my 3rd choice. I do not enjoy debugging serialized JSON mixed in XML. Also, I think there is some advantage for users to be able to easily inspect the mlt file. And embedded JSON would be another level of difficulty for human inspection.

> However, there could be a general usefulness for the ability to access and (de)serialize properties objects as you have implemented

I implemented it in a generic way in case there could be other applications for it. But I did not have anything specific in mind. Some ideas off the top of my head:
* Allow services to keep scalar properties in a more structured format
* Would be a nice way to store (and access) color components of a gradient property
* Allow application data to be stored in a structure/hierarchy

src/mlt++/MltProperties.cpp
ddennedy: But this method has a different signature and the one in Mlt::Service can overload this one if named the same?

src/tests/test_properties/test_properties.cpp
```diff
@@ -30,7 +30,7 @@ extern "C" {
 }
 #include <cfloat>
 
-static const bool kRunLongTests = true;
+static const bool kRunLongTests = false;
```
ddennedy, line 33: This should not be changed.

ddennedy (changes requested):

src/mlt++/MltProperties.cpp
bmatherly: That is also my understanding. I couldn't immediately get the compiler to find the child function. But the two functions have nothing in common except the name. So I am also concerned that the overloaded names might cause confusion for other developers. I can keep working to figure out the compiler issue if you prefer the overloaded function names.

src/tests/test_properties/test_properties.cpp
bmatherly: Good catch. In another file I also see some white space changes that are not required. I will fix those up.

src/mlt++/MltProperties.cpp
ddennedy: I agree it is not an appropriate usage of overloading. I am fine with the names you have used.

bmatherly:
I'm still on the fence about whether we should use this patch of switch to flat properties with "." (dot) notation. But I thought of another possible use. We could use these structures to store a "bin/folder" hierarchy if we ever add that to Shotcut.

ddennedy:
> store a "bin/folder" hierarchy if we ever add that to Shotcut

True but I am biased towards tags instead of hierarchical, mutually exclusive bins for this. I can still using this feature like:
```
<producer>
  <properties name="shotcut:tags">
    <property name="1">dog</property>
    <property name="2">beach</property>
  </properties>
</producer>
```

ddennedy:
FYI, I will accept this after minor cleanup if you choose to take the approach.

bmatherly:
> FYI, I will accept this after minor cleanup if you choose to take the approach.

I have been testing this approach with the WIP markers and I like it. I cleaned up a few things. Feel free to merge to master if you like it.

PR #736, opened 2021-09-25
Fix artifacts in composite transition

j-b-m:
When no luma file was specified, the compositing was performed twice, leading to some artifacts
https://invent.kde.org/multimedia/kdenlive/-/issues/670

### DIFF
--- a/src/modules/core/transition_composite.c
+++ b/src/modules/core/transition_composite.c
@@ -144,6 +144,7 @@ void composite_line_yuv( uint8_t *dest, uint8_t *src, int width, uint8_t *alpha_
 			alpha_a += j;
 		if ( alpha_b )
 			alpha_b += j;
+		return;
 	}
 #endif
 


src/modules/core/transition_composite.c
```diff
@@ -144,6 +144,7 @@ void composite_line_yuv( uint8_t *dest, uint8_t *src, int width, uint8_t *alpha_
 			alpha_a += j;
 		if ( alpha_b )
 			alpha_b += j;
+		return;
```
bmatherly, line 147: I think this will break for images with a width that is not a multiple of 8.

composite_line_yuv_sse2_simple can only operate on widths that are multiple of 8. The "for" loop below must run to process any remainder. For example, for a width of 10, composite_line_yuv_sse2_simple() will process 8 pixels and the for loop must run to process the last 2 pixels.

Can you test your proposed change with other widths that are not a multiple of 8?

bmatherly (changes requested):

j-b-m:
You are right sorry, my analysis was incorrect. The artifacts are in fact caused by another problem - when the compositing is performed on a non even width, the last pixel of the line is incorrectly reflecting the first one of the next line. This seems to be caused by[ this code:](https://github.com/mltframework/mlt/blob/master/src/modules/core/transition_composite.c#L381)

`	// Align chroma of source and destination
	if ( uneven_x != uneven_x_src )
	{
		p_src += 2;
		if ( alpha_b )
			alpha_b += 1;
	}`
Commenting the alpha_b stuff makes the artifacts disappear. Problem can be easily reproduced with this command, using a transparent png image with black on the left edge.

melt -profile atsc_1080p_25 color:red out=50 -track sample.png out=50 -transition composite out=50 geometry="0=-1920 0 1920 1080;50=0 0 1920 1080"

When the compositing width is uneven (due to the geometry param position), you can see a 1 pixel black line on the right edge of the compositing.

ddennedy:
I wonder why Kdenlive is still offering this service with animation; the quality of that is horrible. The [origin](https://github.com/mltframework/mlt/commit/1f47e5f4e04fb5c415bf5630c4e85a52b7b8b5da) of the line `alpha_b += 1` and its condition goes back 14 years.

I made a sample.png and reproduced it. However, I have to wonder if there is another use case where this is required. The commit says "when using crop panning," which has something to do with the undocumented "crop" property. This was related to an old (and dead) contract project I did where I needed to scroll text within the compositing rectangle in realtime. With that said, I am OK to remove it since the reported use case is much more common.

j-b-m:
Thanks Dan for your anser. Should have done this sooner, but I now switched to affine in Kdenlive, so this is not an issue for us anymore. Do you want me to push a change to remove the `alpha_b += 1;` ?

ddennedy:
>  Do you want me to push a change to remove the `alpha_b += 1;` ?

Yes, thanks

PR #737, opened 2021-10-04
Allow RELOCATABLE on Linux too

vpinon:
Hello!

When building in KDE Craft for Linux, MLT doesn't find its plugins as the build-time locations don't correspond to finale locations.
In Kdenlive AppImage we have a startup script, positioning environment variables to pick locations at runtime and make everything work.
To make simpler the debug and the distribution of a binary bundle, I find more convenient to use the win/mac approach to look for the plugins relatively to the app.
So I made the "RELOCATABLE" option useful on Linux too, OFF by default on this platform.
Now Kdenlive Craft build finds frei0r and ladspa plugins without needing environment variables.

Tested with option ON in Craft/CentOS7, OFF on plain CentOS7.


src/framework/mlt_factory.c
```diff
@@ -137,10 +144,14 @@ mlt_repository mlt_factory_init( const char *directory )
...
 		mlt_properties_set( global_properties, "MLT_APPDIR", appdir );
 		free( appdir );
-#elif defined(__APPLE__)  && defined(RELOCATABLE)
+#elif defined(RELOCATABLE)
 		char path[1024];
 		uint32_t size = sizeof( path );
+#  ifdef __APPLE__
 		_NSGetExecutablePath( path, &size );
+#  else
+                readlink( "/proc/self/exe", path, size );
```
ddennedy, line 153: Incorrect white space to indent - should be tabs.

CMakeLists.txt
```diff
@@ -49,9 +49,12 @@ option(SWIG_TCL "Enable SWIG Tcl bindings" OFF)
...
 if(WIN32)
   option(WINDOWS_DEPLOY "Install exes/libs directly to prefix (no subdir /bin)" ON)
-endif()
-if(APPLE)
-  option(RELOCATABLE "Use standard app bundle layout" ON)
+else()
+  if(APPLE)
+    option(RELOCATABLE "Use standard app bundle layout" ON)
+  else()
+    option(RELOCATABLE "Look for plugins relatively to app" OFF)
```
ddennedy, line 56: Change "relatively" (not valid word) to "relative".

ddennedy (changes requested):
I agree to the change, but I have only some very minor cleanups to request. Thanks

CMakeLists.txt
vpinon: Thanks for noticing this, corrected!

src/framework/mlt_factory.c
vpinon: right, fixed!

PR #744, opened 2021-10-28
Add audio cache to avformat producer

### LINKED ISSUE #738: avformat producer does not cache audio
While experimenting with running melt on HTTPS URLs I noticed single-stepping backwards would always result in a seek and a new HTTPS connection having to be set up, verified by watching apache2's access.log on my home server. When running melt with -hide-audio this issue goes away and single-stepping backward/forward works as expected.

Example commands:

    melt -producer avformat:https://samples.mplayerhq.hu/MXF/Quantel%20widescreen%20test.mxf

compared to

    melt -hide-audio -producer avformat:https://samples.mplayerhq.hu/MXF/Quantel%20widescreen%20test.mxf

This seems to happen with any container format, not just MXF. Looking at producer_avformat.c it indeed seems there's no caching for audio.

I may submit a patch for this, reporting it for now while I investigate some other stuff.

Tjoppen:
I based this off of how producer_avformat.c does video caching. Seems to work fine and doesn't leak any more memory than melt did before this patchset.

I purposefully *don't* use mlt_frame_get_audio() because it sets \*samples to zero.

This fixes #738

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -112,6 +112,7 @@ struct producer_avformat_s
 	unsigned int invalid_pts_counter;
 	unsigned int invalid_dts_counter;
 	mlt_cache image_cache;
+	mlt_cache audio_cache;
 	int yuv_colorspace, color_primaries, color_trc;
 	int full_luma;
 	pthread_mutex_t video_mutex;
@@ -2632,6 +2633,9 @@ static int producer_get_audio( mlt_frame frame, void **buffer, mlt_audio_format 
 {
 	// Get the producer
 	producer_avformat self = mlt_frame_pop_audio( frame );
+	mlt_producer producer = self->parent;
+	mlt_properties properties = MLT_PRODUCER_PROPERTIES( producer );
+	mlt_properties frame_properties = MLT_FRAME_PROPERTIES( frame );
 
 	pthread_mutex_lock( &self->audio_mutex );
 	
@@ -2706,6 +2710,25 @@ static int producer_get_audio( mlt_frame frame, void **buffer, mlt_audio_format 
 		AVPacket pkt;
 		mlt_channel_layout layout = mlt_channel_auto;
 
+
+		// Get the audio cache
+		if ( !self->audio_cache )
+		{
+			init_cache(properties, &self->audio_cache);
+		}
+		if ( self->audio_cache )
+		{
+			mlt_frame original = mlt_cache_get_frame( self->audio_cache, position );
+			if ( original )
+			{
+				mlt_frame_get_audio( original, buffer, format, frequency, channels, samples );
+				mlt_properties_set_data( frame_properties, "avformat.audio_cache", original, 0, (mlt_destructor) mlt_frame_close, NULL );
+				mlt_properties_pass_property( frame_properties, MLT_FRAME_PROPERTIES(original), "channel_layout" );
+				goto done_get_audio;
+			}
+		}
+		// Cache miss
+
 		av_init_packet( &pkt );
 		
 		// Caller requested number samples based on requested sample rate.
@@ -2750,8 +2773,6 @@ static int producer_get_audio( mlt_frame frame, void **buffer, mlt_audio_format 
 				}
 				else if ( ret < 0 )
 				{
-					mlt_producer producer = self->parent;
-					mlt_properties properties = MLT_PRODUCER_PROPERTIES( producer );
 					if ( ret != AVERROR_EOF )
 						mlt_log_verbose( MLT_PRODUCER_SERVICE(producer), "av_read_frame returned error %d inside get_audio\n", ret );
 					if ( !self->seekable && mlt_properties_get_int( properties, "reconnect" ) )
@@ -2868,6 +2889,12 @@ static int producer_get_audio( mlt_frame frame, void **buffer, mlt_audio_format 
 				memset( *buffer, silence, *samples * *channels * sizeof_sample );
 			}
 		}
+
+		mlt_properties_set_int( frame_properties, "audio_samples", *samples );
+
+		if ( self->audio_cache ) {
+			mlt_cache_put_frame( self->audio_cache, frame );
+		}
 	}
 	else
 	{
@@ -2875,6 +2902,7 @@ exit_get_audio:
 		// Get silence and don't touch the context
 		mlt_frame_get_audio( frame, buffer, format, frequency, channels, samples );
 	}
+done_get_audio:
 	
 	// Regardless of speed (other than paused), we expect to get the next frame
 	if ( !paused )
@@ -3136,6 +3164,7 @@ static void producer_avformat_close( producer_avformat self )
 
 	// Cleanup caches.
 	mlt_cache_close( self->image_cache );
+	mlt_cache_close( self->audio_cache );
 	if ( self->last_good_frame )
 		mlt_frame_close( self->last_good_frame );
 
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1599,6 +1599,31 @@ static int ignore_send_packet_result(int result)
 	return result >= 0 || result == AVERROR(EAGAIN) || result == AVERROR_EOF || result == AVERROR_INVALIDDATA || result == AVERROR(EINVAL);
 }
 
+static void init_cache(mlt_properties properties, mlt_cache *cache)
+{
+	// if cache size supplied by environment variable
+	int cache_supplied = getenv( "MLT_AVFORMAT_CACHE" ) != NULL;
+	int cache_size = cache_supplied? atoi( getenv( "MLT_AVFORMAT_CACHE" ) ) : 0;
+
+	// cache size supplied via property
+	if ( mlt_properties_get( properties, "cache" ) )
+	{
+		cache_supplied = 1;
+		cache_size = mlt_properties_get_int( properties, "cache" );
+	}
+	if ( mlt_properties_get_int( properties, "noimagecache" ) )
+	{
+		cache_supplied = 1;
+		cache_size = 0;
+	}
+	// create cache if not disabled
+	if ( !cache_supplied || cache_size > 0 )
+		*cache = mlt_cache_init();
+	// set cache size if supplied
+	if ( *cache && cache_supplied )
+		mlt_cache_set_size( *cache, cache_size );
+}
+
 /** Get an image from a frame.
 */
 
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -2701,6 +2710,25 @@ static int producer_get_audio( mlt_frame frame, void **buffer, mlt_audio_format
...
+		if ( !self->audio_cache )
+		{
+			init_cache(properties, &self->audio_cache);
+		}
+		if ( self->audio_cache )
+		{
+			mlt_frame original = mlt_cache_get_frame( self->audio_cache, position );
+			if ( original )
+			{
+				mlt_frame_get_audio( original, buffer, format, frequency, channels, samples );
```
bmatherly, line 2724: I notice that this directly passes the buffer from the cache frame (rather than make a copy). If a filter is applied to the frame in an attached filter, will the cached buffer be modified by the filter? I am not sure if that issue is mitigated in some non-obvious way.

If it would be better to copy the data, I recommend to use mlt_audio_copy()
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_audio.h#L57

src/modules/avformat/producer_avformat.c
```diff
@@ -112,6 +112,7 @@ struct producer_avformat_s
 	unsigned int invalid_pts_counter;
 	unsigned int invalid_dts_counter;
 	mlt_cache image_cache;
+	mlt_cache audio_cache;
```
bmatherly, line 115: In this design, there are two caches - one that holds mlt_frame instances with only images - and another that holds mlt_frame instances with only audio.

Two comments:
1) Could it be possible to have only one cache that holds mlt_frame instances with both image and audio in the same frame?
If two caches is the best approach...
2) Have you considered making the audio cache hold mlt_audio instances instead?
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_audio.h

src/modules/avformat/producer_avformat.c
Tjoppen: I don't really know. Video frames are copied, so copying the audio also is likely the correct approach.

src/modules/avformat/producer_avformat.c
Tjoppen: >Could it be possible to have only one cache that holds mlt_frame instances with both image and audio in the same frame?

This would be best. I am not too familiar with mlt yet, which is why reviews like this are good :)

>Have you considered making the audio cache hold mlt_audio instances instead?

I have not. Guess I'll have to take a look at it.

src/modules/avformat/producer_avformat.c
bmatherly: I do not want to create extra work, but I think it would be good to consider an implementation that uses mlt_audio in the cache, and copies the data  into the frame.

There is an example here that shows how to move data from mlt_audio to an mlt_frame instance:
https://github.com/mltframework/mlt/blob/master/src/modules/core/link_timeremap.c#L194
Your example reminded me that mlt_channel_layout is not being passed in that code - I should probably fix that.

I think the general process would be:
1) If cache miss: get audio in the normal way, then use mlt_audio_new() to create an audio instance, copy the audio from the frame to the instance and push into the cache.
2) If cache hit: get the mlt_audio instance from the cache and copy into the frame to be output.

ddennedy: You do not need to make a copy. `mlt_cache_put_frame()`, which is what this is using, clones the frame with deep copy. That is documented. He is following the pattern done for image cloning, which is a good thing.

src/modules/avformat/producer_avformat.c
ddennedy: > Could it be possible to have only one cache that holds mlt_frame instances with both image and audio in the same frame?

That could make the code unnecessarily more difficult than it already is. Before `mlt_cache_put_frame()`, you try to get the frame from cache. If a hit, update the frame's data, and you still need to `mlt_cache_put_frame()` because `mlt_cache_get_frame()` also clones the frame.

> Have you considered making the audio cache hold mlt_audio instances instead?

Part of what makes `mlt_cache_put_frame()` and `mlt_cache_get_frame()` special is that they understand position (frame number), and this data is position-sensitive. To use `mlt_audio` you would have to have put the position into the name of the cache item. Something like "audio:%d".

bmatherly: Thanks for the lesson, Dan. Sounds like this proposal is already on the best track.

ddennedy: > To use mlt_audio you would have to have put the position into the name of the cache item. Something like "audio:%d".

And actually on second review, that will not work either. A name is only used with `mlt_service_cache_put()` and `mlt_service_cache_get()` and even then you cannot store position-sensitive data unless the data itself has position information internally. For example, there is a "producer_avformat" that is cached like this to only retain the producer's private data for the N most recently used producers (to avoid exhausting resources like file handles and memory). If you put a position into this name, then there will be as many `mlt_cache_s` as there are frames!

Tjoppen: So what should I do? The code works as it is now, but I don't know how it interacts with filters and such.

ddennedy: The complication I was thinking about to reduce redundant data:

- change the private `image_cache` to `frame_cache`
- any time you are you going to `mlt_cache_put_frame()` you need to first `mlt_cache_get_frame()`
- if you get a cache hit, copy the audio or image data as well as relevant properties from the cache frame to the current frame before calling `mlt_cache_put_frame()` on the current frame. In `producer_get_image()` you would need to copy the audio, and in `producer_get_audio()` you need to copy the image.

Unfortunately, while this reduces redundancy, it increases the amount of copying, either of which are often pure overhead. This will add 3 new copy operations when there are cache hits, which is often as typically both the image and the audio are requested for a frame: in `mlt_cache_get_frame()`, from the cache frame to the current frame, and finally in `mlt_cache_put_frame()`. And whenever there is a frame copied, it may be copying image, alpha, and audio data if they have been set. So, up to 9 `memcpy`s of media data plus all the properties that get copied inside each `mlt_frame_clone()` used by the cache.

I need to seriously consider the impact to Shotcut of either increase since many of these users run on modest computers. The bug you reported is a low priority for me. How often does one need to single step frames over HTTP? And this change only covers stepping by the number of frames configured for caching, which by default is 4. I am curious about what is your use case other than a casual observation while tinkering?

Here is another idea: keep this separate audio cache, but prior to putting it into the cache clear the image on the frame. Likewise, when putting a frame into the image cache, clear the audio on the frame. This reduces both the redundancy and the number of copy operations. Unfortunately, you cannot clear the audio or the image(+alpha) of the current frame or it will seriously break things! We would need a way to tell the `mlt_cache_put_frame()` and `mlt_frame_clone()` functions whether to copy audio, image(+alpha), or all.

bmatherly: > Unfortunately, while this reduces redundancy, it increases the amount of copying, either of which are often pure overhead.

I agree that we should work to reduce the number of copies.

Another idea (in addition to the one you suggested) would be to add two new functions:
* mlt_cache_peek_frame()
* mlt_cache_update_frame()
These functions would allow the cache to be inspected and updated without the additional copies. It would also need to be possible to lock the cache during peeking and updating.

The sequence would be:
1. Lock the cache
2. Peek in the cache to check for a frame
3. If a frame is returned, check for audio.
4. Unlock the cache
5. If frame and audio, get the copied frame from the cache using get_frame()

Tjoppen: > Unfortunately, while this reduces redundancy, it increases the amount of copying, either of which are often pure overhead. This will add 3 new copy operations when there are cache hits, which is often as typically both the image and the audio are requested for a frame: in `mlt_cache_get_frame()`, from the cache frame to the current frame, and finally in `mlt_cache_put_frame()`. And whenever there is a frame copied, it may be copying image, alpha, and audio data if they have been set. So, up to 9 `memcpy`s of media data plus all the properties that get copied inside each `mlt_frame_clone()` used by the cache.

These are very modest amounts of data compared to video frames. Less than 8k for a typical 48 kHz stereo 16-bit track at 25 fps. I doubt that is much of an issue.

> I need to seriously consider the impact to Shotcut of either increase since many of these users run on modest computers. The bug you reported is a low priority for me. How often does one need to single step frames over HTTP? And this change only covers stepping by the number of frames configured for caching, which by default is 4. I am curious about what is your use case other than a casual observation while tinkering?

A client wants to be able to edit video directly from say an S3 bucket. One alternative could be to mount the URLs using a curl-based fuse thing instead, with a very large cache. That way the massive delays currently seen when single-stepping HTTP URLs goes away for the most part. These delays are on the order of half a second, since a new TLS connection has to be negotiated for every reverse step.

A related issue is large delays when opening HTTP URLs. It can take up to half a minute to parse an MXF file served over HTTP. But that is an issue with libavformat itself. I should add that I maintain the MXF code in lavf, so I have intimate knowledge about seeking behavior of many of its demuxers.

ddennedy: > These are very modest amounts of data compared to video frames.

If it is just about the audio data being redundant I am not so concerned. I am concerned about the image data:

- In MLT frames typically carry both audio and image.
- The consumer of the avformat module requests the image with frame N.
- The avformat producer stores the image in the frame and caches the frame to its image_cache.
- The consumer requests the audio with same frame N.
- The avformat stores the audio in the frame and caches the frame to its audio_cache.

Now, there are two frames for the same position in separate caches, each with a copy of the image for this frame.

Possibly, I am being a bit paranoid but only because of bad experiences. :-) In any case, upon reviewing the threading models in mlt_consumer.c and their calls to `mlt_frame_get_audio()` and `mlt_frame_get_image()`, I see that audio is always requested before video. However, it is possible that something is not using `mlt_consumer`.

There are another couple of simpler ideas:

1. Have only one `mlt_cache` in the avformat producer's `producer_avformat_s` struct. But there will still be a little more copying.
2. Leave it the basically the way you have it currently in the PR, but add a property to set the amount of audio caching. Leave it unset by default in which case it evaluates to 0 when getting it. Then, you can set it to a higher value for your needs.

In case I am wrong about the order of the calls to get_audio and get_image are wrong in certain scenarios (e.g. unthreaded when consumer property `real_time` = 0), do not underestimate how quickly things add up in complex projects. The total number of frames cached depends on the number of avformat producers cached. The default number of those cached is not simply the mlt_cache default of 4 (or some other low value set). Typically, it depends more on the number of tracks * 2 (see `resize_service_caches()` in `mlt_multrack.c`). Why? To prevent thrashing. Thrashing is when producers are flushed and restored upon every iteration of frames for a period. That can happen when many tracks are being mixed/composited/blended. Why 2x? Because there can be a single track transition that involves 2 producers. In theory, there can be one of these on every track. That sounds paranoid? The consequence is crash or artifacts discovered through support and bug reports.

ddennedy:
I was reviewing some things today, and I agree to merge this after the next release if you simply "add a property to set the amount of audio caching. Leave it unset by default in which case it evaluates to 0 when getting it. Then, you can set it to a higher value for your needs."

Tjoppen:
> I was reviewing some things today, and I agree to merge this if you simply "add a property to set the amount of audio caching. Leave it unset by default in which case it evaluates to 0 when getting it. Then, you can set it to a higher value for your needs."

Worksforme. I lost some interest in this patch because I wrote a URL caching FUSE thing instead. While audio caching helps, it does not help with mlt opening the same file/URL several times instead of just once. This means libavformat has to re-parse the file. This is especially bad with MXF files. I forget the exact number of times mlt re-opens the file, but it's something like seven. So if it takes ffprobe five seconds to probe a HTTP URL then it takes melt around thirty seconds..

bmatherly:
This PR was implemented by https://github.com/mltframework/mlt/pull/903

PR #746, opened 2021-10-31
Add property animation support to avfilter

richardpl:
Looks to work fine in my limited testing. Only video filters supported for now.

BTW, why are av.planes (which is of type flags) are not listed at runtime at all?

### DIFF
--- a/src/modules/avformat/filter_avfilter.c
+++ b/src/modules/avformat/filter_avfilter.c
@@ -770,6 +770,28 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format 
 			}
 		}
 
+		{
+			mlt_service service = MLT_FILTER_SERVICE(filter);
+			mlt_properties prop = MLT_SERVICE_PROPERTIES(service);
+			mlt_position position = mlt_filter_get_position( filter, frame );
+			int length = mlt_filter_get_length2( filter, frame );
+			int count = mlt_properties_count(prop);
+			int i;
+			for (i = 0; i < count; i++)
+			{
+				char *name = mlt_properties_get_name(prop, i);
+				if (!strncmp(name, PARAM_PREFIX, PARAM_PREFIX_LEN))
+				{
+					char *new_val = mlt_properties_anim_get( prop, name, position, length);
+					char *cur_val = NULL;
+					av_opt_get(pdata->avfilter_ctx->priv, name + PARAM_PREFIX_LEN, AV_OPT_SEARCH_CHILDREN, (uint8_t **)&cur_val);
+					if (new_val && cur_val && strcmp(new_val, cur_val))
+						avfilter_graph_send_command(pdata->avfilter_graph, pdata->avfilter->name, name + PARAM_PREFIX_LEN, new_val, NULL, 0, 0);
+					av_free(cur_val);
+				}
+			}
+		}
+
 		// Run the frame through the filter graph
 		ret = av_buffersrc_add_frame( pdata->avbuffsrc_ctx, pdata->avinframe );
 		if( ret < 0 ) {


ddennedy:
Thank you, Paul. I will merge this after I make a new release today to go with the new version of Shotcut. For a long time I thought sendcmd was only a filter and recently discovered this API. Glad to have your assistance.

richardpl:
Also is there any way to check if certain options are not animated - the option value haven't changed, because this may cause slowdown for some filters, where options changing come at some non-trivial runtime cost.

ddennedy:
> why are av.planes (which is of type flags) are not listed at runtime at all?

Can you give the name of a filter? Some filters have planes as an integer and appear:
https://mltframework.org/plugins/FilterAvfilter-kirsch/

> is there any way to check if certain options are not animated - the option value haven't changed

To determine if animated, you could check if the string representation of the property value (entire value using `mlt_properties_get_value()`) contains both a ';' and '=' as a heuristic - except when the value begins and ends with a double-quote. More about the double-quote rules for animated string properties is in fa6df2ded.

To determine if the value has not changed is not directly possible in MLT because a filter does not know the previous frame number that was used unless you save that in your filter's state. With that info, you could get the property value at the previous frame value. Alternatively, I think you might be able to compare the current MLT property value with the corresponding AVOption value?

richardpl:
> > why are av.planes (which is of type flags) are not listed at runtime at all?
> 
> Can you give the name of a filter? Some filters have planes as an integer and appear: https://mltframework.org/plugins/FilterAvfilter-kirsch/

Yes, but not with property_get_animated

> 
> > is there any way to check if certain options are not animated - the option value haven't changed
> 
> To determine if animated, you could check if the string representation of the property value (entire value using `mlt_properties_get_value()`) contains both a ';' and '=' as a heuristic - except when the value begins and ends with a double-quote. More about the double-quote rules for animated string properties is in [fa6df2d](https://github.com/mltframework/mlt/commit/fa6df2dedabe36bfe949d757ebde2d9c97bba62e).
> 
> To determine if the value has not changed is not directly possible in MLT because a filter does not know the previous frame number that was used unless you save that in your filter's state. With that info, you could get the property value at the previous frame value. Alternatively, I think you might be able to compare the current MLT property value with the corresponding AVOption value?

Will check, thanks.

richardpl:
Here it checks if value changed and updates values if it does.

ddennedy:
> Yes, but not with property_get_animated

In MLT, a property does not exist until it is set. If you try to get something unset you get 0 for a numeric getter or NULL for string or data getter. A service can publish a parameter in its metadata that it has not actually initialized because 0, NULL, or unset is its default. Otherwise, it will set the parameter default as a property in its constructor with its default value. In the case of avfilter, the MLT filter is a bridge to libavfilter and understands that the libavfilter has an AVOption backing it with its default value.

ddennedy:
There is a problem:
`melt -profile atsc_720p_25 color:red -attach avfilter.hue 'av.h=0=0;50=90;100=225;-1=0' out=500`

should quickly go from red to green to blue and slowly return to red. However, instead I got:

> [hue @ 0x7f20dc00aa80] [Eval @ 0x7f20f5de7800] Invalid chars '0;50=90;100=225;-1=0' at the end of expression '0=0;50=90;100=180'
> [hue @ 0x7f20dc00aa80] Error when evaluating the expression '0;50=90;100=225;-1=0' for h

That is because the calls to `set_avfilter_options()` when creating the filtergraph are passing the entire MLT property string. Here is what I changed to fix it:
```
--- src/modules/avformat/filter_avfilter.c
+++ src/modules/avformat/filter_avfilter.c
@@ -117,29 +117,31 @@ static mlt_image_format get_supported_image_format( mlt_image_format format )
 		return mlt_image_yuv422;
 	}
 }
 
-static void set_avfilter_options( mlt_filter filter, double scale)
+static void set_avfilter_options( mlt_filter filter, double scale, mlt_frame frame )
 {
 	private_data* pdata = (private_data*)filter->child;
 	mlt_properties filter_properties = MLT_FILTER_PROPERTIES(filter);
 	int i;
 	int count = mlt_properties_count( filter_properties );
 	mlt_properties scale_map = mlt_properties_get_data(filter_properties, "_resolution_scale", NULL);
+	mlt_position position = mlt_filter_get_position( filter, frame );
+	int length = mlt_filter_get_length2( filter, frame );
 
 	for( i = 0; i < count; i++ )
 	{
 		const char *param_name = mlt_properties_get_name( filter_properties, i );
 		if( param_name && strncmp( PARAM_PREFIX, param_name, PARAM_PREFIX_LEN ) == 0 )
 		{
 			const AVOption *opt = av_opt_find( pdata->avfilter_ctx->priv, param_name + PARAM_PREFIX_LEN, 0, 0, 0 );
-			const char* value = mlt_properties_get_value( filter_properties, i );
+			const char* value = mlt_properties_anim_get( filter_properties, param_name, position, length );
 			if( opt && value )
 			{
 				if (scale != 1.0) {
 					double scale2 = mlt_properties_get_double(scale_map, opt->name);
 					if (scale2 != 0.0) {
-						double x = mlt_properties_get_double(filter_properties, param_name);
+						double x = mlt_properties_anim_get_double(filter_properties, param_name, position, length);
 						x *= scale * scale2;
 						mlt_properties_set_double(filter_properties, "_avfilter_temp", x);
 						value = mlt_properties_get(filter_properties, "_avfilter_temp");
 					}
@@ -149,9 +151,9 @@ static void set_avfilter_options( mlt_filter filter, double scale)
 		}
 	}
 }
 
-static void init_audio_filtergraph( mlt_filter filter, mlt_audio_format format, int frequency, int channels )
+static void init_audio_filtergraph( mlt_filter filter, mlt_audio_format format, int frequency, int channels, mlt_frame frame )
 {
 	private_data* pdata = (private_data*)filter->child;
 	const AVFilter *abuffersrc  = avfilter_get_by_name("abuffer");
 	const AVFilter *abuffersink = avfilter_get_by_name("abuffersink");
@@ -256,9 +258,9 @@ static void init_audio_filtergraph( mlt_filter filter, mlt_audio_format format,
 	if( !pdata->avfilter_ctx ) {
 		mlt_log_error( filter, "Cannot create audio filter\n" );
 		goto fail;
 	}
-	set_avfilter_options( filter, 1.0 );
+	set_avfilter_options( filter, 1.0, frame );
 	ret = avfilter_init_str(  pdata->avfilter_ctx, NULL );
 	if( ret < 0 ) {
 		mlt_log_error( filter, "Cannot init filter\n" );
 		goto fail;
@@ -289,9 +291,9 @@ fail:
 	avfilter_graph_free( &pdata->avfilter_graph );
 }
 
 
-static void init_image_filtergraph( mlt_filter filter, mlt_image_format format, int width, int height, double resolution_scale )
+static void init_image_filtergraph( mlt_filter filter, mlt_image_format format, int width, int height, double resolution_scale, mlt_frame frame )
 {
 	private_data* pdata = (private_data*)filter->child;
 	mlt_profile profile = mlt_service_profile(MLT_FILTER_SERVICE(filter));
 	const AVFilter *buffersrc  = avfilter_get_by_name("buffer");
@@ -393,9 +395,9 @@ static void init_image_filtergraph( mlt_filter filter, mlt_image_format format,
 	if( !pdata->avfilter_ctx ) {
 		mlt_log_error( filter, "Cannot create video filter\n" );
 		goto fail;
 	}
-	set_avfilter_options( filter, resolution_scale );
+	set_avfilter_options( filter, resolution_scale, frame );
 
 	if ( !strcmp( "lut3d", pdata->avfilter->name ) ) {
 #if defined(__GLIBC__) || defined(__APPLE__) || (__FreeBSD__)
 		// LUT data files use period for the decimal point regardless of LC_NUMERIC.
@@ -588,9 +590,9 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 	mlt_service_lock( MLT_FILTER_SERVICE( filter ) );
 
 	if( pdata->reset || pdata->format != *format )
 	{
-		init_audio_filtergraph( filter, *format, *frequency, *channels );
+		init_audio_filtergraph( filter, *format, *frequency, *channels, frame );
 		pdata->reset = 0;
 	}
 
 	if( pdata->avfilter_graph )
@@ -694,9 +696,9 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
 
 	if( pdata->reset || pdata->format != *format || pdata->width != *width || pdata->height != *height )
 	{
 		double scale = mlt_profile_scale_width(profile, *width);
-		init_image_filtergraph( filter, *format, *width, *height, scale );
+		init_image_filtergraph( filter, *format, *width, *height, scale, frame );
 		pdata->reset = 0;
 	}
 
 	if( pdata->avfilter_graph )
```

ddennedy:
I ran a quick benchmark on the changes thus far (with keyframes, of course), and with
`time melt -profile uhd_2160p_60 color:red out=5000 length=5000 -attach avfilter.hue av.h=90 -consumer null terminate_on_pause=1 real_time=-1`

Before changes I get
>real	0m55.694s
user	0m55.810s
sys	0m0.193s

After I get
> real	0m55.375s
user	0m55.526s
sys	0m0.173s

I have no explanation for the small speed increase, but I'll take it. :-)

ddennedy:
One last thing, and then I think we can merge: refactor your initial change into a function that is called in both `filter_get_image()` and `filter_get_audio()`

ddennedy:
Or I can take it from here if you prefer.

richardpl:
Yes, please.

ddennedy:
closing because superseded by #747

PR #747, opened 2021-11-04
Send avfilter commands

ddennedy:
@bmatherly Please check this over for me. This is based on #746, but once I started testing with Shotcut in a [modifed Hue/Lightness/Saturation filter](https://github.com/mltframework/shotcut/pull/1176), I had to change some things (some details in comments of 75e07e7).

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -624,3 +624,9 @@ MLT_7.1.0 {
     mlt_properties_get_properties;
     mlt_properties_get_properties_at;
 } MLT_7.0.0;
+
+MLT_7.4.0 {
+  global:
+    mlt_property_is_anim;
+    mlt_properties_is_anim;
+} MLT_7.1.0;
--- a/src/modules/avformat/factory.c
+++ b/src/modules/avformat/factory.c
@@ -229,12 +229,6 @@ static void add_parameters( mlt_properties params, void *object, int req_flags, 
 			mlt_properties_set( p, "type", "string" );
 			break;
 		}
-		mlt_properties_set( p, "mutable", "yes" );
-#if LIBAVUTIL_VERSION_INT >= ((56<<16)+(35<<8)+101)
-		if ((opt->flags & AV_OPT_FLAG_RUNTIME_PARAM)) {
-			mlt_properties_set( p, "animation", "yes" );
-		}
-#endif
 		// If the option belongs to a group (unit) and is not a constant (keyword value)
 		if ( opt->unit && opt->type != AV_OPT_TYPE_CONST )
 		{
@@ -375,7 +369,6 @@ static mlt_properties avfilter_metadata( mlt_service_type type, const char *id, 
 			mlt_properties_set( p, "description", "The MLT position value to set on avfilter frames" );
 			mlt_properties_set( p, "type", "string" );
 			mlt_properties_set( p, "default", "frame" );
-			mlt_properties_set( p, "mutable", "yes" );
 			mlt_properties values = mlt_properties_new();
 			mlt_properties_set_data( p, "values", values, 0, (mlt_destructor) mlt_properties_close, NULL );
 			snprintf( key, 20, "%d", i++ );
--- a/src/modules/avformat/factory.c
+++ b/src/modules/avformat/factory.c
@@ -229,6 +229,12 @@ static void add_parameters( mlt_properties params, void *object, int req_flags, 
 			mlt_properties_set( p, "type", "string" );
 			break;
 		}
+		mlt_properties_set( p, "mutable", "yes" );
+#if LIBAVUTIL_VERSION_INT >= ((56<<16)+(35<<8)+101)
+		if ((opt->flags & AV_OPT_FLAG_RUNTIME_PARAM)) {
+			mlt_properties_set( p, "animation", "yes" );
+		}
+#endif
 		// If the option belongs to a group (unit) and is not a constant (keyword value)
 		if ( opt->unit && opt->type != AV_OPT_TYPE_CONST )
 		{
@@ -369,6 +375,7 @@ static mlt_properties avfilter_metadata( mlt_service_type type, const char *id, 
 			mlt_properties_set( p, "description", "The MLT position value to set on avfilter frames" );
 			mlt_properties_set( p, "type", "string" );
 			mlt_properties_set( p, "default", "frame" );
+			mlt_properties_set( p, "mutable", "yes" );
 			mlt_properties values = mlt_properties_new();
 			mlt_properties_set_data( p, "values", values, 0, (mlt_destructor) mlt_properties_close, NULL );
 			snprintf( key, 20, "%d", i++ );
--- a/src/modules/avformat/filter_avfilter.c
+++ b/src/modules/avformat/filter_avfilter.c
@@ -770,6 +770,28 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format 
 			}
 		}
 
+		{
+			mlt_service service = MLT_FILTER_SERVICE(filter);
+			mlt_properties prop = MLT_SERVICE_PROPERTIES(service);
+			mlt_position position = mlt_filter_get_position( filter, frame );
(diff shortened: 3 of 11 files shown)


src/modules/avformat/filter_avfilter.c
```diff
@@ -58,6 +58,11 @@ typedef struct
 	int reset;
 } private_data;
 
+static inline int non_animated(const char *value)
```
bmatherly, line 61: I wonder if we should consider adding this to mlt_properties. If we do that, would it be possible to check the property for an animation object instead of parsing a string?

src/modules/avformat/filter_avfilter.c
```diff
@@ -150,6 +156,35 @@ static void set_avfilter_options( mlt_filter filter, double scale)
...
+	int count = mlt_properties_count(prop);
+	int i;
+	for (i = 0; i < count; i++)
+	{
+		char *name = mlt_properties_get_name(prop, i);
+		if (!strncmp(name, PARAM_PREFIX, PARAM_PREFIX_LEN))
+		{
+			const AVOption *opt = av_opt_find( pdata->avfilter_ctx->priv, name + PARAM_PREFIX_LEN, 0, 0, 0 );
+			if (opt && opt->flags & AV_OPT_FLAG_RUNTIME_PARAM) {
+				mlt_properties_set_double(prop, "_avfilter_temp", mlt_properties_anim_get_double(prop, name, position, length));
```
bmatherly, line 174: Should we be applying resolution scaling to the value before sending the command here?

src/modules/avformat/factory.c
```diff
@@ -229,6 +229,12 @@ static void add_parameters( mlt_properties params, void *object, int req_flags,
 			mlt_properties_set( p, "type", "string" );
 			break;
 		}
+		mlt_properties_set( p, "mutable", "yes" );
+#if LIBAVUTIL_VERSION_INT >= ((56<<16)+(35<<8)+101)
+		if ((opt->flags & AV_OPT_FLAG_RUNTIME_PARAM)) {
```
bmatherly, line 234: According to this implementation, all animated properties must be numbers. But I do not think this is always true for AV commands. Some AV commands are not represented as numbers. For example, see "width_type" for the equalizer filters. It is a string parameter. 

Maybe there is nothing we can do about that since I do not see a way to query the FFMpeg API to find out if a parameter is a number or string. Maybe we can maintain a blacklist. Or maybe users just need to know that they can not actually animate string parameters.

src/modules/avformat/filter_avfilter.c
ddennedy: When the life a property begins as a string such as from XML or `melt`, there is no animation object until the property has been accessed using an anim getter, which causes it to be parse into `mlt_animation`. However, yes, we can consider moving this into a common property.

src/modules/avformat/filter_avfilter.c
ddennedy: yes, good find

src/modules/avformat/factory.c
ddennedy: [AVOptions](https://ffmpeg.org/doxygen/trunk/structAVOption.html) do have a type. It is true this limits it to numeric only, but like I said a lot of the AVOptions for a filter have the type as string to support expressions even though the expression must evaluate to numeric to be usable. So, I think this might need to be a sacrifice for the majority. I do not want to maintain our own list of types for each filter's options because it is version-sensitive and has been growing a lot. We need to think less about users and more about developers. I have been saying for a while now that MLT is for developers. Thus, the developers need to be careful about how it is used and avoid using keyframes for actual string parameters. In case a developer tries to expose everything generically, it will be a problem they may need to workaround somehow.

src/modules/avformat/filter_avfilter.c
```diff
@@ -70,7 +75,8 @@ static void property_changed( mlt_service owner, mlt_filter filter, mlt_event_da
 			{
 				if( !strcmp( opt->name, name + PARAM_PREFIX_LEN ) )
 				{
-					pdata->reset = 1;
+					const char* value = mlt_properties_get(MLT_FILTER_PROPERTIES(filter), name);
+					pdata->reset = non_animated(value);
```
ddennedy, line 79: I think there is a good opportunity here to skip the reset when the AVOption has the `AV_OPT_FLAG_RUNTIME_PARAM` flag, and to skip setting it below in `set_avfilter_options()` under the sands condition because it will be set in `send_avformat_commands()`.

src/modules/avformat/factory.c
bmatherly: > the developers need to be careful about how it is used and avoid using keyframes for actual string parameters

If I understand the implementation correctly, even non-animated string properties will be converted to a double. I expect that "real" string properties that are not numbers will not survive that conversion. So I think it will be impossible to set the "width_type" parameter (whether animated or not).

Maybe we can accept that limitation for now and state it in the documentation. Or maybe we could add a new is_number() function to mlt_properties and use some heuristic to detect that a string is a number.

src/modules/avformat/filter_avfilter.c
bmatherly: I agree. That would be a good optimization.

src/modules/avformat/factory.c
bmatherly: > If I understand the implementation correctly, even non-animated string properties will be converted to a double.

Nevermind. I see it now. Only converted to double if animated. I think this is our best tradeoff - unless we want to implement number detection.

src/modules/avformat/filter_avfilter.c
ddennedy: I noticed a number of functions in `mlt_property.c` are applying a similar heuristic. Thus, it totally makes sense to refactor this into a common function, and I will.

src/modules/avformat/factory.c
ddennedy: In hindsight, per my comment above about developers and generic UIs, I think now we should not set `animation: yes` (a new attribute in schema 7.0) in any of the filter metadata to prevent that. Basically, a user or developer will have to do their own analysis and testing to deliberately opt into using an animated property value with an avfilter.

bmatherly: I agree with that. Maybe just add a note in the metadata header as a breadcrumb for future developers (or ourselves).

src/modules/avformat/filter_avfilter.c
ddennedy: done

src/modules/avformat/filter_avfilter.c
ddennedy: done

ddennedy: done

src/modules/avformat/filter_avfilter.c
ddennedy: done

ddennedy: done

src/framework/mlt.vers
```diff
@@ -624,3 +624,9 @@ MLT_7.1.0 {
     mlt_properties_get_properties;
     mlt_properties_get_properties_at;
 } MLT_7.0.0;
+
+MLT_7.4.0 {
```
bmatherly, line 628: Why 7.4 instead of 7.2?

bmatherly (approved):
LGTM. No other comments except for the version number question.

src/framework/mlt.vers
ddennedy: 7.1 was wrong. The symbol version is supposed to match the release version. That is not required, but that is this project's convention ever since 6.0.

ddennedy: Or maybe you missed the fact that there was already a 7.2 release.

bmatherly: Oh Yes, that explains it. And you also skip odd numbers for some reason.

ddennedy: Odd version numbers is how we see whether someone is using a release or some interim version and need to get a git sha.

PR #756, opened 2021-12-06
Audio level visualization

bmatherly:
I think this is low risk since it mostly add a new service and only makes minor changes to existing code. But maybe it should wait until the next release.

### DIFF
--- a/src/modules/qt/filter_audiolevelgraph.yml
+++ b/src/modules/qt/filter_audiolevelgraph.yml
@@ -12,6 +12,18 @@ description: >
   An audio visualization filter that draws an audio level meter on the image.
 
 parameters:
+  - identifier: type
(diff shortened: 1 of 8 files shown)


src/modules/qt/filter_audiolevelgraph.yml
```diff
@@ -0,0 +1,157 @@
...
+    readonly: no
+    default: 0
+    minimum: 0
+    maximum: 360
+    mutable: yes
+    widget: spinner
+
+  - identifier: rect
+    title: Rectangle
+    description: >
```
ddennedy, line 95: Are you aware this text block will not keep the extra newlines you have added?
https://yaml-multiline.info/

ddennedy (approved):

PR #757, opened 2021-12-13
Ensure animation serialization includes last key value

bmatherly:
This is a proposed fix. I am open to suggestions about different ways to fix this. But I think it makes sense to force keyframe for the "out" value as is already done for the "in" value.
---------------------------------------------------------------------------

When serializing key frames, the last keyframe would not have a
value after the "=". Upon re-parsing the string, the last value
is interpred to be a "zero".

This fix forces the last key value to be included by marking it
as a key frame.

Fix for issue reported here:
https://forum.shotcut.org/t/keyframes-error-after-cut-of-parts/31180

### DIFF
--- a/src/framework/mlt_animation.c
+++ b/src/framework/mlt_animation.c
@@ -674,8 +674,11 @@ char *mlt_animation_serialize_cut_tf( mlt_animation self, int in, int out, mlt_t
 					break;
 
 				// Special case - crop at the out point
-				if ( item.frame > out )
+				if ( item.frame > out ) {
 					mlt_animation_get_item( self, &item, out );
+					// To ensure correct seeding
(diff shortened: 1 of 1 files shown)


ddennedy (approved):
Yes, this looks good to me, but I am curious about why you chose words of caution. Is it because of the results you got in #758 ?

bmatherly:
The caution is for two reasons:
1) I am unsure if I am misunderstanding the original design intent
2) The code as written 9 years ago - how did we not see this issue sooner? Did we change something else that uses it in an unexpected way?

If this change matches your original design intent, then my confidence in the change is high.

ddennedy:
It matches my design intent as well as I can recall after 9 years. :-)

PR #758, opened 2021-12-13
Add error checking to next_key() and previous_key()

bmatherly:
Some of these results are questionable. See comments.

### DIFF
--- a/src/mlt++/MltAnimation.cpp
+++ b/src/mlt++/MltAnimation.cpp
@@ -1,6 +1,6 @@
 /**
  * MltAnimation.cpp - MLT Wrapper
- * Copyright (C) 2015-2018 Meltytech, LLC
+ * Copyright (C) 2015-2021 Meltytech, LLC
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -101,6 +101,15 @@ mlt_keyframe_type Animation::keyframe_type( int position )
 		return (mlt_keyframe_type) -1;
 }
 
+/** Get the keyfame at the position or the next following.
+ *
+ * If no keyframe exists at or after the position, the return value is invalid
+ *
+ * \deprecated Prefer bool Animation::next_key( int position, int& key )
+ * \param position the frame number at which to start looking for the next keyframe
+ * \return the position of the next keyframe
+ */
+
 int Animation::next_key( int position )
 {
 	struct mlt_animation_item_s item;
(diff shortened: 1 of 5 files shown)


src/tests/test_animation/test_animation.cpp
```diff
@@ -383,6 +383,40 @@ private Q_SLOTS:
...
+		p.anim_get_int("foo", 0);
+		Animation a = p.get_animation("foo");
+		QVERIFY(a.is_valid());
+
+		QCOMPARE(a.next_key(0), 50);
+		QCOMPARE(a.next_key(59), 60);
+		QCOMPARE(a.next_key(60), 60);
+		QCOMPARE(a.next_key(61), 100);
+		QCOMPARE(a.next_key(100), 100);
+		QCOMPARE(a.next_key(101), 1);
```
bmatherly, line 401: This seems like a strange return value for this. It is outside of the range, but maybe it should return an error?

src/tests/test_animation/test_animation.cpp
```diff
@@ -383,6 +383,40 @@ private Q_SLOTS:
...
+	void PreviousKey()
+	{
+		Properties p;
+		p.set("foo", "50=100; 60=60; 100=0");
+		// Cause the string to be interpreted as animated value.
+		p.anim_get_int("foo", 0);
+		Animation a = p.get_animation("foo");
+		QVERIFY(a.is_valid());
+
+		QCOMPARE(a.previous_key(0), 50);
```
bmatherly, line 413: This seems like a strange return value for this. It is outside of the range, but maybe it should return an error instead of the "next" key?

src/tests/test_animation/test_animation.cpp
ddennedy: Yeah, one thing we need to be careful about when adding a unit test is that we are basically agreeing to the tested behavior.
The docs for `mlt_animation_next_key()` says
>  * \return true if there was an error

However, `Animation::next_key()` does this:

```
	int error = mlt_animation_next_key( instance, &item, position );
	if ( !error )
		return item.frame;
	else
		return error;
```
Thus, there is no way to differentiate currently between 1 as true for error and 1 as the frame number. We could change the return from the C function as -1. Then, it is still true for error and not a valid frame number. -1 is not the MLT convention for true, but this would need to be an exception and documented.

src/tests/test_animation/test_animation.cpp
ddennedy: agreed

src/tests/test_animation/test_animation.cpp
bmatherly: > We could change the return from the C function as -1

I thought of this. But then we would have to decide that negative frame values are illegal. Right now it is easy to get negative frame values by setting them negative with mlt_animation_key_set_frame() or mlt_animation_shift_frames().

Another option would be to create a new function that takes a pointer to an integer to be modified and returns a real error code. The current functions could be deprecated.

> However, it might affect Shotcut and be out-of-scope for the next release.

I agree. And maybe the priority is low. Just noticed this inconsistency when debugging and thought I would say something.

If you are inspired by a solution, I am willing to implement, test and integrate with Shotcut. Else, we can leave it as-is for now.

ddennedy: Yes, let's create the new methods, but let's make the parameters references instead of pointers to be consistent with the other methods in this class. You can overload these methods with the same name.

src/mlt++/MltAnimation.cpp
```diff
@@ -101,6 +101,15 @@ mlt_keyframe_type Animation::keyframe_type( int position )
 		return (mlt_keyframe_type) -1;
 }
 
+/** Get the keyfame at the position or the next following.
```
ddennedy, line 104: We do not maintain doxygen comments for the C++ wrapper BTW. You can keep it because you documented as deprecated, but I have no intention to add doxygen comments throughout.

bmatherly: I understand. But these functions to not map directly to the underlying C functions that they use. And I wanted the intent/design to be documented somewhere. Also, I wanted to mark them deprecated so that they show up the next time I search for "deprecated" functions to remove.

I also do not intend to add doxygen comments to the entire C++ API.

ddennedy (approved):

PR #761, opened 2021-12-28
Set RPATH on installed objects

rectalogic:
Building and running melt as follows on macOS 11.6.2 with XCode 13.2.1 and melt fails to find the shared library:
```sh-session
$ mkdir installed mlt-build
$ curl -L https://github.com/mltframework/mlt/releases/download/v7.4.0/mlt-7.4.0.tar.gz | tar xzf -
$ cd mlt-build
$ cmake --install-prefix $PWD/../installed -DMOD_GDK=OFF ../mlt-7.4.0
$ make
$ cmake --install .
$ ../installed/bin/melt
```
melt exits with:
```
dyld: Library not loaded: @rpath/libmlt-7.7.dylib
  Referenced from: /private/tmp/m/mlt-build/../installed/bin/melt
  Reason: image not found
Abort trap: 6
```

Setting rpath as described here https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#always-full-rpath fixes this.


CMakeLists.txt
```diff
@@ -73,6 +73,17 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/out/lib")
...
 
+# https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling
+set(CMAKE_SKIP_BUILD_RPATH FALSE)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
+if("${isSystemDir}" STREQUAL "-1")
+    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+endif("${isSystemDir}" STREQUAL "-1")
```
ddennedy, line 84: I know this is based on code from the cmake wiki, but a fully specified `endif` is inconsistent with the rest of the CMakeLists.txt files, and I am not a fan of it. The [docs](https://cmake.org/cmake/help/latest/command/endif.html) say it is only for backwards compatibility, but we specify a minimum CMake version; so, it is not needed.

Besides that, I still need to test more if this change has any weird side effect on Windows, MacPorts, or Shotcut's relocatable builds. That is kind of a pain. For some of this, I might end up accepting and then restricting if needed.

ddennedy (changes requested):
When I run this on Linux with ninja, it does not change the RUNPATH of `bin/melt-7` at install time. I do see it change the RUNPATH of libraries. I do see it set the RUNPATH in the build tree, which is cool. But
```
ddennedy@Gamerbox-9400:~/src/mlt/build$ objdump -x ~/opt/bin/melt-7 | grep RUNPATH 
  RUNPATH              /home/ddennedy/src/mlt/build/out/lib
```

ddennedy:
Windows (not used) and macOS checks out for me. Reviewing the build script for macOS, and it should be OK with how it modifies RPATH when deploying to app bundle. On Linux, I also tried with make instead of ninja and setting `CMAKE_INSTALL_RPATH_USE_LINK_PATH` false with no change on the installed RPATH for `bin/melt`. I think that would be good to get working in this PR.

But there is another thing I am confused about. Why does it set `CMAKE_INSTALL_RPATH` and then again later when `CMAKE_INSTALL_PREFIX` is not a system directory? What happens when it is a system directory? You copied the code from "Always full RPATH" section of the wiki doc, but I believe Linux distribution packages do not want a RUNPATH. I checked `/usr/bin/qgit` and it has no RPATH or RUNPATH. So, I think this needs to be like

```
# https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling
list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
if("${isSystemDir}" STREQUAL "-1")
    set(CMAKE_SKIP_BUILD_RPATH FALSE)
    set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
else()
    set(CMAKE_SKIP_BUILD_RPATH TRUE)
endif()
```

rectalogic:
> Why does it set `CMAKE_INSTALL_RPATH` and then again later when `CMAKE_INSTALL_PREFIX` is not a system directory?

Yeah that's odd. I think it should set `CMAKE_INSTALL_RPATH` and then clear it if it is a system directory, so that it strips RPATH when installing - we don't need any RPATH if it's being installed to the system directory. So we need to `set(CMAKE_INSTALL_RPATH "")` if installing to a system directory.

We want the binaries in the build directory to have the build RPATH, but when they are installed we want that replaced with the installed RPATH, unless they are being installed to system directories in which case we want no RPATH.

So it seems like it should be like this:
```cmake
set(CMAKE_SKIP_BUILD_RPATH FALSE)
set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
if(NOT "${isSystemDir}" STREQUAL "-1")
    set(CMAKE_INSTALL_RPATH "")
endif()
```

On macOS it does change the RPATH at install time, so the built `melt` and the installed `melt` have different RPATH:

```sh-session
# macOS
# mlt source /Users/aw/Projects/rectalogic/webvfx/vendor/mlt
# mlt build /Users/aw/Projects/rectalogic/webvfx/build/macos/mlt

$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/mlt -DMOD_GDK=OFF ../../../vendor/mlt
$ make
$ cmake --install .

$ otool -l out/bin/melt
Load command 16
          cmd LC_RPATH
      cmdsize 80
         path /Users/aw/Projects/rectalogic/webvfx/build/macos/mlt/out/lib (offset 12)

$ otool -l /tmp/mlt/bin/melt
Load command 18
          cmd LC_RPATH
      cmdsize 32
         path /tmp/mlt/lib (offset 12)

# installed binary was modified
$ cmp out/bin/melt /tmp/mlt/bin/melt
out/bin/melt /tmp/mlt/bin/melt differ: char 21, line 1
```

On linux, as you noticed, it doesn't seem to set RPATH at all but sets RUNPATH, and does not modify the installed binary at all:

```sh-session
# linux
# mlt source /webvfx/vendor/mlt
# mlt build /webvfx/build/linux/mlt

$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/mlt  ../../../vendor/mlt
$ make
$ cmake --install .

$ objdump -x out/bin/melt | grep RPATH
$ objdump -x out/bin/melt | grep RUNPATH
  RUNPATH              /webvfx/build/linux/mlt/out/lib

$ objdump -x /tmp/mlt/bin/melt | grep RPATH
$ objdump -x /tmp/mlt/bin/melt | grep RUNPATH
  RUNPATH              /webvfx/build/linux/mlt/out/lib

# installed binary is identical
$ cmp out/bin/melt /tmp/mlt/bin/melt
```

If I do the above build on linux using mlt master (none of these RPATH changes), I get the same results. RUNPATH is set on both build and installed melt to the build dir, and RPATH is not set.

With the above RPATH settings, I do see cmake log that it is resetting RUNPATH on the shared libraries at install time:

```
-- Installing: /tmp/mlt/lib/mlt-7/libmltrtaudio.so
-- Set runtime path of "/tmp/mlt/lib/mlt-7/libmltrtaudio.so" to "/tmp/mlt/lib"
```

And the build libraries have different RUNPATH than the installed libraries:
```sh-session
$ objdump -x /tmp/mlt/lib/mlt-7/libmltrtaudio.so | grep RUNPATH
  RUNPATH              /tmp/mlt/lib
$ objdump -x out/lib/mlt/libmltrtaudio.so | grep RUNPATH
  RUNPATH              /webvfx/build/linux/mlt/out/lib:
```

So it seems that linux uses RUNPATH instead of RPATH, and the above RPATH cmake change only applies to libraries and not executables on linux.

This still seems useful since the library modules also need to find the mlt sharedlib in either the build or install directories:

```sh-session
$ ldd /tmp/mlt/lib/mlt-7/libmltrtaudio.so
	libmlt-7.so.7 => /tmp/mlt/lib/libmlt-7.so.7 (0x00007fd102798000)
$ ldd out/lib/mlt/libmltrtaudio.so
	libmlt-7.so.7 => /webvfx/build/linux/mlt/out/lib/libmlt-7.so.7 (0x00007fccb1ac0000)
```

However since cmake doesn't seem to touch executables, both the installed and built melt have the same RUNPATH and so the installed melt cannot be run if the build directory is removed.

When I build using mlt master (no RPATH changes), the installed libraries cannot find libmlt:
```sh-session
$ ldd out/lib/mlt/libmltrtaudio.so | grep libmlt
	libmlt-7.so.7 => /webvfx/build/linux/mlt/out/lib/libmlt-7.so.7 (0x00007f0866c1d000)
$ ldd /tmp/mlt/lib/mlt-7/libmltrtaudio.so | grep libmlt
	libmlt-7.so.7 => not found
```

OK, the reason it is not correctly setting RUNPATH on melt is because of the way we install melt.  If I change src/melt/CMakeLists.txt to use `install(TARGETS melt RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})`, then it correctly sets RUNPATH at install time.

```
-- Installing: /tmp/mlt/bin/melt
-- Set runtime path of "/tmp/mlt/bin/melt" to "/tmp/mlt/lib"
```

```sh-session
$ objdump -x /tmp/mlt/bin/melt | grep RUNPATH
  RUNPATH              /tmp/mlt/lib
$ objdump -x out/bin/melt | grep RUNPATH
  RUNPATH              /webvfx/build/linux/mlt/out/lib:
```

rectalogic:
Also, with the above changes, if I install to a system dir then it erases RUNPATH on install since it's not necessary:
```sh-session
$ cmake -DCMAKE_INSTALL_PREFIX=/usr ../../../vendor/mlt
$ make
$ cmake --install .
```

```
-- Installing: /usr/bin/melt
-- Set runtime path of "/usr/bin/melt" to ""
```

```
$ objdump -x /usr/bin/melt | grep RUNPATH
$ objdump -x out/bin/melt | grep RUNPATH
  RUNPATH              /webvfx/build/linux/mlt/out/lib:
```

ddennedy (approved):

ddennedy:
Thanks for figuring out the melt install problem.

PR #770, opened 2022-02-12
Deliver the true average value for the audio level

bmatherly:
Posting this here for discussion (not a real change proposal)
Motivated by this report:
https://forum.shotcut.org/t/audio-peak-meter-shows-different-values/32254/6

What is the "Audio Peak Meter" in Shotcut supposed to represent?
I thought that it would represent the peak sample value for each frame.

But when I investigate the code in the audio level filter, I see that it is averaging the sample values. However, it is some kind of scaled average because of two reasons:
1) The s16 value is divided by 128 and then by 127 later. So it is divided by 16,256 total. However, to scale to 0-1.0, it should be scaled by 2^11 (32,768)
2) It is also scaled to 40/42 = leaving 41 and 42 as "overshoot" values

So the value returned by the filter can not be easily converted to a true dB value unless the two scaling factors are undone. And then it is only an "average" value, not a peak value.

The Shotcut meter is taking the value returned from the filter and converting to dB. The result is that when a -9dB sine wave is presented, the meter shows values between -7.1dB and -7.2dB (higher than the actual peak in the audio!)

The edits in this code review demonstrate the changes that would need to be made to return "average" audio level.

What to do?
* Leave as is (maybe hide the actual values)
* Modify the audio level filter to return a true average value (scaled to 0-1.0) - this patch. Then, rename the Shotcut meter to "audio level" or something similar.
* Make a new peak filter that actually returns the peak value - use that in Shotcut instead?

### DIFF
--- a/src/modules/normalize/filter_audiolevel.c
+++ b/src/modules/normalize/filter_audiolevel.c
@@ -79,7 +79,8 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 
 		for ( s = 0; s < num_samples; s++ )
 		{
-			double sample = fabs( pcm[c + s * num_channels] / 128.0 );
+			double sample = fabs( (double)pcm[c + s * num_channels] / 32768.0 );
+
 			val += sample;
 			if ( sample == 128 )
 				num_oversample++;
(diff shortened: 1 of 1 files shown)


ddennedy:
This is old code, and I ported it from somewhere else I forget probably without fully understanding it. I know I got the `IEC_Scale()` code from Steve Harris' [meterbridge](http://plugin.org.uk/meterbridge/), but the main signal measuring code is not from there. I do not recall why not. I just tested a sine tone from Shotcut through meterbridge's dpm, and it is accurate. I have not formed an opinion yet about what to do.

ddennedy:
> Modify the audio level filter to return a true average value (scaled to 0-1.0) - this patch. 

Do this for the waveform like audio levels.

> Make a new peak filter that actually returns the peak value 

Do this possibly based on meterbridge's dpm code.

bmatherly:
The calculation for peak values is so simple that it was easier to calculate it directly in the application rather than using a filter. So I changed the Shotcut meter to perform the calculation itself:
https://github.com/mltframework/shotcut/commit/f5928ed3dcdd57dcbb7f311cadfb1224ce0b034a

I think the current audiolevels filter is OK for creating non-calibrated views of audio (like the waveforms). So I will probably not make any changes to that filter at this time. I am willing to modify the filter to provide calibrated values (peak, RMS, etc) in the future if anyone needs it.

PR #776, opened 2022-03-06
Box Blur Improvements

bmatherly:
This implementation has 2x performance of the boxblur filter.
It also supports slices.

Use this new box blur function in pillar_image filter
Implement a new box_blur filter using this function

Performance comparison between old and new filters:

**boxblur (old)**
```
time ./Shotcut/Shotcut.app/melt noise out=500 -filter volume:0 -filter boxblur blur=40 -consumer null terminate_on_pause=1 real_time=-1

real	0m24.335s
user	0m24.239s
sys	0m0.070s
```

**box_blur (new)**
```
time ./Shotcut/Shotcut.app/melt noise out=500 -filter volume:0 -filter box_blur -consumer null terminate_on_pause=1 real_time=-1

real	0m6.789s
user	0m12.598s
sys	0m0.139s
```

For discussion:
* Do we want to continue to add image processing functions to the mlt_image module? Or maybe they should go some where else?
* Different name? "box_blur" is not very different from "boxblur".
* Deprecate the old boxblur filter?
* The h/vradius parameters units are: "The horizontal blur radius as a percent of the image size. 100% results in a blur radius of 10% of the image width.". Do we like this, or something else?


src/framework/mlt_image.c
```diff
@@ -425,3 +426,215 @@ void mlt_image_format_planes( mlt_image_format format, int width, int height, vo
...
 }
+
+typedef struct
+{
+	mlt_image src;
+	mlt_image dst;
+	int radius;
+} blur_slice_desc;
+
+static int blur_h_proc(int id, int index, int jobs, void* data)
```
ddennedy, line 437: Image processing functions do not belong in the framework.

src/modules/plus/filter_pillar_echo.c
```diff
@@ -414,7 +251,11 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
 	uint8_t* dst = mlt_pool_alloc( size );
 
 	bilinear_scale_rgba( *image, dst, *width, *height, rect );
-	box_blur( dst, dst, *width, *height, blur );
+	if (blur != 0) {
+		struct mlt_image_s img;
+		mlt_image_set_values( &img, dst, *format, *width, *height );
+		mlt_image_box_blur( &img, blur, blur );
```
ddennedy, line 257: To avoid code duplication the pillar_echo filter should encapsulate a blur filter that can be changed from its default via a string property (e.g. "blur_filter") including properties that begin with "blur." for example to set properties on the nested filter. This is a pattern we have used elsewhere.

ddennedy (changes requested):

src/framework/mlt_image.c
bmatherly: I can accept that. The idea was inspired by FFmpeg which has a gaussian function in swcale. It is part of the public interface, and multiple filters use it.
https://github.com/FFmpeg/FFmpeg/blob/069f7831a2bd1545080121fb6d0773191cd61d70/libswscale/swscale.h#L351

Maybe someday there could be a module in the framework for useful processing functions. I will make a separate proposal if I come up with something.

src/modules/plus/filter_pillar_echo.c
bmatherly: I am familiar with the service encapsulation method. I do not like it for this case because of the extra resource overhead and complication:
* Requires the construction of an entire extra service and passing parameters using the properties
* Requires constructing an extra frame to pass the image into the service
* Does not support frame threading unless the parameters are attached to the frame and interpreted by the encapsulated filter
All this extra complication and overhead - and all I want to do is have a single function that is shared by two files.

I will move this function into a shared file at the module level as we have done for other modules:
https://github.com/mltframework/mlt/tree/master/src/modules/plus/ebur128
https://github.com/mltframework/mlt/blob/master/src/modules/plusgpl/utils.h
https://github.com/mltframework/mlt/blob/master/src/modules/qt/graph.h

bmatherly:
I moved the blur function out of mlt_image

Also added slice processing to the scaling operation in pilar_echo.

Pillar echo improvements for this command on a 4 core machine:
`time melt noise out=500 -filter volume:0 -filter pillar_echo rect="5% 5% 90% 90%" -consumer null  terminate_on_pause=1 real_time=-1`

Before:
real	0m22.636s
user	0m22.599s

After
real	0m9.621s
user	0m23.985s

src/framework/mlt_image.c
ddennedy: This idea is a little complicated within MLT proper. There could be a libmltimgproc library sub-project added that would be a better fit. But then, is there not already enough image processing libraries? And why go into this trouble when all image processing must be completely revisited for higher than 8-bit processing?

src/modules/core/filter_box_blur.yml
```diff
@@ -0,0 +1,35 @@
...
+    minimum: 0
+    default: 1
+
+  - identifier: vradius
+    title: Vertical radius
+    description: >
+      The vertical blur radius as a percent of the image size.
+      100% results in a blur radius of 10% of the image width.
+    type: integer
+    unit: pixels
```
ddennedy, line 31: unit should be percentage (or percent)

src/modules/core/filter_box_blur.c
```diff
@@ -0,0 +1,92 @@
...
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+	mlt_position position = mlt_filter_get_position( filter, frame );
+	mlt_position length = mlt_filter_get_length2( filter, frame );
+	mlt_profile profile = mlt_service_profile( MLT_FILTER_SERVICE( filter ) );
+	double hradius = mlt_properties_anim_get_double( properties, "hradius", position, length );
+	double vradius = mlt_properties_anim_get_double( properties, "vradius", position, length );
+	// Convert from percent to pixels
+	hradius = hradius * (double)profile->width * mlt_profile_scale_width( profile, *width ) / 1000.0;
+	hradius = MAX(lrint(hradius), 0);
+	vradius = vradius * (double)profile->width * mlt_profile_scale_width( profile, *width ) / 1000.0;
```
ddennedy, line 42: Use `mlt_profile_scale_height()` here

src/modules/core/filter_box_blur.yml
```diff
@@ -0,0 +1,35 @@
...
+tags:
+  - Video
+parameters:
+
+  - identifier: hradius
+    title: Horizontal radius
+    description: >
+      The horizontal blur radius as a percent of the image size.
+      100% results in a blur radius of 10% of the image width.
+    type: integer
```
ddennedy, line 18: I am confused a little about integer and percent here because it is being get-ed in the code as a double, and some projects such as frei0r use [0.0, 1.0] as a range for a percentage. I understand it a little better from the description but still kind of awkward. Since there is no maximum, then I can use 1000% for 100% of the image dimension. So, why not keep it simpler and use [0.0, 1.0] or [0.0, 100.0] where 100% = 100% and let me use higher precision than 0.1% of the image dimension? I think you are trying to map the most usable portion of the range into [0, 100] but also not make it limited. I think that makes sense for a UI, and this can simplify a UI for this; but not so sure it makes sense here. I will not object but at the very least, change the type to float.

src/modules/core/filter_pillar_echo.c
```diff
@@ -0,0 +1,314 @@
...
+	mlt_image src = desc->src;
+	mlt_image dst = desc->dst;
+	mlt_rect rect = desc->rect;
+	int slice_line_start, slice_height = mlt_slices_size_slice(jobs, index, src->height, &slice_line_start);
+	int slice_line_end = slice_line_start + slice_height;
+	double srcScale = rect.h / (double)src->height;
+	int linesize = src->width * 4;
+	uint8_t* d = dst->data + (slice_line_start * linesize);
+	int y = 0;
+	for ( y = slice_line_start; y < slice_line_end; y++ )
```
ddennedy, line 81: Just a FYI for a while now we have supported using C99 so you can declare within a for loop, for example.

src/modules/core/filter_pillar_echo.c
```diff
@@ -0,0 +1,314 @@
...
+	rect = constrain_rect( rect, profile->width * scale, profile->height * scale );
+
+	if ( rect.w < 1 || rect.h < 1 )
+	{
+		mlt_log_info( MLT_FILTER_SERVICE(filter), "rect invalid\n" );
+		return mlt_frame_get_image( frame, image, format, width, height, writable );
+	}
+
+	*format = mlt_image_rgba;
+	error = mlt_frame_get_image( frame, image, format, width, height, 1 );
```
ddennedy, line 258: Here you are requesting a writable image from upstream, but that is not required. That can cause some producers to be slightly less efficient.

ddennedy (changes requested):

src/modules/core/filter_box_blur.yml
bmatherly: >  I think you are trying to map the most usable portion of the range into [0, 100] but also not make it limited. I think that makes sense for a UI, and this can simplify a UI for this

You understand my dilemma perfectly. If I set the maximum range to be 100% of the image width, then in the UI I need to scale that down by 1/10th because 100% image blur is not useful in the application (and all the useful values would be in the first 10% of the slider). 

I was hoping you would offer a stronger opinion and then I would change it to whatever your preference is. I agree with changing the type to float. But let me know if you think of a better range/scale for this parameter.

src/modules/core/filter_box_blur.yml
```diff
@@ -0,0 +1,35 @@
+schema_version: 7.0
+type: filter
+identifier: box_blur
+title: Box Blur
+version: 3
```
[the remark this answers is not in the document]
bmatherly: Done

src/modules/core/filter_box_blur.yml
bmatherly: done

src/modules/core/filter_box_blur.c
bmatherly: I want the scale factor to be the same for H and V so the user does not have to guess-and-check the values if they want even h/v blur size. I refactored this code to make it more obvious.

src/modules/core/filter_box_blur.yml
bmatherly: Changed the type to float.

src/modules/core/filter_pillar_echo.c
bmatherly: Nice! Done.

src/modules/core/filter_pillar_echo.c
bmatherly: Good to know. Fixed!

bmatherly:
I believe all comments are resolved. Let me know if you want any other changes.

PR #777, opened 2022-03-07
Add slice processing to spot_remover filter

bmatherly:
Staging until a good time to merge...

Process each channel in a separate thread

Test case with worst case rectanble that fills the image:

no filter:
time ./Shotcut/Shotcut.app/melt noise out=500 -filter volume:0  -consumer null  terminate_on_pause=1 real_time=-1
real	0m1.544s
user	0m1.484s

single thread filter:
time ./Shotcut/Shotcut.app/melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 100% 100%" -consumer null  terminate_on_pause=1 real_time=-1
real	0m5.497s
user	0m5.425s

multi thread filter:
time ./Shotcut/Shotcut.app/melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 100% 100%" -consumer null  terminate_on_pause=1 real_time=-1
real	0m3.745s
user	0m5.585s

### DIFF
--- a/src/modules/plus/filter_spot_remover.c
+++ b/src/modules/plus/filter_spot_remover.c
@@ -79,19 +79,29 @@ static mlt_rect constrain_rect( mlt_rect rect, int max_x, int max_y )
 	return rect;
 }
 
+typedef struct
+{
+	 uint8_t* chan[4]; // pointer to the first value in the channel
+	 int rowCount[4];  // the number of values in each line (row)
+	 int step[4];      // the space between values in each line
+	 mlt_rect rect[4]; // rect the area to be removed
+} slice_desc;
+
 /** Perform spot removal on a channel.
   *
   * Values within the rectangle are replaced with interpolated values.
   * Each value is an interpolation of the first values outside of the rect on
   * the top, bottom, left and right of the value being interpolated.
-  *
-  * \param chan a pointer to the first value in the channel
-  * \param rowCount the number of values in each line (row)
-  * \param step the space between values in each line
-  * \param rect the area to be removed
   */
-static void remove_spot_channel( uint8_t *chan, int rowCount, int step, mlt_rect rect )
+static int remove_spot_channel_proc(int id, int index, int jobs, void* data)
 {
+	(void) id; // unused
+	(void) jobs; // unused
+	slice_desc* desc = ((slice_desc*) data);
+	uint8_t *chan = desc->chan[index];
+	int rowCount = desc->rowCount[index];
+	int step = desc->step[index];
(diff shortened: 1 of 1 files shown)


ddennedy:
Hmm, we have never done this before and made the job count fixed and not dependent upon core count. I think for a dual or single core this will be slower. Have you thought about this?

bmatherly:
> Hmm, we have never done this before and made the job count fixed and not dependent upon core count. I think for a dual or single core this will be slower. Have you thought about this?

In general, I think the typical use case of covering a small spot will always be negligible regardless of slices or not. Here are my test results for large spots vs small spots:

> No Filter
> time melt noise out=500 -filter volume:0  -consumer null  terminate_on_pause=1 real_time=-1
> real	0m1.544s
> user	0m1.484s
> 
> Large spot - No Slices
> time melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 100% 100%" -consumer null  terminate_on_pause=1 real_time=-1
> real	0m5.497s
> user	0m5.425s
> 
> Large spot - With Slices
> time melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 100% 100%" -consumer null  terminate_on_pause=1 real_time=-1
> real	0m3.745s
> user	0m5.585s
> 
> Small spot - No Slices
> time melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 1% 1%" -consumer null  terminate_on_pause=1 real_time=-1
> real	0m1.658s
> user	0m1.559s
> 
> Small spot - With Slices
> time melt noise out=500 -filter volume:0 -filter spot_remover rect="0% 0% 1% 1%" -consumer null  terminate_on_pause=1 real_time=-1
> real	0m1.753s
> user	0m1.666s

For small spots, slice threading will always be slower than single threading (regardless of number of actual cores) because the  thread context switch time outweighs the processing time. But the difference is negligible, I think.

I do not think it is important to try to accommodate such negligible performance improvements for single and dual core systems. 

This patch is actually performing "plane thread processing", not "slice thread processing". The next level of performance improvement would be to perform actual slice thread processing for large spots. But that would require implementing more complex heuristic logic to determine if the spot is large enough to divide by the number of system cores. For example, if a spot is 10x10 pixels, it would not be worthwhile to perform slice processing. If it is 1000x1000, maybe it is worthwhile to split across 16 cores. But not 1000x10 nor 10x1000. I decided not to add the complexity to optimize large spots because I think that is an uncommon use case. Also, it was easy to implement plane thread processing because I did not have to change any of the loop logic.

PR #779, opened 2022-03-16
add slice threading to some oldfilm filters

ddennedy:


src/modules/oldfilm/filter_lines.c
```diff
@@ -37,13 +87,14 @@ static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format
...
-		int w = *width;
-
 		int line_width = mlt_properties_anim_get_int( properties, "line_width", pos, len );
 		int num = mlt_properties_anim_get_int( properties, "num", pos, len );
 		double maxdarker = (double) mlt_properties_anim_get_int( properties, "darker", pos, len );
 		double maxlighter = (double) mlt_properties_anim_get_int( properties, "lighter", pos, len );
+		int full_range = mlt_properties_get_int(MLT_FRAME_PROPERTIES(frame), "full_range");
+		int min = full_range? 0 : 16;
+		int max_luma = full_range? 255 : 235;
+		int max_chroma = full_range? 255 : 240;
```
ddennedy, line 97: @bmatherly I am determining the range limits quite frequently now, and perhaps it should go into the framework. Also, it can be easy to swap 235 and 240. On one hand, it looks to belong to mlt_image. On the other, mlt_frame since it depends on the full_range frame property. I have in the past thought about a mlt_utils.[hc] for something like this, and it could have `mlt_fast_rand` as well. However, please be aware that I changed the `fast_rand` here to ensure the unsigned value fits into an int since that is what `rand()` does, and it was required to prevent unwanted behavior in a filter.

bmatherly: It is my intention that the MLT image object will eventually contain image properties such as aspect ratio, colorspace and full range. But those properties have not been added to the object yet.

Maybe we could bridge that gap with a helper function in mlt_image that will eventually become a member function when the properties are moved into the structure. Regardless, I prefer mlt_image for the chroma/luma max/min.

For the rand functions, I prefer one of these options:
* mlt_math
* mlt_calc
* mlt_util

In general, I dislike files/folders named "util" or "utilities" because they tend to become messy without a defined responsibility. But I understand it makes sense sometimes.

If you are considering a home for the rand() functions in the framework, also consider moving the MAX/MIN/CLAMP functions:

PR #780, opened 2022-03-22
Make it possible to build against Qt6

jlskuz:
This adds a `BUILD_WITH_QT6` cmake option that is disable by default.

Note: This PR is only about the build, I haven't made deeper checks on whether features still work with Qt6.


jlskuz:
Fixed now as suggested

ddennedy:
Today, I did some testing against Qt 6.2.4, and everything seems to be working. I was not able to test kdenlivetitle and typewriter. Also, I did not test every aspect of every other service; just basic - what we call "smoke" - testing.

jlskuz:
Nice to hear!

PR #781, opened 2022-03-22
Remove code that was only used with Qt4

jlskuz:
As far as I can see in the CMake files building against Qt4 is not supported anymore hence we don't need code and version checks for Qt < 5 anymore. Correct me if I am wrong!

This PR also adds QT_VERSION_CHECK everywhere for consistency

**Question:** It seems the `"consumer-cleanup"` event was only used in code used with Qt4. Should the `"consumer-cleanup"` be removed?


ddennedy:
Keep the consumer-cleanup event. It could be handy.

jlskuz:
Ah, maybe better to wait for #780 to avoid merge conflicts

Issue #798, opened 2022-06-19
A couple of modules have no return statement in function returning non-void


frispete:
Attempting to update the package from 7.4.0 to 7.6.0 with openSUSE Tumbleweed (GCC 12) reveals another issue:

[    6s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/kdenlive/filter_wave.c:81:1: error: no return statement in function returning non-void [-Werror=return-type]
[    7s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/oldfilm/filter_tcolor.c:54:1: error: no return statement in function returning non-void [-Werror=return-type]
[    7s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/plus/filter_invert.c:57:1: error: no return statement in function returning non-void [-Werror=return-type]
[    7s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/core/filter_pillar_echo.c:162:1: error: no return statement in function returning non-void [-Werror=return-type]
[    7s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/core/filter_mirror.c:302:1: error: no return statement in function returning non-void [-Werror=return-type]
[    7s] /home/abuild/rpmbuild/BUILD/mlt-7.6.0/src/modules/plus/filter_spot_remover.c:129:1: error: no return statement in function returning non-void [-Werror=return-type]

Since the behavior of such code is not defined, and could compiled with -Wno-return-type compile option, it would be nice to address this.

PR #799, opened 2022-06-19
Supply a proper return value on non-void functions

frispete:
Addresses #798


ddennedy:
Thanks

frispete:
Wonderful, thanks again.

PR #804, opened 2022-06-23
Another take on fixing the wcrtomb issue

frispete:
that boils down to a problem in glibc:
https://github.com/bminor/glibc/commit/9bcd12d223a8990254b65e2dada54faa5d2742f3
that appears, if sources are compiled with _FORTIFY_SOURCE=3, which is standard
by a few distributions already.

Make sure, a pointer, given to wcrtomb as first argument always has at least
MB_CUR_MAX bytes available.


frispete:
Sorry for being a bit too late for, what I think, being the correct fix.

PR #805, opened 2022-06-24
fix unguarded null ptr

OndrejPopp:
Hi Dan,
Here is one unguarded nullptr fix because mlt_environment can return NULL
Could be that this situation is rare, and happens with accesses when mlt is not properly initialized (yet)
I ran into this issue while implementing a new feature and refactoring kdenlive. Due to my refactoring modifications I needed to do the initialialisation of the mlt factory a bit sooner and in the meantime this bug surfaced.

cheers,
Ondrej

### DIFF
--- a/src/framework/mlt_profile.c
+++ b/src/framework/mlt_profile.c
@@ -363,9 +363,17 @@ mlt_properties mlt_profile_list( )
 	if ( prefix == NULL )
 	{
 		prefix = mlt_environment( "MLT_DATA" );
-		filename = calloc( 1, strlen( prefix ) + strlen( PROFILES_DIR ) + 1 );
-		strcpy( filename, prefix );
-		strcat( filename, PROFILES_DIR );
+		if (prefix)
+		{
+			filename = calloc( 1, strlen( prefix ) + strlen( PROFILES_DIR ) + 1 );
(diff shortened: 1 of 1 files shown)


ddennedy:
The problem with this change is that it turns something that clearly fails when the framework state is invalid (not initialized) into something mysteriously not working as expected: MLT profile not loaded because it will not be found in `/profiles/` in the filesystem.

OndrejPopp:
Well, I already had that somewhere else in Mlt where upon failure the default 4:3 profile is returned while an other has been requested. So then, you end up in kdenlive looking at a  bunch of 4:3 profiles instead of the other ones. Just as mysterious thus. Took me a while to track this down while at first I suspected memory corruption, but then I found out what the problem was while mlt started to crash over there.

I would say, personally, there are already some quards over there so one more wouldn't hurt...

Then, if you want to avoid mysterious behavior, then I think you should also change the return of the default 4:3 profile on failure and replace that with an assertion, including thus this point over here where the library crashes.

Then again the end user is not really helped with crashes or assertions so also here it is maybe better to allow for some mysterious default behavior instead of coredumps...

In any case, now you know that Mlt may crash over there possibly when not properly initialized...


⁣Get BlueMail for Android ​

Issue #801, opened 2022-06-21
Make Mlt with Qt5 and Qt6 co-installable


jlskuz:
The packager of openSUSE said that it would be necessary to make mlt with Qt5 and mlt with Qt6 or at least the modules co-installable. You can find the discussion here: https://invent.kde.org/teams/frameworks-devs/kf6-workboard/-/issues/8#note_457528

However I am a bit lost since I do not have enough knowledge on distro packaging and the insides of MLT's module loading to assess the right approach. Ones something is decided I am happy to help with a merge request.

While writing this I also wonder weather it is really necessary to have them co-installable? Theoretically there should not be an issue to run a Qt5 Shotcut, Kdenlive or what ever with a Qt6 mlt or am I wrong? _(EDIT: probably I am wrong (?) since I just thought of `melt` and forgot about our apps linking to mlt)_

ddennedy:
In the past it was instant death (crash on launch) to mix major Qt versions in the same process. This is based primarily Kdenlive using Qt5 was loading MLT built with the OpenCV module, and the OpenCV libraries were linked to Qt4. This goes a way back, but I would not expect to be able to mix them or at least to trust it. 

This request sounds similar to the SDL1/2 situation where we do have 2 MLT modules. That was not supposed to last long, but Flowblade is still using SDL1. In that case, there was code duplication but also changes. Since we now have the qt module working with both Qt 5 and 6 I do not like the idea of code duplication. There is probably a way to build the module twice with CMake that hopefully is not too much work and ugliness. As for the loading, I think it have to be done with a new environment variable. Maybe instead of being Qt specific, it can be generic like a deny list based on the basename minus the extension. For example, I could use `MLT_REPOSITORY_DENY=libmltsdl:libmltqt` to tell it to skip loading the SDL1 and Qt5 modules.

jlskuz:
Okay, thanks for the feedback!

First step of introducing `MLT_REPOSITORY_DENY``done. I guess ones we have two Qt modules we should make `MLT_REPOSITORY_DENY=libmltsdl:libmltqt6` the default?

ddennedy:
libmltsdl will not be in the default as the dual modules are not problematic in its current state. With Qt it is different because I do not want to have new/differently named services per Qt version.

ddennedy:
fixed in #807

PR #806, opened 2022-06-24
Add MLT_REPOSITORY_DENY envvar to skip plugin loading

jlskuz:
Use colon to separate entries

Example:
MLT_REPOSITORY_DENY=libmltplus:libmltavformat:libmltfrei0r

Related to #801

### DIFF
--- a/src/framework/mlt_repository.c
+++ b/src/framework/mlt_repository.c
@@ -98,11 +98,32 @@ mlt_repository mlt_repository_init( const char *directory )
 	free(newpath);
 #endif
 
+	mlt_tokeniser tokeniser = mlt_tokeniser_init();
+	int dl_length = mlt_tokeniser_parse_new( tokeniser, getenv( "MLT_REPOSITORY_DENY" ), ":" );
+
 	// Iterate over files
 	for ( i = 0; i < count; i++ )
 	{
 		int flags = RTLD_NOW;
 		const char *object_name = mlt_properties_get_value( dir, i);
+
+		// check if the plugin was asked to be skipped through MLT_REPOSITORY_DENY
+		int ignore = 0;
+		for (int j = 0; j < dl_length; j++ )
+		{
+			char *denyfile = calloc( 1, strlen( directory ) + strlen( mlt_tokeniser_get_string( tokeniser, j ) ) + 2 );
+			sprintf (denyfile, "%s/%s.", directory, mlt_tokeniser_get_string( tokeniser, j ));
+			ignore += !strncmp(object_name, denyfile, strlen( denyfile ) );
+			free (denyfile);
+		}
+
+		if (ignore)
+		{
+			mlt_log_info(NULL, "%s: skip plugin %s\n", __FUNCTION__, object_name);
+			continue;
+		}
+
+		mlt_log_debug(NULL, "%s: processing plugin at %s\n", __FUNCTION__, object_name);
 
 		// Open the shared object
 		void *object = dlopen( object_name, flags );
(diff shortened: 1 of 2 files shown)


src/framework/mlt_repository.c
```diff
@@ -98,12 +98,51 @@ mlt_repository mlt_repository_init( const char *directory )
 	free(newpath);
 #endif
 
+	char **denylist  = NULL;
+	int dl_length = 0;
+	if (getenv( "MLT_REPOSITORY_DENY" ))
+	{
+		char *p = strtok (getenv( "MLT_REPOSITORY_DENY" ), ":");
```
ddennedy, line 105: How about using `mlt_tokeniser_parse_new()` and related?

src/framework/mlt_repository.c
```diff
@@ -98,12 +98,51 @@ mlt_repository mlt_repository_init( const char *directory )
...
 	for ( i = 0; i < count; i++ )
 	{
 		int flags = RTLD_NOW;
 		const char *object_name = mlt_properties_get_value( dir, i);
 
+		// check if the plugin was asked to be skipped through MLT_REPOSITORY_DENY
+		int ignore = 0;
+		for (int j = 0; j < dl_length; ++j)
+		{
+			char *denyfile = calloc( 1, strlen( directory ) + strlen( denylist[j] ) + strlen( LIBSUF ) + 1 );
```
ddennedy, line 132: I prefer not to consider explicit extension because on macOS it needs to allow either .so or .dylib. Rather it should consider basename minus the extension. In the case of multiple multiple periods in the basename the first one should be considered the start of the extension.

ddennedy (changes requested):

src/framework/mlt_repository.c
jlskuz: Wasn't aware of this, done.

src/framework/mlt_repository.c
jlskuz: Changed

src/framework/mlt_repository.h
```diff
@@ -26,6 +26,12 @@
 #include "mlt_types.h"
 #include "mlt_profile.h"
 
+#ifdef _WIN32
+#  define LIBSUF ".dll"
+#else
+#  define LIBSUF ".so"
+#endif
+
```
ddennedy, line 34: Did you intend to leave `LIBSUF` macro defined here? I am not so sure that is relevant. frei0r is the rare exception that a module needs to use dlopen, and like I said for macOS it might not even be correct for another module's need.

src/framework/mlt_repository.c
```diff
@@ -98,12 +98,33 @@ mlt_repository mlt_repository_init( const char *directory )
 	free(newpath);
 #endif
 
+	mlt_tokeniser tokeniser = mlt_tokeniser_init();
+	int dl_length = mlt_tokeniser_parse_new( tokeniser, getenv( "MLT_REPOSITORY_DENY" ), ":" );
```
ddennedy, line 102: This new environment variable should be documented via a doxygen comment in `mlt_repository.h` or `mlt_factory.h` to make it show up [here](https://www.mltframework.org/doxygen/envvars.html). 
Here is how that is done:
` * \envvar \em MLT_REPOSITORY overrides the default location of the plugin modules, defaults to \p PREFIX_LIB.
`
Thanks

ddennedy (changes requested):

ddennedy (approved):

PR #807, opened 2022-06-25
[qt module] Split to allow installing Qt5 and Qt6 in parallel

jlskuz:
That's how far I've come, but now I am stuck in errors like 

```
CMake Error at src/modules/qt/CMakeLists.txt:20 (add_library):
  Target "mltqt6" links to target "Qt::Core" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
Call Stack (most recent call first):
  src/modules/qt/CMakeLists.txt:114 (mlt_add_qt_module)
```

It seems that it tries to use versionless targets (`Qt::Core` instead of `Qt6::Core`) somewhere, but I don't know where and how to avoid that since `QT_NO_CREATE_VERSIONLESS_TARGETS` is already set :thinking:


jlskuz:
Works now. (Thanks @krop!)

krop:
```
if(BUILD_TESTING)
   find_package(Qt${QT_MAJOR_VERSION} REQUIRED COMPONENTS Core Test)
endif()
```

Needs to be moved between the `if(MOD_QT6) [...] endif'()` and `if(MOD_QT) [...] endif'()` blocks as well.

krop:
and maybe add a comment mentioning Qt6 needs to be found before Qt5 if both MOD_QT and MOD_QT6 are enabled

ddennedy:
Thanks!

PR #813, opened 2022-07-25
New gps graphic filter, updated gps text, move all gps from xml to qt

dany123:
This is a big update to the original GPS text filter (which was moved from XML to QT dir, switching the xml parser to qt's xml reader) including multiple bugfixes and improvements to the original parser and a new filter that I named "GPS Graphic" which uses the same parser but draws everything with QT (including the text, which weirdly is faster/more efficient than the GPS text at drawing the same exact text).

There are 2 bigger changes to the GPS text that also affect the ui: completely removed the timezone stuff from the code (because internal_timegm doesn't ruin times like the standard mktime) and added a new function to support GPS tracks that cross the 180 meridian.

-----

The GPS Graphic filter takes a GPS track file and can draw either a 2D map or a 1D graph from speed, altitude or heart rate (if present in file) with a few special colouring schemes. There is also a special Speedometer drawing and a "Follow" mode for the map/graph types. 
There's also a late addition to the filter (that came with a lot of headache): the ability to add an image as a background (intended to be used as a map, but can be anything).

I think I fixed most of the bugs, but there is still one thing I kinda wish I managed to patch: there is a weird stretching when using the "follow dot" mode, especially at the edge of the tracks. It is usually barely noticeable but becomes obvious if using a background with text in it as stretched letters on a map are very easy to spot. It's not a deal breaker, just annoying.

Here are some samples:

Full track (no background) colored by speed and a zoomed in view with background.
image

A full map with background and an altitude graph below:
image

The speedometer and a speed graph:
image


dany123:
> /home/runner/work/mlt/mlt/src/modules/qt/gps_drawing.cpp:62:11: error: ‘class QLineF’ has no member named ‘intersects’; did you mean ‘intersect’?

How can this throw an error? it exists: https://doc.qt.io/qt-5/qlinef.html#intersects

src/modules/qt/filter_gpstext.yml
```diff
@@ -143,6 +143,19 @@ parameters:
     mutable: yes
     widget: text
     
+  - identifier: swap_180_meridian
```
bmatherly, line 146: Why would this be an option? Shouldn't this be the only way it works?

bmatherly:
> How can this throw an error? it exists: https://doc.qt.io/qt-5/qlinef.html#intersects

intersects was added in Qt 5.14. This build is running against Qt 5.12.

src/modules/qt/filter_gpstext.yml
dany123: In theory yes, in practice because I didn't implement this from the begining I am basically working around the issue (instead of finding a 100% working fix - which I don't know how to/and also would probably need a lot more changes; I've seen bigger projects having major issues with the 180 meridian so I think this is great compared to before). I basically shifting all longitudes around starting at file read, process everything with the regular routines (that are not aware if values were shifted or not) and then when displaying data to the output I shift everything back on a need bassis.

I want to leave this to the user because it's not something a lot of people would use/need, and it only works correctly if the track doesn't both cross 180 and 0 meridians. So if someone does a full circle around the globe it will still fail.

This was mostly implemented for the graphic part bcause it just looked unnaceptable even if 1 single point crossed the 180:
image

dany123:
> intersects was added in Qt 5.14. This build is running against Qt 5.12.

Oh, I was under the impression we're using the 5.x LTS version so 5.15. 
I'll manually check for line intersection.

bmatherly:
> Oh, I was under the impression we're using the 5.x LTS version so 5.15.
> I'll manually check for line intersection.

Shotcut requires 5.15, but MLT does not.

Alternately, you could make 5.14 a compile time version check and exclude this module if the requirement is not met.

src/modules/qt/filter_gpstext.yml
bmatherly: I would like to see if we can find a general solution for this and exclude the user option. I am concerned that it will create confusion and questions from users.

I think it is a common convention to check the line length and always draw the shortest line possible between two points - assuming that there is never more than 180 degrees between two points.

Does your code also handle map edges? In the case of a whole world map, a line might have to be broken into two segments to wrap off the map and continue on the other side (rather than drawing a line across the whole map.

dany123: > I would like to see if we can find a general solution for this and exclude the user option. I am concerned that it will create confusion and questions from users.

Hmm, I could check if there's points both between 179 and 180 and 180 and -179 and enable it automatically. I'm worried about a full map circle though... I guess I could exclude the swap if there are points near 0.

>I think it is a common convention to check the line length and always draw the shortest line possible between two points - assuming that there is never more than 180 degrees between two points.

The way I draw is I first compute the gps bounds so a whole world would simply draw between -180,-90 and 180,90 then for each point (in draw_main_line_graph()) I compute it's location as a percentage between those edges and then compute that percentage of the drawing rect (in the get_gpspoint_to_rect() call)  and do a p.drawLine() between 2 consecutive points. This is actually where the lineF.intersects() is needed, for efficiency, I only draw points if the segment intersects the drawing rect.

>Does your code also handle map edges? 

It should handle edges but just as a result because of how I draw the lines.

>In the case of a whole world map, a line might have to be broken into two segments to wrap off the map and continue on the other side (rather than drawing a line across the whole map.

This is not treated and is the main reason I did the swap_180 workaround. 
I don't really have a good solution for this, what you're suggesting makes sense theoretically, but how would I check for that reliably? Keep in mind there are ~3600 pairs of points per hour of standard 1s GPS tracking and I need to check this every frame (because of the follow dot feature) so even a simple formula would affect export time (which again, I'd like to keep low, I'm currently getting realtime exporting speed on x265 and all core usage with close to100% cpu which I'm very happy with - and even ~3x on hardware encoding).
And to be honest, this is rather a niche situation, the most common usage would be a real gps track of <2hours at 1s interval over a short location which very likely stays on land. I don't think we should focus in detail on something as extreme as random points on the entire globe, at least not without fixing other more visible issues first.

dany123:
This might take a bit more time than expected :| for no reason at all my cmake decided to crap out and not work anymore (both in cmd and vscode) gcc.exe " is not able to compile a simple test program", I'll just clean up and restart from begining because it's driving me crazy... and I literally did not change anything at all since 3 days ago.

dany123:
So it compiles ok now but there's still some warnings related to printing long long types with %d but I'm not sure what will be ok for the official compiler, %ll and llu? and %z for size_t type?

ddennedy:
> So it compiles ok now but there's still some warnings related to printing long long types with %d

Use `PRId64` etc macros available through `#include <inttypes.h>`. You can search the code base for how we use them.

dany123:
>Use PRId64 etc macros available through #include <inttypes.h>. You can search the code base for how we use them.

I gave them a try but the macros in inttypes.h seem to be guarded by this:
`#if !defined(__cplusplus) || defined(__STDC_FORMAT_MACROS)`
and being a C++ file I need the `#define __STDC_FORMAT_MACROS` to have access to them, but the header is already included somewhere else first so it doesn't get re-included to know of my local #define. 

I think I need to leave them like this for now. Or avoid long long either by force casting to int the smaller values or splitting into 2 ints for printing.

ddennedy:
See "Format Macros" in https://en.cppreference.com/w/cpp/types/integer for the `<cinttypes>` header and hints about spacing.

dany123:
I'm afraid I can't get PRIx64 to work no matter what I try. And it doesn't seem to be a spacing issue, even something like this: `printf("%s\n", PRIu64 );` doesn't compile (or `#warning PRIu64` ).

src/modules/qt/filter_gpstext.yml
dany123: I've been thinking about this and I think removing the option might actually work for most people so I'll try to automatically detect if a swap is needed and just do it at file read regardless of anything else.

bmatherly:
This change compiles for me without warnings on Ubuntu:

```
diff --git a/src/modules/qt/gps_parser.cpp b/src/modules/qt/gps_parser.cpp
index bce4cd21..4f53da92 100644
--- a/src/modules/qt/gps_parser.cpp
+++ b/src/modules/qt/gps_parser.cpp
@@ -18,10 +18,14 @@
  */
 
 #define __USE_XOPEN
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
 
 #include "gps_parser.h"
 
+#include <inttypes.h>
+
 #define _x (const xmlChar*)
 #define _s (const char*)
 #define _qxml_getthestring(s) qUtf8Printable((s).toString())
@@ -386,7 +390,7 @@ void recalculate_gps_data(gps_private_data gdata)
 	}
 	if (gdata.gps_points_p == NULL) {
 		if ((*gdata.ptr_to_gps_points_p = (gps_point_proc*) calloc(*gdata.gps_points_size, sizeof(gps_point_proc))) == NULL) {
-			mlt_log_warning(gdata.filter, "calloc error, size=%d\n", *gdata.gps_points_size*sizeof(gps_point_proc));
+			mlt_log_warning(gdata.filter, "calloc error, size=%" PRId64 "\n", *gdata.gps_points_size*sizeof(gps_point_proc));
 			return;
 		}
 		else { //alloc ok
@@ -635,7 +639,7 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
 	}
 	if (gdata.gps_points_p == NULL) {
 		if ((*gdata.ptr_to_gps_points_p = (gps_point_proc*) calloc(*gdata.gps_points_size, sizeof(gps_point_proc))) == NULL) {
-			mlt_log_warning(gdata.filter, "calloc failed, size = %d\n", *gdata.gps_points_size * sizeof(gps_point_proc));
+			mlt_log_warning(gdata.filter, "calloc failed, size = %" PRId64 "\n", *gdata.gps_points_size * sizeof(gps_point_proc));
 			return;
 		}
 		else 
@@ -823,7 +827,7 @@ void qxml_parse_gpx(QXmlStreamReader &reader, gps_point_ll **gps_list, int *coun
 				last_time = crt_point.time;
 			}
 			else 
-				mlt_log_info(NULL, "qxml_parse_tcx: skipping point due to time [%d] %f,%f - crt:%d, last:%d\n", *count_pts, crt_point.lat, crt_point.lon, crt_point.time, last_time);
+				mlt_log_info(NULL, "qxml_parse_tcx: skipping point due to time [%d] %f,%f - crt:%" PRId64 ", last:%" PRId64 "\n", *count_pts, crt_point.lat, crt_point.lon, crt_point.time, last_time);
 		}
 	}
 }
@@ -901,7 +905,7 @@ void qxml_parse_tcx(QXmlStreamReader &reader, gps_point_ll **gps_list, int *coun
 				last_time = crt_point.time;
 			}
 			else 
-				mlt_log_info(NULL, "qxml_parse_tcx: skipping point due to time [%d] %f,%f - crt:%d, last:%d\n", *count_pts, crt_point.lat, crt_point.lon, crt_point.time, last_time);
+				mlt_log_info(NULL, "qxml_parse_tcx: skipping point due to time [%d] %f,%f - crt:%" PRId64 ", last:%" PRId64 "\n", *count_pts, crt_point.lat, crt_point.lon, crt_point.time, last_time);
 		}
 	}
 }
@@ -955,7 +959,7 @@ int qxml_parse_file(gps_private_data gdata)
 	}
 	if (qxml_reader.hasError()) 
 	{
-		mlt_log_info(NULL, "qxml_reader.hasError! line:%d, errString:%s\n", qxml_reader.lineNumber(), qUtf8Printable(qxml_reader.errorString()));
+		mlt_log_info(NULL, "qxml_reader.hasError! line:%d, errString:%s\n", (int)qxml_reader.lineNumber(), qUtf8Printable(qxml_reader.errorString()));
 		return 0;
 	}
 
@@ -964,7 +968,7 @@ int qxml_parse_file(gps_private_data gdata)
 	*gdata.ptr_to_gps_points_r = (gps_point_raw*) calloc(count_pts, sizeof(gps_point_raw));
 	gps_array = *gdata.ptr_to_gps_points_r; //just an alias
 	if (gps_array == NULL) {
-		mlt_log_error(gdata.filter, "malloc error (size=%d)\n", count_pts * sizeof(gps_point_raw));
+		mlt_log_error(gdata.filter, "malloc error (size=%" PRId64 ")\n", count_pts * sizeof(gps_point_raw));
 	}
 	*gdata.gps_points_size = count_pts;
 
@@ -987,4 +991,4 @@ int qxml_parse_file(gps_private_data gdata)
 // 	}
 
 	return 1;
-}
\ No newline at end of file
+}
```

src/modules/qt/gps_parser.h
```diff
@@ -29,20 +29,15 @@
 #include <time.h>
```
bmatherly, line 29: It is a good practice to have as few includes in a header file as possible. I think that many of these are not used. Or are only being used by cpp files that include this header. I suggest to delete the includes that are not require by this header and move includes to cpp files if they are used there.

dany123:
> This change compiles for me without warnings on Ubuntu:

I tried something exactly like that the first time.
Even with your change, still fails for me, might be a Windows thing.

```

$ cmake --build .
[1/4] Automatic MOC for target mltqt
[2/3] Building CXX object src/modules/qt/CMakeFiles/mltqt.dir/gps_parser.cpp.obj
FAILED: src/modules/qt/CMakeFiles/mltqt.dir/gps_parser.cpp.obj
C:\Qt\Tools\mingw810_64\bin\g++.exe -DGPL3 -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NO_DEBUG -DQT_SVG_LIB -DQT_WIDGETS_LIB -DQT_XML_LIB -DUSE_EXIF -DUSE_FFTW -DUSE_QT_OPENGL -Dmltqt_EXPORTS -IC:/Projects/Shotcut/src/mlt/src/modules/qt/mltqt_autogen/include -IC:/Projects/Shotcut/src/mlt/src/mlt++ -IC:/Projects/Shotcut/src/mlt/src/framework/.. -isystem C:/Qt/5.15.2/mingw81_64/include -isystem C:/Qt/5.15.2/mingw81_64/include/QtCore -isystem C:/Qt/5.15.2/mingw81_64/./mkspecs/win32-g++ -isystem C:/Qt/5.15.2/mingw81_64/include/QtXml -isystem C:/Qt/5.15.2/mingw81_64/include/QtSvg -isystem C:/Qt/5.15.2/mingw81_64/include/QtWidgets -isystem C:/Qt/5.15.2/mingw81_64/include/QtGui -isystem C:/Qt/5.15.2/mingw81_64/include/QtANGLE -isystem C:/msys64_v2/mingw64/include -Ic:/Projects/Shotcut/include -DHAVE_STRUCT_TIMESPEC -mmmx -msse -msse2 -std=c++14 -MD -MT src/modules/qt/CMakeFiles/mltqt.dir/gps_parser.cpp.obj -MF src\modules\qt\CMakeFiles\mltqt.dir\gps_parser.cpp.obj.d -o src/modules/qt/CMakeFiles/mltqt.dir/gps_parser.cpp.obj -c C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp
In file included from C:/Projects/Shotcut/src/mlt/src/framework/mlt.h:58,
                 from C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.h:22,
                 from C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:24:
C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp: In function 'void recalculate_gps_data(gps_private_data)':
C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:392:57: error: expected ')' before 'PRId64'
    mlt_log_warning(gdata.filter, "calloc error, size=%" PRId64 "\n", *gdata.gps_points_size*sizeof(gps_point_proc));
                                                         ^~~~~~
C:/Projects/Shotcut/src/mlt/src/framework/mlt_log.h:85:88: note: in definition of macro 'mlt_log_warning'
 #define mlt_log_warning(service, format, args...) mlt_log((service), MLT_LOG_WARNING, (format), ## args)
                                                                                        ^~~~~~
C:/Projects/Shotcut/src/mlt/src/framework/mlt_log.h:85:87: note: to match this '('
 #define mlt_log_warning(service, format, args...) mlt_log((service), MLT_LOG_WARNING, (format), ## args)
                                                                                       ^
C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:392:4: note: in expansion of macro 'mlt_log_warning'
    mlt_log_warning(gdata.filter, "calloc error, size=%" PRId64 "\n", *gdata.gps_points_size*sizeof(gps_point_proc));
    ^~~~~~~~~~~~~~~
C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:392:116: error: expected ')' before ';' token
    mlt_log_warning(gdata.filter, "calloc error, size=%" PRId64 "\n", *gdata.gps_points_size*sizeof(gps_point_proc));
                                                                                                                    ^
In file included from C:/Projects/Shotcut/src/mlt/src/framework/mlt.h:58,
                 from C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.h:22,
                 from C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:24:
C:/Projects/Shotcut/src/mlt/src/framework/mlt_log.h:85:58: note: to match this '('
 #define mlt_log_warning(service, format, args...) mlt_log((service), MLT_LOG_WARNING, (format), ## args)
                                                          ^
C:/Projects/Shotcut/src/mlt/src/modules/qt/gps_parser.cpp:392:4: note: in expansion of macro 'mlt_log_warning'
    mlt_log_warning(gdata.filter, "calloc error, size=%" PRId64 "\n", *gdata.gps_points_size*sizeof(gps_point_proc));
    ^~~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.

```

src/modules/qt/gps_parser.h
dany123: Missed this in my final cleanup, done.

src/modules/qt/filter_gpstext.yml
dany123: Automated this and removed user option in commit c954cbe
I have a single test .gpx file found on openstreetmap but seems to pass ok. Regular files are also left as is.

bmatherly:
> I tried something exactly like that the first time.
> Even with your change, still fails for me, might be a Windows thing.

Another option would be to use "%d" and then cast the parameters to (int) if you know they will not exceed INT_MAX. Otherwise, you probably just need to sort out the include order. This is why it is good to reduce the includes and especially exclude headers from the .h files (which I see you made good progress on).

ddennedy:
Please be aware that I merged a PR to better support Qt 6 that created a conflict. After you resolve this, I will check your code for Qt 6 compatibility since I realize that might not be convenient for you, and I am ready to do it.

dany123:
Ok, i'm on holiday for 1 more day so will update tue/wed.

dany123:
> After you resolve this, I will check your code for Qt 6 compatibility since I realize that might not be convenient for you, and I am ready to do it.

I tried to make qt6 work but even though I installed it with pacman it doesn't seem to be picked up by "cmake -G Ninja ."

src/modules/qt/filter_gpsgraphic.cpp
```diff
@@ -0,0 +1,716 @@
...
+		ret = (double) (val - min) / (max - min);
+	return ret;		
+}
+
+//computes used_crops from pdata->ui_crops (they change on each frame so can't store them in pdata)
+static void process_frame_properties (mlt_filter filter, mlt_frame frame, s_base_crops &used_crops)
+{
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+	private_data* pdata = (private_data*)filter->child;
+	int64_t time_now = get_current_frame_time_ms(filter, frame) + pdata->gps_offset;
```
ddennedy, line 292: `properties` and `time_now` are never used.

src/modules/qt/filter_gpsgraphic.cpp
```diff
@@ -0,0 +1,716 @@
...
+		used_crops.left = pdata->ui_crops.left;
+		used_crops.right = pdata->ui_crops.right;
+	}
+}
+
+static int filter_get_image( mlt_frame frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
+{
+	int error = 0;
+	mlt_filter filter = (mlt_filter)mlt_frame_pop_service( frame );
+	private_data* pdata = (private_data*)filter->child;
```
ddennedy, line 341: `pdata` is never used.

src/modules/qt/filter_gpstext.cpp
```diff
@@ -494,66 +426,19 @@ static void process_filter_properties(mlt_filter filter, mlt_frame frame)
...
-//	mlt_log_info(NULL, "get_gps_str, vid_time: %d s, i_gps=%d, time[i]=%d s\n", video_time_synced/1000, i_gps, pdata->gps_points_r[i_gps].time/1000);
-	gps_point_to_output(filter, i_gps, video_time_synced, keyword, gps_text);
-	strncat( text, gps_text, MAX_TEXT_LEN - strlen( text ) - 1);
-}
-
 /** Replaces file_datetime_now with absolute time-date string (video created + current timecode)
  *  (time includes speed_multiplier and updates per second) */
 static void get_current_frame_time_str(char* keyword, mlt_filter filter, mlt_frame frame, char* result )
 {
 	private_data* pdata = (private_data*)filter->child;
```
ddennedy, line 438: `pdata` here is never used.

src/modules/qt/gps_drawing.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+//draws a small dot/disc on the map/graph according to the current video time + sync
+void draw_now_dot(mlt_filter filter, mlt_frame frame, QPainter &p, s_base_crops &used_crops)
+{
+	private_data* pdata = (private_data*)filter->child;
+	mlt_rect rect = pdata->img_rect;
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+	int thickness =  mlt_properties_get_int( properties, "thickness" );
+	mlt_color dot_color = mlt_properties_get_color(properties, "now_dot_color");
+	QVector<QColor> colors = get_graph_colors( properties );
```
ddennedy, line 159: `colors` is never used.

src/modules/qt/gps_drawing.cpp
```diff
@@ -0,0 +1,647 @@
...
+		else if (color_style == gpsg_color_by_solid_past_future)
+		{
+			if (i <= i_now) 
+			{
+				p.setPen( pen_solid_color0 );
+			}
+			else if (i > i_now) 
+				p.setPen( pen_solid_color1 );
+		}
+		else if (color_style == gpsg_color_by_solid_past && i <= i_now || color_style == gpsg_color_by_solid_future && i > i_now)
```
ddennedy, line 277: This expression needs more parentheses: it is unreadable to mix && and || without them.

src/modules/qt/gps_drawing.cpp
```diff
@@ -0,0 +1,647 @@
...
+		{
+			p.setPen( pen_solid_color0 );
+		}
+		else if (color_style == gpsg_color_by_solid_past && i > i_now || color_style == gpsg_color_by_solid_future && i <= i_now)
```
ddennedy, line 281: This expression needs more parentheses: it is unreadable to mix && and || without them.

src/modules/qt/gps_drawing.cpp
```diff
@@ -0,0 +1,647 @@
...
+}
+
+//draws a speedometer from the current data source (for map it displays % of total)
+void draw_main_speedometer(mlt_filter filter, mlt_frame frame, QPainter &p, s_base_crops &used_crops)
+{
+	private_data* pdata = (private_data*)filter->child;
+	mlt_rect rect = pdata->img_rect;
+	mlt_properties properties = MLT_FILTER_PROPERTIES( filter );
+	int color_style = mlt_properties_get_int( properties, "color_style" ) ;
+	int thickness = qAbs( mlt_properties_get_int( properties, "thickness" ) );
```
ddennedy, line 425: `color_style` and `thickness` are never used.

src/modules/qt/gps_parser.cpp
```diff
@@ -663,15 +752,13 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
...
 
-/* xml parsing */
+//File parsing with QT's XML stream reader
 
-// Parses a .gcx file into a gps_point_raw linked list
-void xml_parse_gpx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *count_pts)
+// Parses a .gpx file into a gps_point_raw linked list
+void qxml_parse_gpx(QXmlStreamReader &reader, gps_point_ll **gps_list, int *count_pts)
 {
 	int i;
```
ddennedy, line 760: `i` is never used.

src/modules/qt/gps_parser.cpp
```diff
@@ -686,71 +773,68 @@ void xml_parse_gpx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *coun
...
-		else mlt_log_info(NULL, "xml_parse_gpx: skipping point due to time [%d] %f,%f - crt:%"PRId64", last:%"PRId64"\n", i, crt_point.lat, crt_point.lon, crt_point.time, last_time);
 	}
 }
 
+
 // Parses a .tcx file into a gps_point_raw linked list
-void xml_parse_tcx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *count_pts)
+void qxml_parse_tcx(QXmlStreamReader &reader, gps_point_ll **gps_list, int *count_pts)
 {
 	int i;
```
ddennedy, line 836: `i` is never used.

src/modules/qt/gps_parser.cpp
```diff
@@ -686,71 +773,68 @@ void xml_parse_gpx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *coun
...
-		}
-		if (xmlHasProp(crt_node, _x("lon"))) {
-			xmlChar* str = xmlGetProp(crt_node, _x("lon"));
-			crt_point.lon = strtod(_s(str), NULL);
-			xmlFree(str);
-		}
-		for (gps_val = crt_node->children; gps_val; gps_val = gps_val->next)
+		reader.readNext();
+		// mlt_log_info(NULL, " readNext tag: %s type = %d\n", _qxml_getthestring(reader.name()), reader.tokenType());
+		if (reader.isStartElement() && reader.name() == "trkpt")
```
ddennedy, line 780: There are very many compilation errors with Qt 6 like:
```
../src/modules/qt/gps_parser.cpp: In function ‘void qxml_parse_gpx(QXmlStreamReader&, gps_point_ll**, int*)’:
../src/modules/qt/gps_parser.cpp:780:51: error: conversion from ‘const char [6]’ to ‘QChar’ is ambiguous
  780 |   if (reader.isStartElement() && reader.name() == "trkpt")
      |                                                   ^~~~~~~
In file included from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstring.h:50,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qobject.h:47,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qiodevice.h:46,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qxmlstream.h:43,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/QXmlStreamReader:1,
                 from ../src/modules/qt/gps_parser.h:24,
                 from ../src/modules/qt/gps_parser.cpp:24:
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:126:45: note: candidate: ‘constexpr QChar::QChar(char)’ <near match>
  126 |     QT_ASCII_CAST_WARN constexpr Q_IMPLICIT QChar(char c) noexcept : ucs(uchar(c)) { }
      |                                             ^~~~~
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:126:45: note:   conversion of argument 1 would be ill-formed:
../src/modules/qt/gps_parser.cpp:780:51: error: invalid conversion from ‘const char*’ to ‘char’ [-fpermissive]
  780 |   if (reader.isStartElement() && reader.name() == "trkpt")
      |                                                   ^~~~~~~
      |                                                   |
      |                                                   const char*
In file included from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstring.h:50,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qobject.h:47,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qiodevice.h:46,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qxmlstream.h:43,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/QXmlStreamReader:1,
                 from ../src/modules/qt/gps_parser.h:24,
                 from ../src/modules/qt/gps_parser.cpp:24:
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:119:26: note: candidate: ‘constexpr QChar::QChar(char16_t)’ <near match>
  119 |     constexpr Q_IMPLICIT QChar(char16_t ch) noexcept : ucs(ch) {}
      |                          ^~~~~
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:119:26: note:   conversion of argument 1 would be ill-formed:
../src/modules/qt/gps_parser.cpp:780:51: error: invalid conversion from ‘const char*’ to ‘char16_t’ [-fpermissive]
  780 |   if (reader.isStartElement() && reader.name() == "trkpt")
      |                                                   ^~~~~~~
      |                                                   |
      |                                                   const char*
In file included from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstring.h:50,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qobject.h:47,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qiodevice.h:46,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qxmlstream.h:43,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/QXmlStreamReader:1,
                 from ../src/modules/qt/gps_parser.h:24,
                 from ../src/modules/qt/gps_parser.cpp:24:
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:114:26: note: candidate: ‘constexpr QChar::QChar(short int)’ <near match>
  114 |     constexpr Q_IMPLICIT QChar(short rc) noexcept : ucs(char16_t(rc)) {}
      |                          ^~~~~
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:114:26: note:   conversion of argument 1 would be ill-formed:
../src/modules/qt/gps_parser.cpp:780:51: error: invalid conversion from ‘const char*’ to ‘short int’ [-fpermissive]
  780 |   if (reader.isStartElement() && reader.name() == "trkpt")
      |                                                   ^~~~~~~
      |                                                   |
      |                                                   const char*
In file included from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstring.h:50,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qobject.h:47,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qiodevice.h:46,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qxmlstream.h:43,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/QXmlStreamReader:1,
                 from ../src/modules/qt/gps_parser.h:24,
                 from ../src/modules/qt/gps_parser.cpp:24:
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:112:26: note: candidate: ‘constexpr QChar::QChar(ushort)’ <near match>
  112 |     constexpr Q_IMPLICIT QChar(ushort rc) noexcept : ucs(rc) {}
      |                          ^~~~~
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:112:26: note:   conversion of argument 1 would be ill-formed:
../src/modules/qt/gps_parser.cpp:780:51: error: invalid conversion from ‘const char*’ to ‘ushort’ {aka ‘short unsigned int’} [-fpermissive]
  780 |   if (reader.isStartElement() && reader.name() == "trkpt")
      |                                                   ^~~~~~~
      |                                                   |
      |                                                   const char*
In file included from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qchar.h:666,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstring.h:50,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qobject.h:47,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qiodevice.h:46,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qxmlstream.h:43,
                 from /home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/QXmlStreamReader:1,
                 from ../src/modules/qt/gps_parser.h:24,
                 from ../src/modules/qt/gps_parser.cpp:24:
/home/ddennedy/Qt/6.2.4/gcc_64/include/QtCore/qstringview.h:425:51: note:   initializing argument 2 of ‘bool operator==(QStringView, QChar)’
  425 |     friend bool operator==(QStringView lhs, QChar rhs) noexcept { return lhs == QStringView(&rhs, 1); }
      |                                             ~~~~~~^~~
```
Yeah, I know, very ugly C++ compilation errors.

src/modules/qt/gps_parser.cpp
```diff
@@ -686,71 +773,68 @@ void xml_parse_gpx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *coun
...
-			else if (strncmp(_s(gps_val->name), "extensions", strlen("extensions")) == 0)
+			gps_point_raw crt_point = uninit_gps_raw_point;
+			
+			QXmlStreamAttributes attributes = reader.attributes();
+			if (attributes.hasAttribute("lat"))
+				crt_point.lat = attributes.value("lat").toDouble();
+			if (attributes.hasAttribute("lon"))
+				crt_point.lon = attributes.value("lon").toDouble();
+			
+			while (reader.readNext() && !(reader.name() == "trkpt" && reader.tokenType() == QXmlStreamReader::EndElement)) //until closing trkpt
```
ddennedy, line 790: compile error on Qt 6:
`../src/modules/qt/gps_parser.cpp:790:51: error: conversion from ‘const char [6]’ to ‘QChar’ is ambiguous`

src/modules/qt/gps_parser.cpp
```diff
@@ -686,71 +773,68 @@ void xml_parse_gpx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *coun
...
+				else if (reader.name() == "course")
+					crt_point.bearing = reader.readElementText().toDouble();
+				else if (reader.name() == "extensions")
+				{
+					reader.readNextStartElement();
+					if (reader.name() == "TrackPointExtension")
+					{
+						reader.readNextStartElement();
+						if (reader.name() == "hr")
+							crt_point.hr = reader.readElementText().toDouble();
```
ddennedy, line 811: All of these name comparisons fail to compile on Qt 6:
`../src/modules/qt/gps_parser.cpp:796:26: error: conversion from ‘const char [4]’ to ‘QChar’ is ambiguous`

src/modules/qt/gps_parser.cpp
```diff
@@ -767,156 +851,159 @@ void xml_parse_tcx(xmlNodeSetPtr found_nodes, gps_point_ll **gps_list, int *coun
...
-	for (i=0; i<found_nodes->nodeNr; i++)
+	while (!reader.atEnd() && !reader.hasError())
 	{
-		gps_point_raw crt_point = uninit_gps_raw_point;
-		crt_node = found_nodes->nodeTab[i];
-
-		for (gps_val = crt_node->children; gps_val; gps_val = gps_val->next)
+		reader.readNext();
+		// mlt_log_info(NULL, " readNextStartElement tag: %s \n", _qxml_getthestring(reader.name()));
+		if (reader.isStartElement() && reader.name() == "Trackpoint")
```
ddennedy, line 858: Qt6 `../src/modules/qt/gps_parser.cpp:858:51: error: invalid conversion from ‘const char*’ to ‘char16_t’ [-fpermissive]`

ddennedy (changes requested):

ddennedy:
> I tried to make qt6 work but even though I installed it with pacman it doesn't seem to be picked up by "cmake -G Ninja ."

You need to invoke cmake with MLT options `-DMOD_QT=OFF -DMOD_QT6=ON` and possibly to find Qt `-DCMAKE_PREFIX_PATH=...` to the Qt install prefix, for example, if you install the Qt 6 SDK from the Qt installer.

dany123:
> You need to invoke cmake with MLT options `-DMOD_QT=OFF -DMOD_QT6=ON` and possibly to find Qt `-DCMAKE_PREFIX_PATH=...` to the Qt install prefix, for example, if you install the Qt 6 SDK from the Qt installer.

This worked, thanks. But wow there's 10000+ lines of errors, half of them complaining about the != operator in an included header:

>C:/Qt/Tools/mingw810_64/lib/gcc/x86_64-w64-mingw32/8.1.0/include/c++/bits/fs_path.h:237:60: note:   'std::filesystem::__cxx11::path' is not derived from 'const std::optional<_Tp>'
    || (__p.has_root_name() && __p.root_name() != root_name()))

I'm gonna assume that this is just some bad linking and fix the stuff in my .cpp.

src/modules/qt/gps_drawing.cpp
dany123: I've only commented them out for now, I'm considering using them in a future update.

src/modules/qt/filter_gpsgraphic.yml
```diff
@@ -0,0 +1,469 @@
...
+copyright: Meltytech, LLC
+creator: Daniel F
+license: LGPLv2.1
+language: en
+tags:
+  - Video
+description: Overlay GPS-related graphics onto the video
+
+parameters:
+  - identifier: gps.file
```
ddennedy, line 15: Congratulations, it compiles fine on Qt 6 now!
Please change "gps.file" in both filters to "resource" as that is our standard property name for this kind of parameter, and the xml module and Shotcut already know how to handle this property name to make and handle a relative path. Since I did not catch this for gpstext already you should add it as an undocumented backup (not published in this YAML) for when "resource" returns NULL. Also, you should consider making this the filter's argument (`arg` in `filter_gpsgraphic_init()`). Finally, it is not indicated as required, but it seems like it is to be useful. Making it required and argument does not necessarily mean it needs to be supplied as an argument. Also, you do not need to enforce the requirement, but it just helps logically to understand. Basically, if a user tries to use the filter and does not set this parameter and cannot get it to do anything, `required: yes` is a strong hint that you need to supply it to have some effect.

ddennedy (changes requested):

src/modules/qt/filter_gpsgraphic.yml
dany123: > Since I did not catch this for gpstext already you should add it as an undocumented backup (not published in this YAML) for when "resource" returns NULL

Just to double check I understood this: you mean replace all references of gps.file with resource (including in .yml for gps_text) but keep it backwards compatible in the .cpp to also accept gps.file? 

> Also, you should consider making this the filter's argument (arg in filter_gpsgraphic_init()). 

I'm not sure how that arg is populated, do I need to do more than this line?
`	mlt_properties_set_string( properties, "resource", arg );`

ddennedy: > replace all references of gps.file with resource

Yes

> keep it backwards compatible in the .cpp to also accept gps.file?

Only for gpstext since that is the only one released so far. We do not want to break existing projects and scripts.

> do I need to do more than this line?

No, that works good even when NULL.

src/modules/qt/filter_gpstext.cpp
```diff
@@ -695,14 +553,15 @@ static mlt_frame filter_process( mlt_filter filter, mlt_frame frame )
...
 	//read and process file if needed
-	process_file(filter);
+	process_file(filter, frame);
+	process_filter_properties(filter, frame);
 
 	// Apply keyword substitution before passing the text to the filter.
-	char* result = calloc( 1, MAX_TEXT_LEN );
+	char* result = (char*)calloc( 1, MAX_TEXT_LEN );
 	substitute_keywords( filter, result, dynamic_text, frame );
 	mlt_properties_set_string( text_filter_properties, "argument", result );
```
bmatherly, line 566: I know I am late to the party here. But upon inspection, it appears that this filter will not support parallel processing. The "argument" property could be overwritten by subsequent calls to filter_process() before the image is actually processed in another thread.

The string should be copied and pushed on the frame to support parallel processing as is done in the text filter:
https://github.com/mltframework/mlt/blob/master/src/modules/plus/filter_text.c#L166

bmatherly, line 566: I know I am late to the party here. But upon inspection, it appears that this filter will not support parallel processing. The "argument" property could be overwritten by subsequent calls to filter_process() before the image is actually processed in another thread.

The string should be copied and pushed on the frame to support parallel processing as is done in the text filter:
https://github.com/mltframework/mlt/blob/master/src/modules/plus/filter_text.c#L166

dany123: I did notice gps text was slower than gps graphic on export but thought it was just text being inefficient in itself. I'll have a look.

dany123: I'm a bit confused on how to solve this. Do I need to move the processing part to a get_image implementation? If so, how would the `return mlt_filter_process( text_filter, frame );` work there? 
Do I need to replace this particular `mlt_properties_set_string( text_filter_properties, "argument", result );` with a `
	mlt_frame_push_service( frame, strdup( argument ) );` ? But how would the text_filter know that this is the `argument` and not some random data? 
Also shouldn't `mlt_properties text_filter_properties = mlt_frame_unique_properties` be thread safe? As in it appends the argument to a particular/correct frame only? Or does this not work when passing the text to filter_text?

bmatherly: I had to spend a little more time reminding myself how this works. Now I remember that we made a special provision for the text filters to save the string on the frame to support parallel processing when being used as an encapsulated filter. You are using it properly. Sorry for the false alarm.

PR #816, opened 2022-07-30
Avoid to load both qt modules

jlskuz:
In case there is the intention to load both qt modules, qt6 will be
blocked. Use MLT_REPOSITORY_DENY to block qt5 (mltqt) instead

### DIFF
--- a/src/framework/mlt_repository.c
+++ b/src/framework/mlt_repository.c
@@ -101,6 +101,24 @@ mlt_repository mlt_repository_init( const char *directory )
 	mlt_tokeniser tokeniser = mlt_tokeniser_init();
 	int dl_length = mlt_tokeniser_parse_new( tokeniser, getenv( "MLT_REPOSITORY_DENY" ), ":" );
 
+	// check if both qt5 and qt6 modules are available…
+	int qtmodules = 0;
+	for ( i = 0; i < count; i++ )
+	{
+		const char *object_name = mlt_properties_get_value( dir, i);
+		qtmodules += strncmp(object_name, "mltqt", strlen( "mltqt" ) );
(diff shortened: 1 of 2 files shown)


jlskuz:
This is relevant after #807 got merged

ddennedy:
It is unfortunately ugly to include so much module-aware code in the framework like this, but it seems necessary.

src/framework/mlt_repository.c
```diff
@@ -101,6 +101,24 @@ mlt_repository mlt_repository_init( const char *directory )
...
+	for ( i = 0; i < count; i++ )
+	{
+		const char *object_name = mlt_properties_get_value( dir, i);
+		qtmodules += strncmp(object_name, "mltqt", strlen( "mltqt" ) );
+		qtmodules += strncmp(object_name, "mltqt6", strlen( "mltqt6" ) );
+	}
+	// …and not blocked
+	for (int j = 0; j < dl_length; j++ )
+	{
+		char *denyfile = calloc( 1, strlen( directory ) + strlen( mlt_tokeniser_get_string( tokeniser, j ) ) + 2 );
```
ddennedy, line 115: This needs to be `+ 3` for the trailing null character because `+ 2` only accounts for the `/` and `.` characters in the format string.

ddennedy (changes requested):

src/framework/mlt_repository.c
jlskuz: I know why I usually develop in high level C++ :laughing:

jlskuz:
> It is unfortunately ugly to include so much module-aware code in the framework like this, but it seems necessary.

Yes, indeed feels indeed not very nice, but I don't see another solution either…

PR #821, opened 2022-09-16
Define image format in producer's get_frame method so we can possibly...

j-b-m:
Here is my proposal as discussed to define the image format in each producer's get_frame method, before attempting to fetch an image. This allows possible optimizations like I do in qtblend (requesting the best suited image format by default.

I tried to add the change in all possible producers. Some points to be discussed/reviewed:
**avformat** producer:
I included a minor refactoring to avoid code duplication
**color** producer:
There is some code duplication between get_frame and get_image. Also I set the default format to yuv422 when the color has an opaque alpha component and rgba otherwise, maybe that is not the best.
**blipflash and noise** producers
Handle both yuv/rgba cases themselves in get_image so I didn't set the format
**hold and consumer** producers
Does not seem possible / useful for these 2 as we don't necessarily know the format at that stage

Let me know ŵhat you think, in any case this leads to some nice performance improvement in qtblend transition.

### DIFF
--- a/src/framework/mlt_image.c
+++ b/src/framework/mlt_image.c
@@ -581,3 +581,20 @@ void mlt_image_format_planes( mlt_image_format format, int width, int height, vo
 		strides[3] = 0;
 	};
 }
+
+
+
+/** Check if the alpha channel of an rgba image is opaque
+ *
+ * \public \memberof mlt_image_s
+ * \param image the image buffer
+ * \param width width of the image in pixels
+ * \param height height of the image in pixels
+ */
+int mlt_image_rgba_opaque(uint8_t *image, int width, int height)
+{
+	for (int i = 0; i < width * height; ++i) {
+		if (image[4 * i + 3] != 0xff) return 0;
+	}
+	return 1;
+}
--- a/src/modules/core/producer_colour.c
+++ b/src/modules/core/producer_colour.c
@@ -78,13 +78,6 @@ static int producer_get_image( mlt_frame frame, uint8_t **buffer, mlt_image_form
 	mlt_image_format current_format = mlt_properties_get_int( producer_props, "_format" );
 
 	// Parse the colour
-	if ( now && strchr( now, '/' ) )
-	{
-		now = strdup( strrchr( now, '/' ) + 1 );
-		mlt_properties_set( producer_props, "resource", now );
-		free( now );
-		now = mlt_properties_get( producer_props, "resource" );
-	}
 	mlt_color color = mlt_properties_get_color( producer_props, "resource" );
 
 	if ( mlt_properties_get( producer_props, "mlt_image_format") )
--- a/src/modules/core/producer_colour.c
+++ b/src/modules/core/producer_colour.c
@@ -252,6 +252,17 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
 		if ( mlt_properties_get( producer_props, "colour" ) != NULL )
 			mlt_properties_set( producer_props, "resource", mlt_properties_get( producer_props, "colour" ) );
 		
+		char *colorstring = mlt_properties_get( producer_props, "resource" );
+		if ( colorstring && strchr( colorstring, '/' ) )
+		{
+			colorstring = strdup( strrchr( colorstring, '/' ) + 1 );
+			mlt_properties_set( producer_props, "resource", colorstring );
+			free( colorstring );
+		}
+		mlt_color color = mlt_properties_get_color( producer_props, "resource" );
+		// Inform framework of the default frame format for this producer
+		mlt_properties_set_int( properties, "format", color.a < 255 ? mlt_image_rgba : mlt_image_yuv422 );
+
 		// Push the get_image method
 		mlt_frame_push_service( *frame, producer );
 		mlt_frame_push_get_image( *frame, producer_get_image );
--- a/src/modules/core/producer_colour.c
+++ b/src/modules/core/producer_colour.c
@@ -251,7 +251,7 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
 		// colour is an alias for resource
 		if ( mlt_properties_get( producer_props, "colour" ) != NULL )
 			mlt_properties_set( producer_props, "resource", mlt_properties_get( producer_props, "colour" ) );
-		
+
 		char *colorstring = mlt_properties_get( producer_props, "resource" );
 		if ( colorstring && strchr( colorstring, '/' ) )
 		{
@@ -259,9 +259,17 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
 			mlt_properties_set( producer_props, "resource", colorstring );
 			free( colorstring );
 		}
-		mlt_color color = mlt_properties_get_color( producer_props, "resource" );
-		// Inform framework of the default frame format for this producer
-		mlt_properties_set_int( properties, "format", color.a < 255 ? mlt_image_rgba : mlt_image_yuv422 );
+
+		// Check if we have a predefined image format
+		if ( mlt_properties_exists( producer_props, "mlt_image_format") )
+		{
+			int image_format = mlt_image_format_id( mlt_properties_get( producer_props, "mlt_image_format") );
+			mlt_properties_set_int( properties, "format", image_format );
+		} else {
+			mlt_color color = mlt_properties_get_color( producer_props, "resource" );
+			// Inform framework of the default frame format for this producer
+			mlt_properties_set_int( properties, "format", color.a < 255 ? mlt_image_rgba : mlt_image_yuv422 );
+		}
 
 		// Push the get_image method
 		mlt_frame_push_service( *frame, producer );
--- a/src/modules/qt/transition_qtblend.cpp
+++ b/src/modules/qt/transition_qtblend.cpp
@@ -25,6 +25,16 @@
 #include <QPainter>
 #include <QTransform>
 
+static int is_opaque( uint8_t *image, int width, int height )
+{
+	int pixels = width * height + 1;
+	while ( --pixels ) {
+		if ( image[3] != 0xff ) return 0;
+		image += 4;
+	}
+	return 1;
+}
+
 static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *format, int *width, int *height, int writable )
 {
 	int error = 0;
@@ -56,10 +66,10 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
 	bool distort = mlt_properties_get_int( transition_properties, "distort" );
 
 	// Potention optimization if the producers do set their native format before fetching image
-	/*if (mlt_properties_get_int( b_properties, "format" ) == mlt_image_rgba) {
+	if (mlt_properties_get_int( b_properties, "format" ) == mlt_image_rgba) {
 		hasAlpha = true;
 		*format = mlt_image_rgba;
-	}*/
+	}
 
 	if ( b_height == 0  || (!distort && ( b_height < *height || b_width < *width) ) )
 	{
@@ -184,13 +194,19 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
 	// Check if we have transparency
 	int request_width = b_width;
 	int request_height = b_height;
+	bool imageFetched = false;
 	if ( !hasAlpha || *format == mlt_image_rgba )
 	{
 		// fetch image
 		error = mlt_frame_get_image( b_frame, &b_image, format, &b_width, &b_height, 0 );
-		if ( *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) )
+		bool imageFetched = true;
+		if (!hasAlpha && *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) )
 		{
 			hasAlpha = true;
+		}
+		if (hasAlpha && is_opaque( b_image, b_width, b_height ) )
+		{
+			hasAlpha = false;
 		}
 	}
 	if ( !hasAlpha )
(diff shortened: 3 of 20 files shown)


src/modules/core/producer_colour.c
```diff
@@ -252,6 +252,17 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
...
+		char *colorstring = mlt_properties_get( producer_props, "resource" );
+		if ( colorstring && strchr( colorstring, '/' ) )
+		{
+			colorstring = strdup( strrchr( colorstring, '/' ) + 1 );
+			mlt_properties_set( producer_props, "resource", colorstring );
+			free( colorstring );
+		}
+		mlt_color color = mlt_properties_get_color( producer_props, "resource" );
+		// Inform framework of the default frame format for this producer
+		mlt_properties_set_int( properties, "format", color.a < 255 ? mlt_image_rgba : mlt_image_yuv422 );
```
ddennedy, line 264: This does not factor in the `mlt_image_format` property on the color producer.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -25,6 +25,16 @@
 #include <QPainter>
 #include <QTransform>
 
+static int is_opaque( uint8_t *image, int width, int height )
```
ddennedy, line 28: Since this appears in `transition_frei0r.c` I think it should be refactored into the mlt_image class simply for convenience but not to require a `mlt_image` struct as that would be inefficient. Brian has a goal to include mlt_image pointers on a new get_image callback someday. So, we need to anticipate that someday there will be a `mlt_image_is_opaque()` that takes a `mlt_image` pointer. Thus, lets call this one `mlt_image_rgba_opaque()`.

Also, Brian and I recently learned that compilers can optimize a for loop with buffer array indexing better than a while loop with pointer arithmetic. So, we ought to refactor for that as well:
```
int mlt_image_rgba_opaque(uint8_t *image, int width, int height)
{
	for (int i = 0; i < width * height; ++i) {
		if (image[4 * i + 3] != 0xff) return 0;
	}
	return 1;
}
```
And do not forget to add it to `mlt.vers`.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -56,10 +66,10 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
 	bool distort = mlt_properties_get_int( transition_properties, "distort" );
 
 	// Potention optimization if the producers do set their native format before fetching image
-	/*if (mlt_properties_get_int( b_properties, "format" ) == mlt_image_rgba) {
```
ddennedy, line 59: What is the status of this commented-out code?

ddennedy (changes requested):

j-b-m:
On Freitag, 16. September 2022 22:55:44 CEST Dan Dennedy wrote:
> @ddennedy requested changes on this pull request.

Thanks for the review.

> > @@ -252,6 +252,17 @@ static int producer_get_frame( mlt_producer producer,
> > mlt_frame_ptr frame, int i
>  		if ( mlt_properties_get( producer_props, "colour" ) != 
NULL )
>  			mlt_properties_set( producer_props, 
"resource", mlt_properties_get(
> producer_props, "colour" ) );
> 
> +		char *colorstring = mlt_properties_get( producer_props, 
"resource" );
> +		if ( colorstring && strchr( colorstring, '/' ) )
> +		{
> +			colorstring = strdup( strrchr( colorstring, 
'/' ) + 1 );
> +			mlt_properties_set( producer_props, 
"resource", colorstring );
> +			free( colorstring );
> +		}
> +		mlt_color color = mlt_properties_get_color( 
producer_props, "resource" );
> +		// Inform framework of the default frame format for this 
producer
> +		mlt_properties_set_int( properties, "format", color.a < 
255 ?
> mlt_image_rgba : mlt_image_yuv422 );
> 
> This does not factor in the `mlt_image_format` property on the color
> producer.

Yes, I thought about it after my initial commit, will fix that.

> > @@ -25,6 +25,16 @@
> 
>  #include <QPainter>
>  #include <QTransform>
> 
> +static int is_opaque( uint8_t *image, int width, int height )
> 
> Since this appears in `transition_frei0r.c` I think it should be refactored
> into the mlt_image class simply for convenience but not to require a
> `mlt_image` struct as that would be inefficient. Brian has a goal to
> include mlt_image pointers on a new get_image callback someday. So, we need
> to anticipate that someday there will be a `mlt_image_is_opaque()` that
> takes a `mlt_image` pointer. Thus, lets call this one
> `mlt_image_rgba_opaque()`.
> 
> Also, Brian and I recently learned that compilers can optimize a for loop
> with buffer array indexing better than a while loop with pointer
> arithmetic. So, we ought to refactor for that as well: ```
> int mlt_image_rgba_opaque(uint8_t *image, int width, int height)
> {
> 	for (int i = 0; i < width * height; ++i) {
> 		if (image[4 * i + 3] != 0xff) return 0;
> 	}
> 	return 1;
> }
> ```
> And do not forget to add it to `mlt.vers`.

Ok, will do.

> > @@ -56,10 +66,10 @@ static int get_image( mlt_frame a_frame, uint8_t
> > **image, mlt_image_format *form
>  	bool distort = mlt_properties_get_int( transition_properties, 
"distort" );
> 
>  	// Potention optimization if the producers do set their native 
format
> before fetching image -	/*if (mlt_properties_get_int( b_properties,
> "format" ) == mlt_image_rgba) {
> 
> What is the status of this commented-out code?

This is the part that provides a nice speedup when compositing a producer that 
provides rgba image by default, I can change the comment to remove 
"Potential".

I will be on vacation until the 25th of september, will work on implementing 
your suggestions as soon as I am back.

Best regards,

Jean-Baptiste

bmatherly:
Do you think this documentation is still sufficient?
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_frame.h#L68
I wonder if we should add some additional text to explain that before an image is retrieved the property represents the native format of the producer.

Can you provide some melt command line examples that I can use to test the performance improvement before and after the change?

j-b-m:
So I have now updated my pull request taking Dan's comment into account. Here is a short overview of the performance gain and commands used on my system (10 years old intel i7 with 4 cores 8 threads). All tests were performed twice to get a mean time, using full HD sources (videos and images).

**Compositing an image with alpha over a YUV420p video:**
time melt sample_video.mp4 out=2000 -track picture.png out=2000 -transition qtblend -consumer avformat:test.mp4

qtblend (before the change):
2m00s

qtblend with my "format" optimization:
1m41s

frei0r.cairoblend:
1m40s

**Compositing an image without alpha (jpeg) over a YUV420p video:**
time melt sample_video.mp4 out=2000 -track picture.jpeg out=2000 -transition qtblend -consumer avformat:test.mp4

qtblend (before the change):
0m14s

qtblend with my "format" optimization:
0m14s

frei0r.cairoblend:
1m40s

**Compositing two YUV420p videos (without alpha):**
time melt sample_video.mp4 out=2000 -track melt sample_video.mp4 out=2000 -transition qtblend -consumer avformat:test.mp4

qtblend (before the change):
1m23s

qtblend with my "format" optimization:
1m23s

frei0r.cairoblend:
1m34s

src/modules/core/producer_colour.c
```diff
@@ -251,7 +251,26 @@ static int producer_get_frame( mlt_producer producer, mlt_frame_ptr frame, int i
...
 			mlt_properties_set( producer_props, "resource", mlt_properties_get( producer_props, "colour" ) );
-		
+
+		char *colorstring = mlt_properties_get( producer_props, "resource" );
+		if ( colorstring && strchr( colorstring, '/' ) )
+		{
+			colorstring = strdup( strrchr( colorstring, '/' ) + 1 );
+			mlt_properties_set( producer_props, "resource", colorstring );
+			free( colorstring );
+		}
```
ddennedy, line 261: Since we have this code here now, we no longer need it above in the get_image function.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -184,14 +184,20 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
...
 	int request_width = b_width;
 	int request_height = b_height;
+	bool imageFetched = false;
 	if ( !hasAlpha || *format == mlt_image_rgba )
 	{
 		// fetch image
 		error = mlt_frame_get_image( b_frame, &b_image, format, &b_width, &b_height, 0 );
-		if ( *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) )
+		bool imageFetched = true;
+		if (!hasAlpha && *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) )
```
ddennedy, line 193: This expression does not have enough parentheses for readability.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -184,14 +184,20 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
...
 		{
 			hasAlpha = true;
 		}
+		if ( hasAlpha && mlt_image_rgba_opaque( b_image, b_width, b_height ) )
+		{
+			hasAlpha = false;
+		}
```
ddennedy, line 200: Consider `hasAlpha = hasAlpha && mlt_image_rgba_opaque( b_image, b_width, b_height );`

ddennedy (changes requested):

src/modules/qt/transition_qtblend.cpp
```diff
@@ -184,14 +184,17 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
...
 	{
 		// fetch image
 		error = mlt_frame_get_image( b_frame, &b_image, format, &b_width, &b_height, 0 );
-		if ( *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) )
+		bool imageFetched = true;
+		if ( !hasAlpha && ( *format == mlt_image_rgba || mlt_frame_get_alpha( b_frame ) ) )
 		{
 			hasAlpha = true;
 		}
+		hasAlpha = hasAlpha && mlt_image_rgba_opaque( b_image, b_width, b_height );
```
bmatherly, line 197: Shouldn't this be "! mlt_image_rgba_opaque"?

src/modules/qt/transition_qtblend.cpp
```diff
@@ -210,18 +213,17 @@ static int get_image( mlt_frame a_frame, uint8_t **image, mlt_image_format *form
 			*height = b_height;
 		}
 		*image = b_image;
-		mlt_frame_replace_image( a_frame, b_image, *format, *width, *height );
+		// mlt_frame_replace_image( a_frame, b_image, *format, *width, *height );
```
bmatherly, line 216: If this is not needed, then delete it?

src/framework/mlt_image.c
```diff
@@ -581,3 +581,20 @@ void mlt_image_format_planes( mlt_image_format format, int width, int height, vo
...
+
+
+/** Check if the alpha channel of an rgba image is opaque
+ *
+ * \public \memberof mlt_image_s
+ * \param image the image buffer
+ * \param width width of the image in pixels
+ * \param height height of the image in pixels
+ */
+int mlt_image_rgba_opaque(uint8_t *image, int width, int height)
```
bmatherly, line 594: This function should operate on an mlt_image object. Here is a proposed refactor (NOT TESTED)!
https://github.com/bmatherly/mlt/commit/a408493198ed517c1d42e8536d4d0e0c287576f0

bmatherly (changes requested):

src/framework/mlt_image.c
ddennedy: I disagree. I do not want to require constructing a mlt_image object just to use this function. Look at where it came from and how it’s being used.

src/modules/qt/transition_qtblend.cpp
ddennedy: Yes

src/framework/mlt_image.c
bmatherly: Hmm. OK. I guess I do not feel too strongly about it. But if we do not write it to take an mlt_image struct, that means we will have to deprecate this new function eventually when the code is changed to use mlt_image and I will probably be the one to rewrite the new function. Did you view my proposed change? It is only 2 extra lines of code to create the mlt_image object and make the function future proof.

Also, this function would be the only function in mlt_image.h that operates on a raw buffer instead of an mlt_image struct. But I do not think of another better place to put it.

src/modules/qt/transition_qtblend.cpp
j-b-m: Thanks for noticing - I did this after waking up in the middle of the night, not the best moment for such stuff.

src/framework/mlt_image.c
ddennedy: I did consider the mlt_image-based approach before suggesting `mlt_image_rgba_opaque` to avoid a future name conflict. Then, there would not not necessarily be a need to break API. I am going to accept this PR as-is for JB's sake. Please refactor it afterward to your `mlt_image_is_opaque` and drop `mlt_image_rgba_opaque` before release, which will be by end of year at the latest but possibly as soon as end of October.

ddennedy (approved):

bmatherly (approved):

src/framework/mlt_image.c
bmatherly: > I am going to accept this PR as-is for JB's sake. Please refactor it afterward to your mlt_image_is_opaque and drop mlt_image_rgba_opaque before release, which will be by end of year at the latest but possibly as soon as end of October.

Agreed. Let's get this one merged.

ddennedy:
FYI @j-b-m I reverted the changes I made in `transition_frei0r.c` to take advantage of this new work due to this bug:  https://forum.shotcut.org/t/clip-transition-applies-to-lower-layers/34878
Basically `mlt_image_rgba_opaque()` is naive, and does not factor in transparent padding added in filter `resize`. I guess you can add more code (see `filter_resize.c`) to check if the image will be padded or not.

j-b-m:
@ddennedy thanks for letting me know. If I am not mistaken, we avoid the problem in qtblend by checking if the display aspect ratio of the b-frame is different from profile and enforce alpha in that case:
https://github.com/mltframework/mlt/blob/813fa681798c11a32229f5fbcc8127fca31aad73/src/modules/qt/transition_qtblend.cpp#L173

But I can make further tests...

PR #823, opened 2022-09-16
Allow outputing frame audio peak in dB

j-b-m:
in Kdenlive we use the "audiolevel" filter to get the peak audio level for each track, using the _audio_level.<N> property.
With this change, the filter can now optionally output the peak audio level in dB instead of a percentage, making it easier to use in vu-meter. No changes were done on the existing functionnalities

### DIFF
--- a/src/modules/normalize/filter_audiolevel.c
+++ b/src/modules/normalize/filter_audiolevel.c
@@ -61,7 +61,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 	mlt_properties filter_props = MLT_FILTER_PROPERTIES( filter );
 
 	int iec_scale = mlt_properties_get_int( filter_props, "iec_scale" );
-	int peak = mlt_properties_get_int( filter_props, "peak" );
+	int peak = mlt_properties_get_int( filter_props, "dbpeak" );
 	*format = mlt_audio_s16;
 	int error = mlt_frame_get_audio( frame, buffer, format, frequency, channels, samples );
 	if ( error || !buffer ) return error;
@@ -93,7 +93,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 			}
 			else
 			{
-				level = 20 * log10( (double)peakVal / (double)INT16_MAX );
+				level = AMPTODBFS( (double)peakVal / (double)INT16_MAX );
 			}
 		}
 		else
--- a/src/modules/normalize/filter_audiolevel.c
+++ b/src/modules/normalize/filter_audiolevel.c
@@ -61,7 +61,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 	mlt_properties filter_props = MLT_FILTER_PROPERTIES( filter );
 
 	int iec_scale = mlt_properties_get_int( filter_props, "iec_scale" );
-	int peak = mlt_properties_get_int( filter_props, "dbpeak" );
+	int dbPeak = mlt_properties_get_int( filter_props, "dbpeak" );
 	*format = mlt_audio_s16;
 	int error = mlt_frame_get_audio( frame, buffer, format, frequency, channels, samples );
 	if ( error || !buffer ) return error;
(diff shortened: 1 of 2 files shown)


src/modules/normalize/filter_audiolevel.yml
```diff
@@ -28,6 +28,14 @@ parameters:
     default: 1
     widget: checkbox
 
+  - identifier: peak
+    title: output samples peak value in dB instead of percentage in _audio_level.<N>
+    type: integer
+    minimum: 0
+    maximum: 1
```
ddennedy, line 35: Use `boolean` here and get rid of minimum and maximum. To do so, you need to change the `schema_version` above to "7.0". And please update the `iec_scale` parameter to match.

src/modules/normalize/filter_audiolevel.c
```diff
@@ -74,32 +75,54 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
...
+				level = -100;
+			}
 			else
-				num_oversample = 0;
-			// 10 samples @max => show max signal
-			if ( num_oversample > 10 )
 			{
-				level = 1.0;
-				break;
+				level = 20 * log10( (double)peakVal / (double)INT16_MAX );
```
ddennedy, line 96: Use `AMPTODBFS()` here.

src/modules/normalize/filter_audiolevel.c
```diff
@@ -61,6 +61,7 @@ static int filter_get_audio( mlt_frame frame, void **buffer, mlt_audio_format *f
 	mlt_properties filter_props = MLT_FILTER_PROPERTIES( filter );
 
 	int iec_scale = mlt_properties_get_int( filter_props, "iec_scale" );
+	int peak = mlt_properties_get_int( filter_props, "peak" );
```
ddennedy, line 64: Why "peak" instead of something about dbFS unit of measure? Also, the current code is already supposed to be like a peak meter when iec_scale=1 as described [here on wikipedia](https://en.wikipedia.org/wiki/Peak_programme_meter#IEC_60268-18_Digital_PPM). Looking at the new code below, this looks like a different, non-averaging implementation of peak. Is it true peak?

ddennedy (changes requested):

j-b-m:
On Freitag, 16. September 2022 21:42:26 CEST Dan Dennedy wrote:

Thanks for the review.

> @ddennedy requested changes on this pull request.
> 
> > +    type: integer
> 
> +    minimum: 0
> +    maximum: 1
> 
> Use `boolean` here and get rid of minimum and maximum. To do so, you need to
> change the `schema_version` above to "7.0". And please update the
> `iec_scale` parameter to match.

Done.
 			{
> 
> -				level = 1.0;
> -				break;
> +				level = 20 * log10( 
(double)peakVal / (double)INT16_MAX );
> 
> Use `AMPTODBFS()` here.

Done.
 
> > @@ -61,6 +61,7 @@ static int filter_get_audio( mlt_frame frame, void
> > **buffer, mlt_audio_format *f
>  	mlt_properties filter_props = MLT_FILTER_PROPERTIES( filter );
> 
>  	int iec_scale = mlt_properties_get_int( filter_props, "iec_scale" 
);
> +	int peak = mlt_properties_get_int( filter_props, "peak" );
> 
> Why "peak" instead of something about dbFS unit of measure? Also, the
> current code is already supposed to be like a peak meter when iec_scale=1
> as described [here on
> wikipedia](https://en.wikipedia.org/wiki/Peak_programme_meter#IEC_60268-18_
> Digital_PPM). Looking at the new code below, this looks like a different,
> non-averaging implementation of peak. Is it true peak?

Sure the parameter name was not the best. I now changed it to "dbpeak". And in 
fact, the code was borrowed from Shotcut's audio peak  scope:

https://github.com/mltframework/shotcut/blob/master/src/widgets/scopes/
audiopeakmeterscopewidget.cpp#L46

I am not an expert in audio scales, but when playing a standardized, stable 
sine test tone with 1 kHz and -9 dB, 
using iec_scale=1, the _audio_level property is set to 0.83
when using my updated code  (same as shotcut), the _audio_level property is 
set to -8.99.

See also the shotcut commit referencing issues in audiolevel filter:
https://github.com/mltframework/shotcut/commit/
f5928ed3dcdd57dcbb7f311cadfb1224ce0b034a

Best regards,

Jean-Baptiste

bmatherly:
What should the user expect if they set iec_scale=true and dbpeak=true? Will we keep adding more boolean parameters as other people request different calculations? I wonder if we should have one parameter with three options: percent, iec_scale, peak. We could deprecate the iec_scale parameter.

Also, what if someone wants both iec_scale and peak at the same time? Would it make sense to always calculate all three values? If you reformat the code to calculate all three values during one loop through the samples, the extra computation would be trivial. But we would have to decide what to name all three value properties on the frame. Maybe that is too complicated.

ddennedy:
> What should the user expect if they set iec_scale=true and dbpeak=true?

Do bear in mind that "true" and "false" are not accepted literal values for a a "boolean" parameter, and that `metaschema.yaml` says only 0 or 1. This is because the metadata is documentation oriented, but the implementation in mlt_properties uses int. In any case, for a logical reading in a comment it is fine. I just want to clear that up since I raised changing the metadata parameter type.

With that said, the question is interesting because you could make a very small change to this PR to support that combination.

ddennedy (changes requested):

bmatherly:
LGTM

PR #828, opened 2022-10-05
Make mlt_color animatable (keyframable)

jlskuz:


src/framework/mlt_property.c
```diff
@@ -821,7 +822,7 @@ char *mlt_property_get_string_l_tf( mlt_property self, locale_t locale, mlt_time
 		setlocale( LC_NUMERIC, localename );
 #endif // _WIN32
 
-		if ( self->types & mlt_prop_int )
+		if ( self->types & mlt_prop_int || self->types & mlt_prop_color ) // warning
```
bmatherly, line 825: Why do you have a warning here?

In the case of color, should it be converted to a hex string?

src/framework/mlt_property.c
```diff
@@ -937,7 +938,7 @@ void mlt_property_pass( mlt_property self, mlt_property that )
 
 	if ( self->types & mlt_prop_int64 )
 		self->prop_int64 = that->prop_int64;
-	else if ( self->types & mlt_prop_int )
+	else if ( self->types & mlt_prop_int || self->types & mlt_prop_color ) // warning
```
bmatherly, line 941: I also wonder why the warning is needed here.

src/framework/mlt_property.c
```diff
@@ -1279,7 +1281,50 @@ int mlt_property_interpolate( mlt_property self, mlt_property p[],
...
+	{
+		mlt_color value = { 0xff, 0xff, 0xff, 0xff };
+		if ( interp == mlt_keyframe_linear )
+		{
+			mlt_color colors[2];
+			mlt_color zero = { 0xff, 0xff, 0xff, 0xff };
+			colors[0] = p[1]? mlt_property_get_color( p[1], fps, locale ) : zero;
+			if ( p[2] )
+			{
+				colors[1] = mlt_property_get_color( p[2], fps, locale );
```
bmatherly, line 1294: This should work OK. But might surprise some people since it will not naturally go "around the rainbow". For example, to get from green to yellow, it might go through brown. I wonder if we should convert both colors to HSL, interpolate the HSL values, and then convert back to RGB. This is not a blocking comment from me - just posing it for discussion.

bmatherly (changes requested):
This is a fun idea. I think users will appreciate this.

src/framework/mlt_property.c
jlskuz: Well the warning comment doesn't belong here. I just added it for me during development and forgot to remove it. Solved.

Not sure about the hex string.

src/framework/mlt_property.c
jlskuz: See above

src/framework/mlt_property.c
jlskuz: Well first when I read your comment I thought it makes sense to use HSL, I haven't thought about this at all. However I checked in Blender and Glaxnimate how they interpolate colors and it seems they are both not going through the HSL approach.

src/framework/mlt_property.c
ddennedy: I cannot see the context for this anymore, but a hex string is not the preferred or typical format. CSS color string is.

src/framework/mlt_property.c
ddennedy: HSL might be nice for hue, but saturation and lightness will still not wrap. Performance will be worse too, but it is unclear the impact. But since Blender and Glaxnimate do it this way, it might be preferable.

bmatherly: I agree with this conclusion. Let's leave it as-is. Thanks for the discussion.

src/framework/mlt_property.c
```diff
@@ -821,7 +822,7 @@ char *mlt_property_get_string_l_tf( mlt_property self, locale_t locale, mlt_time
 		setlocale( LC_NUMERIC, localename );
 #endif // _WIN32
 
-		if ( self->types & mlt_prop_int )
+		if ( self->types & mlt_prop_int || self->types & mlt_prop_color )
```
bmatherly, line 825: Continuing the conversation from the previous review comment...

> ... a hex string is not the preferred or typical format. CSS color string is.

I agree with that comment. There are many places in the code that use hex strings which I think we could convert to CSS style in the future:
https://github.com/mltframework/mlt/blob/master/src/modules/qt/filter_audiospectrum.cpp#L349
But that is for another day.

For this line of code specifically, I think we should return a HTML/CSS color code if someone requests a string on a property that was previously set as a color type. I would like to see that change  before we merge so that behavior does not need to  be changed in the future and possibly break compatibility.

src/framework/mlt_property.c
bmatherly: Did you intend to remove this "//warning"?

src/mlt++/MltProperties.cpp
```diff
@@ -371,6 +371,16 @@ int Properties::set( const char *name, mlt_color value )
 	return mlt_properties_set_color( get_properties(), name, value );
 }
 
+mlt_color Properties::anim_get_color(const char *name, int position, int length)
```
bmatherly, line 374: I would really love to see unit tests for these functions here:
https://github.com/mltframework/mlt/blob/a7ecae88bf23014b9359ca9f792670efcf3d1942/src/tests/test_properties/test_properties.cpp#L165

But I am not sure if that should be required for this request. @jlskuz do you have any interest in trying to add unit tests? It is quite easy with the test framework we have. But if you are not interested, I can withdraw this comment.

bmatherly (changes requested):

src/framework/mlt_property.c
jlskuz: Okay, will look into this.

src/mlt++/MltProperties.cpp
jlskuz: I think it makes sense to have unit tests for this. Will try to write some.

jlskuz:
I just noticed that the smooth keyframe interpolation does not work properly in some cases, but could not see why at a first glance.

src/framework/mlt_property.c
```diff
@@ -1279,7 +1281,50 @@ int mlt_property_interpolate( mlt_property self, mlt_property p[],
...
+			colors[1] = p[1]? mlt_property_get_color( p[1], fps, locale ) : zero;
+			if ( p[2] )
+			{
+				colors[0] = p[0]? mlt_property_get_color( p[0], fps, locale ) : zero;
+				colors[2] = p[2]? mlt_property_get_color( p[2], fps, locale ) : zero;
+				colors[3] = p[3]? mlt_property_get_color( p[3], fps, locale ) : zero;
+				value.r = catmull_rom_interpolate( colors[0].r, colors[1].r, colors[2].r, colors[3].r, progress );
+				value.g = catmull_rom_interpolate( colors[0].g, colors[1].g, colors[2].g, colors[3].g, progress );
+				value.b = catmull_rom_interpolate( colors[0].b, colors[1].b, colors[2].b, colors[3].b, progress );
+				value.a = catmull_rom_interpolate( colors[0].a, colors[1].a, colors[2].a, colors[3].a, progress );
```
ddennedy, line 1318: You need to clamp these values to [0,255]. Drawing a curve through a set of points does not understand value ranges.

src/framework/mlt_property.c
jlskuz: Looking in to this again, I am really confused by the fact that MLT seems to use "wrong" CSS color strings in the pattern `#AARRGGBB` while correct CSS hex color strings have the alpha channel in the last position like `#RRGGBBAA`. Unfortunately I guess there is no way to change this behavior since it would break things, right?

jlskuz: Okay, https://www.educative.io/answers/what-is-the-difference-between-rgba-and-argb somehow explains it. However it is still confusing.

ddennedy: As best as I can recall at the time this was implemented CSS did not support an alpha component, and made  compatible with QColor: https://doc.qt.io/qt-6/qcolor.html

src/framework/mlt_property.c
jlskuz: removed

src/framework/mlt_property.c
jlskuz: Requesting a string gives you a `#AARRGGBB` string now, all covered by tests

src/mlt++/MltProperties.cpp
jlskuz: I think I covered everything related to animated colors with tests now.

src/framework/mlt_property.c
jlskuz: Done, and the issues I noticed seem to be fixed. Thanks for the hint.

jlskuz:
If we could have a release before November 3 (dependency freeze for our next Kdenlive version) this would be awesome, however it is not a hard requirement, so please don't feel to much pressure on this.

ddennedy:
I planned to make a release before merging this. I will try to do before Nov 3.

bmatherly (approved):

bmatherly:
I looked through this and I do not have any further comments. Thanks for adding the tests.

src/framework/mlt.vers
```diff
@@ -647,4 +647,10 @@ MLT_7.10.0 {
     mlt_image_fill_checkerboard;
     mlt_image_fill_white;
     mlt_image_rgba_opaque;
+    mlt_property_set_color;
+    mlt_property_get_color;
+    mlt_property_anim_set_color;
+    mlt_property_anim_get_color;
+    mlt_properties_anim_set_color;
+    mlt_properties_anim_get_color;
```
ddennedy, line 655: These symbols now need to moved to a new block with name `MLT_7.12.0`.

src/mlt++/mlt++.vers
```diff
@@ -717,3 +717,11 @@ MLT_7.6.0 {
       "Mlt::Properties::copy(Mlt::Properties&, char const*)";
     };
 } MLT_7.4.0;
+
+MLT_7.10.0 {
```
ddennedy, line 721: Change this symbols block name to `MLT_7.12.0`

ddennedy (changes requested):

src/framework/mlt.vers
jlskuz: Done

src/mlt++/mlt++.vers
jlskuz: Done

jlskuz:
Thanks for the release, BTW!

ddennedy (approved):

Issue #835, opened 2022-10-28
 Audio and video out of sync in avformat producer.


Adonis-Song:
Hi. 
I met some trouble when I use melt to export video.

My mlt version is v7.8.0. System version is Mac Big Sur 11.5.1 .

My melt command is
```
melt ~/Downloads/test.mp4 -consumer avformat:./result.mp4
```

But `result.mp4`  is out of sync.  

After using ffprobe to print info of the source video, I find the start time of video stream is not 0 and the start time of audio stream is 0.

And then I debug source code, find this code in function  `decode_audio`
```
	// If we're behind, ignore this packet
	// Skip this on non-seekable, audio-only inputs.
	if ( !discarded && pkt->pts >= 0 && ( self->seekable || self->video_format ) && *ignore == 0 && audio_used > samples / 2 )
	{
		int64_t pts = pkt->pts;
		if ( self->first_pts != AV_NOPTS_VALUE )
			pts -= self->first_pts;
		else if ( context->start_time != AV_NOPTS_VALUE && self->video_index != -1 )
			pts -= context->start_time;
		double timebase = av_q2d( context->streams[ index ]->time_base );
		int64_t int_position = llrint( timebase * pts * fps );
		int64_t req_position = llrint( timecode * fps );
		int64_t req_pts =      llrint( timecode / timebase );

```

but the timebase of `self->first_pts` is video stream's timebase , the timebase of `context->start_time` is AV_TIME_BASE_Q. So it should be as follows
 ```
	// If we're behind, ignore this packet
	// Skip this on non-seekable, audio-only inputs.
	if ( !discarded && pkt->pts >= 0 && ( self->seekable || self->video_format ) && *ignore == 0 && audio_used > samples / 2 )
	{
        double timebase = av_q2d( context->streams[ index ]->time_base );
        int64_t pts = pkt->pts;
		if ( self->first_pts != AV_NOPTS_VALUE )
			pts -= av_rescale_q(self->first_pts,
                                context->streams[self->video_index]->time_base,
                                context->streams[ index ]->time_base);
		else if ( context->start_time != AV_NOPTS_VALUE && self->video_index != -1 )
			pts -= av_rescale_q(context->start_time,
                                context->streams[self->video_index]->time_base,
                                AV_TIME_BASE_Q);
```
 
After changing to this code ,  my exported video is sync

Adonis-Song:
Does anyone know if my modification is correct? Should I make a PR? 
@bmatherly Do you have any advice?

[...]

PR #838, opened 2022-11-03
fix video and audio out of sync in avformat producer

Adonis-Song:
from issue : https://github.com/mltframework/mlt/issues/835#issue-1426892051


ddennedy:
I have to urgently make a new version 7.12 because there are some major bugs in the recently released 7.10. So this will not be included in the new release because it requires more testing including beyond my own. After merge, it should get into some Shotcut release for additional broad testing.

ddennedy:
Thanks, this passed my review and testing. I was just waiting for one more stable Shotcut release before merging. Ready now!

Issue #832, opened 2022-10-20
melt has memory leak if producer is xml


Adonis-Song:
Hi. 
My mlt version is v7.8.0. System version is Mac Big Sur 11.5.1 .

When I debug melt, I find producer would not be close if I pass a path of xml to melt.

This is my command:

```
./melt /Users/apple/test.xml
```

And this is xml:
```
<?xml version="1.0" encoding="utf-8"?>
<mlt LC_NUMERIC="zh_CN.UTF-8" version="7.8.0" root="/Users/apple/Documents/vizard/bpmlt/build/out/bin">
  <consumer f="mp4" movflags="+faststart" vcodec="libx264" crf="23" mlt_service="avformat" width="1920" height="1080" rescale="bilinear" preset="fast" acodec="aac" frame_rate_num="25" frame_rate_den="1" target="/Users/apple/Documents/mediaframework/profile-test/1/1-5-mlt.mp4"/>
  <producer id="producer0">
    <property name="resource">/Users/apple/Documents/mediaframework/profile-test/1/1.mp4</property>
    <property name="mlt_service">avformat</property>
  </producer>
  <playlist id="playlist0">
    <entry producer="producer0" in="00:00:00.000" out="00:01:48.990"/>
  </playlist>
  <tractor id="tractor0" in="" out="">
    <track producer="playlist0"/>
  </tractor>
</mlt>
```

![20221020-115027](https://user-images.githubusercontent.com/34226411/196852200-7171869b-694b-41ab-a182-319972190f58.jpeg)

When I debug this point, I find 'melt' would not be close ,because it's refs is two.

If I add these code to close it , melt whill be crashed. 

```
            int count = mlt_properties_ref_count( MLT_PRODUCER_PROPERTIES( melt ));
            while (count > 1)
                mlt_producer_close(melt);
            mlt_producer_close(melt);
```

So I donot know why the refs of `melt` is two and when I try to close it , the process of melt will be crashed .

bmatherly:
```
            int count = mlt_properties_ref_count( MLT_PRODUCER_PROPERTIES( melt ));
            while (count > 1)
                mlt_producer_close(melt);
            mlt_producer_close(melt);
```

This while loop is an infinite loop because count is never changed.

Adonis-Song:
o.o  
Sorry, I copy  wrong code. It should be as follow:
```
            while (mlt_properties_ref_count( MLT_PRODUCER_PROPERTIES( melt )) > 1)
[...]

PR #841, opened 2022-11-17
fix memory leak

Adonis-Song:
fix memory leak when xml file has 'consumer' tag but does not have 'profile' tag


Adonis-Song:
from issue https://github.com/mltframework/mlt/issues/832

ddennedy (commented):
This is what I was thinking, but I still need to test. Please fix the indentation in the meantime.

bmatherly:
This change looks good to me. I tested it and it seems to work.

PR #844, opened 2022-11-21
Add swresample link

bmatherly:
Problem:
When performing sample rate conversion, the resampling filters add a delay. Since filters can not "look ahead", the only option is to make up samples when prefilling the resampling buffer. These made up samples can cause an audible distortion or "tick".

Solution:
Add sample rate conversion links. The links can look ahead by reading the audio from the next frame. This allows the resampling buffer to be prefilled with real samples from the next frame and avoids the distortion/tick.

Test case
Download these two files:
https://drive.google.com/file/d/1hK_5nvlKgrwISsRxm9bSKnizOPYQVIZz/view?usp=share_link
https://drive.google.com/file/d/1uxNu40iCIu7Vr6OO-o9Xmr1_Bb6Ov0N_/view?usp=share_link
The timeline is a 41kHz audio clip with splits every 2 minutes. The clips are continuous (no frames have been removed)
Open the project and play the timeline. Observe an audible "click" at each split.
Next, open the project using Shotcut with this patch applied. Observe that there are not "clicks" at the splits.


bmatherly:
Based on my testing, this is ready to be merged when we are allowing major changes again.

src/mlt++/MltChain.cpp
```diff
@@ -27,10 +27,23 @@ Chain::Chain( ) :
...
+		id = NULL;
+	}
+
+	mlt_producer source = mlt_factory_producer( profile.get_profile(), id, service );
+	if ( source )
+	{
+		instance = mlt_chain_init( profile.get_profile() );
+		mlt_chain_set_source( instance, source );
+		if ( id == NULL )
+			mlt_chain_attach_normalisers( instance );
```
bmatherly, line 44: This might be worth a discussion. The loader producer does not return chains. So the normalizer link can not be added in the loader. I chose to add it here - mimicking the behavior of the loader. I am open to suggestions.

ddennedy: If a chain could be sub-classed and added to repository and factor ala a producer, would there be a need for a any other chain besides a loader chain that contains this logic?

src/framework/mlt_chain.c
```diff
@@ -393,6 +393,37 @@ void mlt_chain_close( mlt_chain self )
...
+		mlt_chain_base *base = self->local;
+		if ( base->link_count && mlt_properties_get_int( MLT_LINK_PROPERTIES( base->links[ i ] ), "_loader" ) )
+		{
+			mlt_log_warning( MLT_CHAIN_SERVICE(self), "Chain already has normalizers\n");
+			return;
+		}
+		mlt_link resample = mlt_factory_link( "swresample", NULL );
+		if ( !resample )
+		{
+			resample = mlt_factory_link( "resample", NULL );
```
ddennedy, line 416: I wonder if there should be an ini file for this. Could there ever be like like a speed link added for every chain in order to do a better trick playback? Or a rate-doubling deinterlace filter?

bmatherly: That is a good idea. I aspire to make a frame doubling deinterlacer.

src/mlt++/MltChain.cpp
bmatherly: A chain is a mechanism to connect links to a producer. So I can't think of a reason to make a subclass (except for the loader idea). Maybe the loader would be enough reason to make a registry for it.

Two other ideas:
* I could add a function to the factory which would add the normalizer logic - mlt_factory_chain(profile, service, resource)
* Somehow I could signal to the loader producer that it should encapsulate the producer in a chain. 

I will look into this some more.

ddennedy: I vote to skip a loader chain and add mlt_factory_chain. I was thinking maybe to add an option to melt to ask it to add chains to the producers.

bmatherly: In Melt, you can already add chains to producers by using the "-chain" option. But maybe you mean to add a chain to all producers without the need to specify the "-chain" option for each one?

ddennedy: Oh yeah, maybe a global option later, but I think the normalizing links should be done (indirectly) by melt in addition to xml to mlt++

bmatherly: I have most of this implemented (factory and ini file). However, I found an issue with this approach. If normalizing links are only added by the factory, then there is no way for an application to make a normalizing chain out of a producer that it already has created.

We do this in Shotcut here:
https://github.com/mltframework/shotcut/blob/287e3fae68c3509e3eddce934af91bcc3c9a3744/src/mltcontroller.cpp#L961

A couple of options:
1) Keep the mlt_chain_attach_normalisers() function so that applications can use it
2) Require the application to use the factory and then they will have to reapply the properties from the existing producer to the new chain
3) I could modify Shotcut so that all producers are always created as chains by the factory.

The advantage for #3 is that it would allow users to apply time remapping to image sequences and "xml as clips". Time remapping on "color" and "noise" clips probably does not make sense. But that probably doesn't matter. I am willing to do the work and debugging for this. But it would be a larger change with probably more  breakage than originally planned.

ddennedy: 1.) seems fine, and I suppose a corresponding `Mlt::Chain::attach_normalizers()` C++ method.
Please change spelling to be normal US English (z instead of s)? There is not any other symbols that use "normalise" that we need to be consistent. Charles Yates is Scottish and used all these British variations, and I have been slowly moving away from them (along with other abnormal things like code formatting).

frdbr: > The advantage for #3 is that it would allow users to apply time remapping to image sequences and "xml as clips". Time remapping on "color" and "noise" clips probably does not make sense. But that probably doesn't matter. I am willing to do the work and debugging for this. But it would be a larger change with probably more breakage than originally planned.

This is going to be very helpful when working with timelapses. Been wanting time remapping to images for a long time. Thanks!

bmatherly: I have added Mlt::Chain::attach_normalizers()

I also changed many instances of "noralise" to "normalize". I changed as many as I could find that would not break backwards compatability.

src/framework/mlt_chain.c
bmatherly: This is done now.

src/modules/core/chain_loader.ini
```diff
@@ -0,0 +1,9 @@
+# This file defines the chain normalizers, their fallbacks and the order they're attached
```
bmatherly, line 1: I put this file in "core" so that it can be by the loader.ini file for producers. I am open minded about the name and location.

Also, I notice that the existing loader.ini is used in two places:

https://github.com/mltframework/mlt/blob/master/src/modules/core/producer_loader.c#L221

https://github.com/mltframework/mlt/blob/0f14c1406e986900ec40de3a7b101333dcf5db8f/src/modules/core/consumer_multi.c#L123

If we are interested, I am willing to move this code into a common function mlt_producer_attach_normalizers() to be similar to the new chain function. This would be a project for a different pull request. Let me know if you are interested.

ddennedy:
Hmm now this is so many changes I’m losing motivation to review and merge it. I think you should not have changed all the cases of “normalise.” I simply did not want that in a new public symbol. Oh well.

src/modules/core/chain_loader.ini
ddennedy: Not interested

ddennedy: This is the wrong location because it is not used by the core module. It goes in the same location as metaschema.yaml

ddennedy:
I need a clear test case for this.

src/modules/normalize/filter_volume.c
```diff
@@ -413,15 +419,15 @@ static mlt_frame filter_process( mlt_filter filter, mlt_frame frame )
 		{
 			amplitude *= mlt_filter_get_progress( filter, frame );
 		}
-		mlt_properties_set_int( instance_props, "normalise", 1 );
+		mlt_properties_set_int( instance_props, "normalize", 1 );
```
ddennedy, line 422: I think we need to set both of these properties, or I would prefer to revert the property name change. It is not important enough to change it IMO.

bmatherly: Done

src/modules/core/chain_loader.ini
bmatherly: Moved to framework and same location as metaschema.yaml.

bmatherly:
> I need a clear test case for this.

I updated the pull request description with a simple test case.

src/framework/mlt_chain.c
```diff
@@ -393,6 +394,56 @@ void mlt_chain_close( mlt_chain self )
...
+
+	static mlt_properties normalisers = NULL;
+
+	mlt_tokeniser tokenizer = mlt_tokeniser_init( );
+
+	// We only need to load the normalising properties once
+	if ( normalisers == NULL )
+	{
+		char temp[ 1024 ];
+		sprintf( temp, "%s/chain_normalizers.ini", mlt_environment( "MLT_DATA" ) );
```
ddennedy, line 421: Please use `PATH_MAX` instead of hardcoded 1024, and `snprintf()`. I realize that is not being used everywhere, in particular the loader producer, and I will make a separate change to address that.

src/framework/mlt_chain.c
```diff
@@ -393,6 +394,56 @@ void mlt_chain_close( mlt_chain self )
...
+		char *value = mlt_properties_get_value( normalisers, i );
+		mlt_tokeniser_parse_new( tokenizer, value, "," );
+		for ( int j = 0; j < mlt_tokeniser_count( tokenizer ); j ++ )
+		{
+			mlt_link link = mlt_factory_link( mlt_tokeniser_get_string( tokenizer, j ), NULL );
+			if ( link )
+			{
+				mlt_properties_set_int( MLT_LINK_PROPERTIES(link), "_loader", 1 );
+				mlt_chain_attach( self, link );
+				mlt_chain_move_link( self, mlt_chain_link_count( self ) - 1, 0 );
```
ddennedy, line 438: Please be aware that this move link call means that links will be added in the reverse order from the ini file. That is not a big deal and never may be. Feel free to ignore it for now, but FYI.

src/modules/avformat/link_swresample.yml
```diff
@@ -0,0 +1,17 @@
...
+license: LGPLv2.1
+language: en
+url: http://www.ffmpeg.org/
+tags:
+  - Audio
+  - Hidden
+description: Converts the audio sample rate.
+notes: >
+  This is not intended to be created directly. Rather, the loader producer
+  loads it if it is available to set the audio resampler to normalize inputs
```
ddennedy, line 16: It is a chain rather than the loader producer.

src/modules/core/producer_melt.c
```diff
@@ -56,7 +56,7 @@ mlt_producer producer_melt_file_init( mlt_profile profile, mlt_service_type type
 	if ( result != NULL )
 	{
 		mlt_properties_set( MLT_PRODUCER_PROPERTIES( result ), "resource", file );
-		mlt_properties_set_int( MLT_PRODUCER_PROPERTIES( result ), "loader_normalised", 1);
+		mlt_properties_set_int( MLT_PRODUCER_PROPERTIES( result ), "loader_normalized", 1);
```
ddennedy, line 59: producer melt's "-chain" handler is not calling `mlt_chain_attach_normalizers()`.

src/modules/resample/link_resample.yml
```diff
@@ -0,0 +1,17 @@
...
+language: en
+url: http://www.mega-nerd.com/SRC/
+tags:
+  - Audio
+  - Hidden
+description: >
+  Adjust an audio stream's sampling rate.
+  
+  This link can be added to a chain to normalize audio from the producer to
+  provide the rate requested by the consumer.
```
ddennedy, line 17: This part should be worded similar to `link_swresample.yml`.

ddennedy (changes requested):

src/framework/mlt_chain.c
bmatherly: done

src/framework/mlt_chain.c
bmatherly: Good comment. Noted.

src/modules/avformat/link_swresample.yml
bmatherly: updated to match link_resample

src/modules/core/producer_melt.c
bmatherly: Done

src/modules/resample/link_resample.yml
bmatherly: Done

PR #848, opened 2022-12-08
Fall back to r_frame_rate when avg_frame_rate is 0/0

sirf:
`avg_frame_rate` is sometimes `0/0` in some video streams, particularly in mxf files. Not sure why.

This change will grab the frame rate from `r_frame_rate` when `avg_frame_rate` is not a finite number.
It is theoretically possible that `r_frame_rate` will also be `0/0` for some file, but with this change the likelihood of obtaining a correct value increases, and with every file I've seen so far `r_frame_rate` has had the correct value when `avg_frame_rate` was `0/0`.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -403,7 +403,8 @@ static mlt_properties find_default_streams( producer_avformat self )
 				}
 				mlt_properties_set( meta_media, key, "video" );
 				snprintf( key, sizeof(key), "meta.media.%u.stream.frame_rate", i );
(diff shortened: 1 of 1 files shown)


ddennedy:
Your change only affects the reporting metadata and does not affect the operational aspect of frame rate in the avformat producer. But the idea is sound. Search for other uses of `avg_frame_rate`. The most important usage is around line 2231. However, it does check if the conversion to double is 0 or not-a-number and use `r_frame_rate` instead. How is this change helping you?

sirf:
I present the meta information to the user. Most stuff under `meta.media.%d.stream...` is shown to the user nicely formatted for each stream, and the frame rate issue was bothering me so I thought I'd make an attempt to fix it.

PR #859, opened 2023-01-12
Add filtergraph support for avformat producer

DerFetzer:
Currently I am working with footage in S-Log3 (10bit/yuv422p10le).

My workflow has been to apply any corrections that need or benefit from 10bit (e.g. exposure, color space transform, LUT) during transcoding with ffmpeg to 8bit and then to work with the resulting files in kdenlive.

To get rid of the overhead of transcoding all files before using them I implemented a new property 'filtergraph' on the avformat producer where you can input a libavfilter filter graph string as in ffmpeg's '-vf' parameter.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -875,10 +875,6 @@ static int setup_filters(producer_avformat self)
 				inputs->next = NULL;
 
 				if (!error) error = (avfilter_graph_parse(self->vfilter_graph, self->filtergraph, inputs, outputs, NULL) < 0);
-				if (error) {
-					avfilter_inout_free(&outputs);
-					avfilter_inout_free(&inputs);
-				}
 		}
 		if (self->vfilter_graph) {
 			if (!error && !self->filtergraph) error = ( avfilter_link(self->vfilter_in, 0, last_filter, 0) < 0 );
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -132,7 +132,6 @@ struct producer_avformat_s
 	int autorotate;
 	int is_audio_synchronizing;
 	int video_send_result;
-	char* filtergraph;
 #if USE_HWACCEL
 	struct {
 		int pix_fmt;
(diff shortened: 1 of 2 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -830,32 +831,53 @@ static int insert_filter(AVFilterGraph *graph, AVFilterContext **last_filter, co
...
 {
 	int error = 0;
 
-	if (!self->vfilter_graph && self->autorotate && self->video_index != -1) {
-		mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
-		double theta  = get_rotation(properties, self->video_format->streams[self->video_index]);
+	if (!self->vfilter_graph && (self->autorotate || self->filtergraph) && self->video_index != -1) {
+		AVFilterContext *last_filter = NULL;
+		if (self->autorotate) {
+				mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
```
ddennedy, line 841: Thanks for your contribution! Too much indentation here.

src/modules/avformat/producer_avformat.c
```diff
@@ -1019,7 +1041,8 @@ static int producer_open(producer_avformat self, mlt_profile profile, const char
 #ifdef AVFILTER
 				if (!test_open) {
 					self->autorotate = !mlt_properties_get(properties, "autorotate") || mlt_properties_get_int(properties, "autorotate");
-					error = setup_autorotate_filters(self);
+					self->filtergraph = mlt_properties_get(properties, "filtergraph");
```
ddennedy, line 1044: Why are you saving this value instead of simply getting the property when needed? For example, wherever you access `self->filtergraph` instead
```
const char *filtergraph = mlt_properties_get(properties, "filtergraph");
if ((self->autorotate  || self->filtergraph)...
```
below, or
`if (!error) error = (avfilter_graph_parse(self->vfilter_graph, filtergraph, inputs, outputs, NULL) < 0);`
above.

src/modules/avformat/producer_avformat.yml
```diff
@@ -246,3 +246,8 @@ parameters:
       multiple of 90 degrees.
     type: integer
     unit: degrees
+
+  - identifier: filtergraph
+    title: Filtergraph
+    description: >
+      Filtergraph to apply to resource. Uses libavfilter syntax.
```
ddennedy, line 253: This needs a `type: string` and `description` is short enough to include on the same line.

ddennedy (changes requested):

bmatherly:
This is a neat idea. But I wonder how you will expose this to the user in KDENLIVE. I guess the application will generate the filtergraph string from the user settings? And then parse the filtergraph string to display the settings back to the user?

We are close to being able to do this natively since we added YUV422P16
https://github.com/mltframework/mlt/commit/039af9ba2a50d60e980e9aa63a8249612f2c7a3f

src/modules/avformat/producer_avformat.c
DerFetzer: I just wanted to mimic the behavior of dealing with autorotate but of course I can change it to a direct property access.

DerFetzer:
I am not very familiar with the Kdenlive codebase but this could be a viable solution.
In the first step it could even be sufficient to just give the user a non-validating text input where they input the filtergraph directly.

DerFetzer:
> We are close to being able to do this natively since we added YUV422P16 [039af9b](https://github.com/mltframework/mlt/commit/039af9ba2a50d60e980e9aa63a8249612f2c7a3f)

What is missing to make this possible? My solution was the quickest idea I could come up with to solve my current problem.

bmatherly:
>> We are close to being able to do this natively since we added YUV422P16 [039af9b](https://github.com/mltframework/mlt/commit/039af9ba2a50d60e980e9aa63a8249612f2c7a3f)

> What is missing to make this possible?

We have an MLT filter that can provide most [FFMpeg filters](https://github.com/mltframework/mlt/blob/master/src/modules/avformat/filter_avfilter.c)
So the user would just use those like normal in the UI. Then, we would need some way to tell the filters to request images in YUV422P16 format rather than the consumer requested format. That could be added with a property.

> In the first step it could even be sufficient to just give the user a non-validating text input where they input the filtergraph directly.

I see what you mean. That might be good for a developer. But not very friendly for a normal user. That is why I am trying to think of a solution that could use the standard interface of filters. But I do not mean to distract from this pull request too much.

DerFetzer:
> > > We are close to being able to do this natively since we added YUV422P16 [039af9b](https://github.com/mltframework/mlt/commit/039af9ba2a50d60e980e9aa63a8249612f2c7a3f)
> 
> > What is missing to make this possible?
> 
> We have an MLT filter that can provide most [FFMpeg filters](https://github.com/mltframework/mlt/blob/master/src/modules/avformat/filter_avfilter.c) So the user would just use those like normal in the UI. Then, we would need some way to tell the filters to request images in YUV422P16 format rather than the consumer requested format. That could be added with a property.
> 
Thank you very much for the explanation! I'll have a look at the filter source to understand it better.
> > In the first step it could even be sufficient to just give the user a non-validating text input where they input the filtergraph directly.
> 
> I see what you mean. That might be good for a developer. But not very friendly for a normal user. That is why I am trying to think of a solution that could use the standard interface of filters.

Yes, I totally agree with you.

ddennedy:
Something to consider is that the current approach probably lets one use libavfilter audio and video sources whereas the MLT avfilter will not. So, perhaps both can be done. 

A problem with the MLT avfilter approach is that nothing prevents one from using a non-avfilter between avfilters, and no other filters and transitions support `mlt_image_yuv422p16` resulting in a lossy conversion to something else. So, it will create a user support problem if the tool does not provide some way to limit how filters are used and in which order. I imagine it would be easy to add a new pre-processing field to the Kdenlive Properties panel possibly with some configurable presets for the filtergaph. Of course, that would not be a very friendly way to provide color and exposure adjustment prior to conversion to 8-bit Rec. 709.

src/modules/avformat/producer_avformat.c
ddennedy: I see. Upon reviewing `self->autorotate` I probably did it this because it was not a simple get property:
`self->autorotate = !mlt_properties_get(properties, "autorotate") || mlt_properties_get_int(properties, "autorotate");`
And then used from 4 different functions.

ddennedy (approved):

bmatherly:
> So, it will create a user support problem if the tool does not provide some way to limit how filters are used and in which order.

Yes. That is a good point. I suppose it could be a paradigm similar to how Shotcut requires the GPU filters to be applied first (when GPU is enabled).

PR #865, opened 2023-01-27
producer_avformat: Use filtergraph dimensions and pixel format where appropriate

Tjoppen:
This is necessary since avfilter may change dimensions and/or pixel format.
Autorotate is handled more succinctly in set_image_size().


Tjoppen:
Worth noting is that producer_avformat still uselessly calls sliced_h_pix_fmt_conv_proc() even when the frame is in the correct format. Avoiding the resulting memcpy()s would require having lavfi render into the buffer supplied to producer_get_image()

ddennedy:
Thanks, this fixes your filtergraph example using scale and format. However, it is not faster because

>  Avoiding the resulting memcpy()s would require having lavfi render into the buffer supplied to producer_get_image()

FFmpeg and MLT image buffers are not organized in the same manner. So, you need to bring the scaling into avformat producer, and use filtergraph to do things before scaling and pixfmt/colorspace conversion.

PR #869, opened 2023-02-05
fix mlt_chain inits movit on wrong thread

ddennedy:
OpenGL has thread affinity, and mlt_chain is initializing Movit on the wrong thread because it calls `mlt_profile_from_producer`, which calls `mlt_frame_get_image`, which invokes the Movit OpenGL. Rather, the mlt_consumer read ahead (render) thread must be the only thread to call `mlt_frame_get_image`. This is documented [here](https://www.mltframework.org/docs/opengl/).   `mlt_profile_from_producer` calls `mlt_frame_get_image` because only certain information such as interlace/progressive is reliable after decoding a frame. However, mlt_chain only seems to be concerned about frame rate, and that is obtained with avformat simply by getting a frame.
"meta.media.frame_rate_num" and "meta.media.frame_rate_den" are conventions supported by producers that have frame rates - decklink and avformat currently.
What are the consequences of the `mlt_chain_base.source_profile` resolution and aspect ratio based on the composition profile rather than source?

### DIFF
--- a/src/framework/mlt_chain.c
+++ b/src/framework/mlt_chain.c
@@ -84,7 +84,7 @@ mlt_chain mlt_chain_init( mlt_profile profile )
 			// Generate local space
 			self->local = calloc( 1, sizeof( mlt_chain_base ) );
 			mlt_chain_base* base = self->local;
-			base->source_profile = mlt_profile_clone(profile);
+			base->source_profile = NULL;
 
 			// Listen to property changes to pass along to the source
 			mlt_events_listen( MLT_CHAIN_PROPERTIES(self), self, "property-changed", ( mlt_listener )chain_property_changed );
@@ -125,6 +125,7 @@ void mlt_chain_set_source( mlt_chain self, mlt_producer source )
 		mlt_properties_inc_ref( source_properties );
 
 		// Save the native source producer frame rate
+		base->source_profile = mlt_profile_clone(mlt_service_profile(MLT_CHAIN_SERVICE(self)));
 		mlt_frame frame = NULL;
 		mlt_service_get_frame(MLT_PRODUCER_SERVICE(source), &frame, 0);
 		mlt_frame_close(frame);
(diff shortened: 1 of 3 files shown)


src/framework/mlt_chain.c
```diff
@@ -124,9 +124,16 @@ void mlt_chain_set_source( mlt_chain self, mlt_producer source )
...
-		base->source_profile = mlt_profile_init(NULL);
-		mlt_profile_from_producer( base->source_profile, base->source );
+		// Save the native source producer frame rate
+		mlt_frame frame = NULL;
+		mlt_service_get_frame(MLT_PRODUCER_SERVICE(source), &frame, 0);
+		mlt_frame_close(frame);
+		if (mlt_properties_get_int(source_properties, "meta.media.frame_rate_num") > 0 &&
+		    mlt_properties_get_int(source_properties, "meta.media.frame_rate_den") > 0)
+		{
+			base->source_profile->frame_rate_num = mlt_properties_get_int(source_properties, "meta.media.frame_rate_num");
```
bmatherly, line 134: Invalid pointer. This profile was closed on line 121.

src/framework/mlt_chain.c
```diff
@@ -84,7 +84,7 @@ mlt_chain mlt_chain_init( mlt_profile profile )
 			// Generate local space
 			self->local = calloc( 1, sizeof( mlt_chain_base ) );
 			mlt_chain_base* base = self->local;
-			base->source_profile = NULL;
+			base->source_profile = mlt_profile_clone(profile);
```
bmatherly, line 87: I suggest to leave this NULL and move the clone down to be after line 121.

bmatherly (changes requested):
Seems like this is worth a try. 

Similar to what we do in timewarp:
https://github.com/mltframework/mlt/blob/9c3bb7d997a78fed3142c2ef95988623d21dce15/src/modules/core/producer_timewarp.c#L247

I recommend that you test with a timeremap filter (link) since that is currently the only scenario that uses the source profile.

bmatherly (approved):

Issue #870, opened 2023-02-08
avformat_producer: demuxing and decoding should be pipelined


Tjoppen:
Currently demuxing and decoding is done in the same thread in producer_avformat.c. This prevents playing files that have very high bitrate, since the demuxer cannot work while the decoder is doings its thing. The samples at https://dpel.aswf.io/ are examples of files that trigger this problem, for example ASC_StEM2_178_UHD_ST2084_1000nits_Rec2020_Stereo_ProRes4444XQ.mov which is just above 1 Gbit/s. Disk access thus cuts into decoding time, preventing realtime playback in many cases where it would be otherwise possible. On one machine ffmpeg is able to decode this file at ~1.4x realtime whereas melt only manages 0.95x realtime.

I gave real_time=2 on the consumer a try but this didn't improve the situation, less so even higher values. buffer=10 also didn't help. Perhaps I'm missing something. Simplest command that displays this:

```bash
$ time melt -producer avformat:$HOME/samples/dpel.aswf.io/ASC_StEM2_178_UHD_ST2084_1000nits_Rec2020_Stereo_ProRes4444XQ.mov in=1 out=2400 -consumer avformat:/dev/null f=null real_time=2
[snip]
real    1m45.118s
user    13m8.348s
sys     1m35.049s
```

Obviously hard to test without access to the same machine but the general idea should be clear: it takes more than 100s to decode 100s of essence that ffmpeg manages to decode in 70s. For reference just reading 100s from disk with dd takes 45s (bs=10M).

Tjoppen:
I've started implementing this using pthreads and conditions. I'll send a PR if it pans out. I also discovered that the producer doesn't make use of AVDISCARD_ALL where appropriate.

ddennedy:
fixed in 51e0563ec4af199be12f68d3b32700912df8dac7

PR #874, opened 2023-02-13
Threaded avformat producer

Tjoppen:
This brings melt quite a bit closer to ffmpeg in terms of wall time for decoding 4k ProRes on an AWS c5d.18xlarge with network attached storage (EBS). From 90 to 77 seconds for a 100s sample, where ffmpeg takes 67 seconds.

This could maybe be improved to do similar threading for audio, with the AVDISCARD_ALL branch this is unlikely to improve things with seekable sources.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -121,6 +121,11 @@ struct producer_avformat_s
 	pthread_mutex_t packets_mutex;
 	pthread_mutex_t open_mutex;
 	int is_mutex_init;
+	pthread_t packets_thread;
+	pthread_cond_t packets_cond;
+	int packets_thread_ret;    // latest non-zero non-EGAIN return on av_read_frame() in packets_thread
+	int packets_thread_stop;   // non-zero when packets_thread is to stop
+	int is_thread_init;
 	AVRational video_time_base;
 	mlt_frame last_good_frame; // for video error concealment
 	int last_good_position;    // for video error concealment
@@ -1272,6 +1277,19 @@ static int seek_video( producer_avformat self, mlt_position position,
 			avcodec_flush_buffers( self->video_codec );
 			self->video_send_result = 0;
 
+			// let packets_worker know we handled EOF
+			if ( self->packets_thread_ret == AVERROR_EOF ) {
+				self->packets_thread_ret = 0;
+			}
+
+			// empty vpackets
+			while ( mlt_deque_count( self->vpackets ) > 0 ) {
+				AVPacket *tmp = (AVPacket*) mlt_deque_pop_front( self->vpackets );
+				av_packet_free( &tmp );
+			}
+
+			pthread_cond_signal( &self->packets_cond );
+
 			// Remove the cached info relating to the previous position
 			self->current_position = POSITION_INVALID;
 			self->last_position = POSITION_INVALID;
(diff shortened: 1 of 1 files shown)


Tjoppen:
Related to #870

ddennedy:
Thanks for tackling this. I started testing this by seeing how it might improve the start of playback on my high res/bitrate GoPro videos (4K60, 10-bit, 120 Mb/s). Often for me, these stall a few seconds near the beginning. With this branch, the same video in a fresh app session seems to have started playback slightly less choppy; it was difficult to tell. But near the end of the short clip (but not yet the end), it went into an infinite loop of
`av_read_frame returned error -541478725 inside packets_worker`.

So, I created a test project to set an out point on the clip a few seconds before the end to continue comparing the performance and user experience in Shotcut. The `mlt_consumer` reports roughly the same amount of  dropped video frames (using real_time=1) than master. At the end of playback of the trimmed clip, Shotcut does a seek to the out point; upon that, that this branch produces a garbled video frame with errors:
```
[hevc @ 0x556973d5d800] Could not find ref with POC 25
[hevc @ 0x556973d5d800] Could not find ref with POC 2
```
image

This was reproducible - even after changing the out point and trying other files. These behaviors are not normal and reproducible with master branch. 
The loop problem was also present in some iPhone HEVC and H.264 videos, and it was reproduced with `melt -verbose <filename>` (using default sdl2 consumer). I recommend doing some real world, interactive testing and not only simple decode or transcode tests.

src/modules/avformat/producer_avformat.c
```diff
@@ -1658,6 +1681,53 @@ static int ignore_send_packet_result(int result)
...
+				}
+			} else {
+				mlt_log_verbose( MLT_PRODUCER_SERVICE(self->parent), "av_read_frame returned error %d inside packets_worker\n", ret );
+			}
+
+			pthread_cond_signal( &self->packets_cond );
+		}
+
+		av_packet_unref( pkt );
+		pthread_mutex_unlock( &self->packets_mutex );
```
bmatherly, line 1727: Micro-optimization: the mutex is unlocked here only to be immediately locked at the top of the loop. Suggest to move the lock/unlock outside of the loop to avoid this.

src/modules/avformat/producer_avformat.c
```diff
@@ -1658,6 +1681,53 @@ static int ignore_send_packet_result(int result)
...
+	AVPacket *pkt = av_packet_alloc();
+	if ( !pkt ) {
+		mlt_log_error( MLT_PRODUCER_SERVICE(self->parent), "av_packet_alloc failed\n" );
+		exit( EXIT_FAILURE );
+	}
+
+	for (;;) {
+		pthread_mutex_lock( &self->packets_mutex );
+
+	check_stop:
```
bmatherly, line 1696: I am not sure, but I expect the packets worker will also need to handle seek so that seeking does not occur during reading. Maybe add a "check_seek" step in this loop?

Tjoppen:
@ddennedy Actually we've done realtime tests including seeking, but only within what is essentially just the one clip. Does the problem also happen in kdenlive?

src/modules/avformat/producer_avformat.c
Tjoppen: Good idea. pthread_cond_wait() will unlock the mutex when appropriate, and it will be finally unlocked on return. I'll make the change in a second version of this PR.

src/modules/avformat/producer_avformat.c
Tjoppen: packets_mutex is locked prior to seeks, unless there is some other seeking logic I missed.

Tjoppen:
Updated the patchset. I messed around in melt with -consumer sdl2 and it seems to behave the same as in master now, including giving the most recent frame when seeking to the end of the clip rather than the last frame in the clip.

ddennedy:
The looping problem is fixed for me now too, but I still get the garbled frame often times simply pausing playback within Shotcut (it also does a seek to the most recent frame displayed). I also saw some in the middle of playback. These are 1 sec. fixed+closed GOP GoPro HEVC videos with no B-frames, but it is not limited to them either. Also seen with iPhone HEVC video.

I do not develop or even build Kdenlive anymore. Shotcut is the no.1 priority here since I am the lead developer of both. I will not merge anything that degrades Shotcut. In fact, if I do not see much benefit I may choose not to merge it to avoid the risk of some other regression. I am still running some tests to see some performance benefit with real world consumer- or prosumer-level files. If I do see a benefit, and the PR seems close, I may try to help and debug. I do see a little with very high (>100 Mb/s) bitrate.

Tjoppen:
You should use samples that have bitrates close to what your storage can deliver, say on the order of 500 Mbit/s for HDDs and 5 Gbit/s for SSDs. We're using ProRes and JPEG2000 samples that easily hit these bitrates, especially lossless JPEG2000.

The sample I've been testing most with uses both P and B frames and I did see some garbled output while developing but not in the state that the patchset is in at the moment. I've suspected that the vpackets deque needs to be cleared when seeking, but producer_get_audio() appears to be able to push vpackets just before seek_video() is called. But perhaps the code hasn't hit that possibility so far for various reasons. Does the garbling persist with the following change?

```patch
diff --git a/src/modules/avformat/producer_avformat.c b/src/modules/avformat/producer_avformat.c
index 86dbd1c6..9f51afac 100644
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1282,9 +1282,16 @@ static int seek_video( producer_avformat self, mlt_position position,
                        // let packets_worker know we handled EOF
                        if ( self->packets_thread_ret == AVERROR_EOF ) {
                                self->packets_thread_ret = 0;
-                               pthread_cond_signal( &self->packets_cond );
                        }
 
+                       // empty vpackets
+                       while ( mlt_deque_count( &self->vpackets ) > 0 ) {
+                               AVPacket *tmp = (AVPacket*) mlt_deque_pop_front( self->vpackets );
+                               av_packet_free( &tmp );
+                       }
+
+                       pthread_cond_signal( &self->packets_cond );
+
                        // Remove the cached info relating to the previous position
                        self->current_position = POSITION_INVALID;
                        self->last_position = POSITION_INVALID;
```

I'm going to see if I can get a build of Shotcut going in the meantime.

Tjoppen:
I got shotcut to build and work with this patchset. Clearing vpackets causes shotcut to hang, so that's not a solution. I plan on staring at the code a bit more to reach enlightenment as to how its moving parts interact. Lunch for now.

frdbr:
As a videomaker I really appreciate this and would love to help test with Kdenlive. (I just need to figure out how to compile MLT with this patch).

Tjoppen:
@frdbr Cool! It shouldn't be harder to compile than master. Locally I've had to disable MOD_RTAUDIO, but that should be it. If you build your own ffmpeg remember to configure it with --enable-shared

I can't get shotcut to actually behave badly with any sample I can think of. I used various movies, youtube videos etc., all long GOP rather than intra only. H.264, JPEG2000 and ProRes. 30/1.001, 25 and 24/1.001 fps.

ddennedy:
> Does the garbling persist with the following change?

No, it is working good now! Please update PR.

Tjoppen:
I tracked down the hang I talked about to me calling mlt_deque_count() with &self->vpackets not self->vpackets. This is something that the compiler should have caught. Please consider adding -Wall and -Werror to CFLAGS. I'll make a ticket for that.

Is seeking ever done for non-seekable sources? I presume not, even though libavformat sometimes permits small amounts of seeking (within its internal avio buffer).

ddennedy:
My IDE warned me about `mlt_deque_count( &self->vpackets )` so I fixed it before testing yesterday. No, it does not try to seek on non-seekable source except upon open a file source to determine whether it provides somewhat accurate seek.

PR #878, opened 2023-02-19
Add link_avdeinterlace and link_deinterlace

bmatherly:
Posting for comment and review

Notes:
* This does not implement frame doubling. That will be part of a future project
* I would like to make the avdeinterlace filter the default (first in order) so that bwdif is available. But i would like to do more testing first.

### DIFF
--- a/src/framework/chain_normalizers.ini
+++ b/src/framework/chain_normalizers.ini
@@ -16,6 +16,6 @@ movitconvert=movit.convert
 imageconvert=avcolor_space,imageconvert
 
 # audio links
+audioconvert=audioconvert
 resampler=swresample,resample
 channels=audiochannels
-audioconvert=audioconvert
--- a/src/framework/chain_normalizers.ini
+++ b/src/framework/chain_normalizers.ini
@@ -6,4 +6,5 @@
 # The names of the services on the right dictate the preference used (if unavailable
 # the second and third are applied as applicable).
 
+deinterlacer=deinterlace,avdeinterlace
 resampler=swresample,resample
--- a/src/framework/chain_normalizers.ini
+++ b/src/framework/chain_normalizers.ini
@@ -6,5 +6,16 @@
 # The names of the services on the right dictate the preference used (if unavailable
 # the second and third are applied as applicable).
 
-deinterlacer=deinterlace,avdeinterlace
+# image links
+deinterlace=deinterlace,avdeinterlace
+fieldorder=fieldorder
+crop=movit.crop,crop:1
+rescaler=movit.resample,swscale,gtkrescale,rescale
+resizer=movit.resize,resize
+movitconvert=movit.convert
+imageconvert=avcolor_space,imageconvert
+
+# audio links
 resampler=swresample,resample
+channels=audiochannels
+audioconvert=audioconvert
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -663,8 +663,12 @@ MLT_7.14.0 {
   global:
     mlt_producer_probe;
     mlt_chain_attach_normalizers;
+} MLT_7.12.0;
+
+MLT_7.16.0 {
+  global:
     mlt_deinterlacer_name;
     mlt_deinterlacer_id;
     mlt_link_filter_init;
     mlt_link_filter_metadata;
-} MLT_7.12.0;
+} MLT_7.14.0;
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -663,4 +663,6 @@ MLT_7.14.0 {
   global:
     mlt_producer_probe;
     mlt_chain_attach_normalizers;
+    mlt_deinterlace_name;
+    mlt_deinterlace_id;
 } MLT_7.12.0;
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -663,6 +663,6 @@ MLT_7.14.0 {
   global:
     mlt_producer_probe;
     mlt_chain_attach_normalizers;
-    mlt_deinterlace_name;
-    mlt_deinterlace_id;
+    mlt_deinterlacer_name;
+    mlt_deinterlacer_id;
 } MLT_7.12.0;
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -665,4 +665,6 @@ MLT_7.14.0 {
     mlt_chain_attach_normalizers;
     mlt_deinterlacer_name;
     mlt_deinterlacer_id;
+    mlt_link_filter_init;
+    mlt_link_filter_metadata;
 } MLT_7.12.0;
--- a/src/framework/mlt_image.h
+++ b/src/framework/mlt_image.h
@@ -60,6 +60,8 @@ extern void mlt_image_fill_opaque( mlt_image self );
 extern const char * mlt_image_format_name( mlt_image_format format );
 extern mlt_image_format mlt_image_format_id( const char * name );
 extern int mlt_image_rgba_opaque(uint8_t *image, int width, int height);
+extern const char * mlt_deinterlace_name( mlt_deinterlace_method method );
+extern mlt_deinterlace_method mlt_deinterlace_id( const char * name );
 
 // Deprecated functions
 extern int mlt_image_format_size( mlt_image_format format, int width, int height, int *bpp );
--- a/src/framework/mlt_image.h
+++ b/src/framework/mlt_image.h
@@ -60,8 +60,6 @@ extern void mlt_image_fill_opaque( mlt_image self );
 extern const char * mlt_image_format_name( mlt_image_format format );
 extern mlt_image_format mlt_image_format_id( const char * name );
 extern int mlt_image_rgba_opaque(uint8_t *image, int width, int height);
-extern const char * mlt_deinterlace_name( mlt_deinterlace_method method );
-extern mlt_deinterlace_method mlt_deinterlace_id( const char * name );
 
 // Deprecated functions
 extern int mlt_image_format_size( mlt_image_format format, int width, int height, int *bpp );
--- a/src/framework/mlt_types.h
+++ b/src/framework/mlt_types.h
@@ -123,18 +123,18 @@ mlt_colorspace;
 
 typedef enum
 {
-	mlt_deinterlace_method_none,
-	mlt_deinterlace_method_bob,
-	mlt_deinterlace_method_weave,
-	mlt_deinterlace_method_onefield,
-	mlt_deinterlace_method_linearblend,
-	mlt_deinterlace_method_greedy,
-	mlt_deinterlace_method_yadif_nospatial,
-	mlt_deinterlace_method_yadif,
-	mlt_deinterlace_method_bwdif,
-	mlt_deinterlace_method_invalid,
+	mlt_deinterlacer_none,
+	mlt_deinterlacer_bob,
+	mlt_deinterlacer_weave,
+	mlt_deinterlacer_onefield,
+	mlt_deinterlacer_linearblend,
+	mlt_deinterlacer_greedy,
+	mlt_deinterlacer_yadif_nospatial,
+	mlt_deinterlacer_yadif,
+	mlt_deinterlacer_bwdif,
+	mlt_deinterlacer_invalid,
 }
-mlt_deinterlace_method;
+mlt_deinterlacer;
 
 /** The time string formats */
 
@@ -291,6 +291,9 @@ extern char* strptime(const char *buf, const char *fmt, struct tm *tm);
 #define mlt_fopen fopen
 #define MLT_DIRLIST_DELIMITER ":"
 #endif /* ifdef _WIN32 */
+
+extern const char * mlt_deinterlacer_name( mlt_deinterlacer method );
+extern mlt_deinterlacer mlt_deinterlacer_id( const char * name );
 
 #ifdef __cplusplus
 }
--- a/src/framework/mlt_types.h
+++ b/src/framework/mlt_types.h
@@ -121,6 +121,21 @@ typedef enum
 }
 mlt_colorspace;
 
+typedef enum
+{
+	mlt_deinterlace_method_none,
+	mlt_deinterlace_method_bob,
+	mlt_deinterlace_method_weave,
+	mlt_deinterlace_method_onefield,
+	mlt_deinterlace_method_linearblend,
+	mlt_deinterlace_method_greedy,
+	mlt_deinterlace_method_yadif_nospatial,
+	mlt_deinterlace_method_yadif,
+	mlt_deinterlace_method_bwdif,
+	mlt_deinterlace_method_invalid,
+}
+mlt_deinterlace_method;
+
 /** The time string formats */
 
 typedef enum
--- a/src/framework/mlt_types.h
+++ b/src/framework/mlt_types.h
@@ -124,10 +124,10 @@ mlt_colorspace;
 typedef enum
 {
 	mlt_deinterlacer_none,
-	mlt_deinterlacer_bob,
-	mlt_deinterlacer_weave,
 	mlt_deinterlacer_onefield,
 	mlt_deinterlacer_linearblend,
+	mlt_deinterlacer_weave,
+	mlt_deinterlacer_bob,
 	mlt_deinterlacer_greedy,
 	mlt_deinterlacer_yadif_nospatial,
 	mlt_deinterlacer_yadif,
--- a/src/modules/avformat/common.h
+++ b/src/modules/avformat/common.h
@@ -22,6 +22,9 @@
 
 #include <framework/mlt.h>
 #include <libswscale/swscale.h>
+#include <libavutil/frame.h>
+
+#define MLT_SWS_FLAGS "bicubic+accurate_rnd+full_chroma_int+full_chroma_inp"
 
 int mlt_to_av_sample_format( mlt_audio_format format );
 int64_t mlt_to_av_channel_layout( mlt_channel_layout layout );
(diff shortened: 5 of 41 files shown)


src/framework/mlt_image.h
```diff
@@ -60,6 +60,8 @@ extern void mlt_image_fill_opaque( mlt_image self );
 extern const char * mlt_image_format_name( mlt_image_format format );
 extern mlt_image_format mlt_image_format_id( const char * name );
 extern int mlt_image_rgba_opaque(uint8_t *image, int width, int height);
+extern const char * mlt_deinterlace_name( mlt_deinterlace_method method );
```
bmatherly, line 63: I am open minded about where these functions should live. The functions themselves do not operate on images. Maybe better somewhere else.

src/framework/mlt_types.h
```diff
@@ -121,6 +121,21 @@ typedef enum
 }
 mlt_colorspace;
 
+typedef enum
+{
+	mlt_deinterlace_method_none,
+	mlt_deinterlace_method_bob,
```
bmatherly, line 127: maybe "method" is not needed - just "mlt_deinterlace_bob" etc.

src/modules/avformat/link_avdeinterlace.c
```diff
@@ -0,0 +1,617 @@
...
+	}
+	ret = avfilter_init_str( pdata->avbuffsrc_ctx, NULL );
+	if( ret < 0 ) {
+		mlt_log_error( self, "Cannot init buffer source\n" );
+		goto fail;
+	}
+	prev_ctx = pdata->avbuffsrc_ctx;
+
+	// Initialize the deinterlace filter context
+	if ( pdata->method <= mlt_deinterlace_method_onefield )
```
bmatherly, line 162: Currently implemented: onefield, linearblend, yadif_nospatial, yadif, bwdif. FFMpeg does not have some of the  modes from the xine module: bob, weave, greedy. But it does have some others: cubicblend, etc.

src/framework/mlt_types.h
ddennedy: Considering the "deinterlace_method" property is deprecated, and the new property name is "deinterlacer" these should be named `mlt_deinterlacer_` . See 312d2b3b2c44b0054a18499dc63036e0d5c1c55c and 493fa4b642099d295ea5b2b78f85a366a3d8c9ad for those changes about a year ago.

src/modules/avformat/link_avdeinterlace.c
ddennedy: The "bw" in "bwdif" is for bob and weave, but I do not know if you can put it into one more or another. You could map both to bwdiff. All unknown should fall to some default, maybe yadif since it is both fast and good. bwdif might be better quality, but how is performance compared?

src/framework/mlt_image.h
bmatherly: Good idea. Done.

src/framework/mlt_types.h
bmatherly: I like that better. Done.

src/modules/avformat/link_avdeinterlace.c
bmatherly: > The "bw" in "bwdif" is for bob and weave, but I do not know if you can put it into one more or another. 

bwdif performs motion adaptive deinterlacing based on yadif with the use of w3fdif and cubic interpolation algorithms. There are not options to simplify it down to a classic "bob" or "weave". Personally, I think that bob and weave have fallen out of relevance since we have much higher performance computers now. It is OK with me if our fastest deinterlace methods are one field and linear blend.

> All unknown should fall to some default, maybe yadif since it is both fast and good.

I prefer to let each implementation define its own default. I chose the best option as default for both:
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/modules/avformat/link_avdeinterlace.c#L256
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/modules/xine/common.c#L99

> bwdif might be better quality, but how is performance compared?

In my past experience, bwdif uses more resources. But in this modern day of computing, I think it is negligible. I'll do some performance comparisons to be sure.

I tried to list the types in the enum in ascending order of difficulty/quality:
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/framework/mlt_types.h#L126
the idea is that if a service does not support a requested mode, it will fail down to the next best mode that it does support
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/modules/xine/common.c#L90
When I do some performance testing, I will be sure to put them in best order.

bmatherly:
While doing some performance testing, I found some bugs in the deinterlacer. I found out that the CPU acceleration test was being skipped and all the xine deinterlacers were failing back to linearblend regardless of what deinterlacer the user chose. 

Here are the performance results from my testing:
image

Some conclusions:
* I reordered the deinterlacer priority based on these metrics
* The FFMpeg yadif is almost twice as fast as the xine yadif implementation.

ddennedy:
Are you planning a change in this PR to use #879?

Are you planning in this PR to remove yadif from the deinterlace filter and the "_need_previous_next" mlt_service property? I guess that could be detrimental to the apps (flowblade?) and users (melt without `-chain` and old projects/scripts) not yet using chains. Maybe we indicate it is deprecated and remove it in the future.

src/modules/avformat/common.h
```diff
@@ -22,6 +22,9 @@
 
 #include <framework/mlt.h>
 #include <libswscale/swscale.h>
+#include <libavutil/frame.h>
+
+#define MLT_SWS_FLAGS "bicubic+accurate_rnd+full_chroma_int+full_chroma_inp"
```
ddennedy, line 27: `MLT_SWS_FLAGS` is very similar to `mlt_get_sws_flags`. Since it for use with avfilter-related things consider to change its name to `MLT_AVFILTER_SWS_FLAGS`

src/modules/avformat/link_avdeinterlace.c
```diff
@@ -0,0 +1,617 @@
...
+	mlt_properties frame_properties = MLT_FRAME_PROPERTIES( frame );
+
+	struct mlt_image_s srcimg = {0};
+	struct mlt_image_s dstimg = {0};
+	mlt_deinterlacer method = mlt_deinterlacer_id( mlt_properties_get( frame_properties, "consumer.deinterlacer" ) );
+
+	if ( !mlt_properties_get_int( frame_properties, "consumer.progressive" ) ||
+		 method == mlt_deinterlacer_none )
+	{
+		mlt_log_info( MLT_LINK_SERVICE(self), "Do not deinterlace - consumer is interlaced\n" );
```
ddennedy, line 338: Everywhere there is `mlt_log_info` needs reevaluation about that log level. Shotcut release builds log at info level; so, there could be many new messages flooding its log. Shotcut melt jobs and many melt users run with verbose level. So, some should be verbose and maybe some debug. Debug output is usually too much, and I often change certain log calls to verbose or info to investigate something and then change it back.

src/framework/chain_normalizers.ini
```diff
@@ -6,4 +6,5 @@
 # The names of the services on the right dictate the preference used (if unavailable
 # the second and third are applied as applicable).
 
+deinterlacer=deinterlace,avdeinterlace
```
ddennedy, line 9: I was just thinking about adding a movit-based deinterlacer when I though to myself filter or link? It turns out to be a link because it being YADIF-based takes 5 fields for input buffers and can be driven to frame double. I am going to do something else while waiting for this to merge, and let me know if you want to tackle that next as I think it would a good next step.

But help me understand the priority between filter and link for the same purpose. Of course, if you are not using a chain then a link is not used. The chain wraps a producer that has normalizing filters attached. So how does the smarter link-based resampler get used if there is a filter-based resampler upstream that already did the work? This along with chains and links in general needs a document in the miscellaneous section of https://mltframework.org/docs/. Do not need to go into huge depth but cover the basic concepts, purposes, and ways they get included including API, melt, and XML.

src/modules/avformat/common.h
bmatherly: Good idea. Done

src/modules/avformat/link_avdeinterlace.c
bmatherly: I think we are mostly on the same page. I find it helpful when the Shotcut log includes some one-time informational messages. For example, when a resources is allocated for the first time in a filter. I do not use mlt_log_info() for recurring messages. I did change a couple of log levels in this last patch.

src/framework/chain_normalizers.ini
bmatherly: > I was just thinking about adding a movit-based deinterlacer when I though to myself filter or link? It turns out to be a link because it being YADIF-based takes 5 fields for input buffers and can be driven to frame double. I am going to do something else while waiting for this to merge, and let me know if you want to tackle that next as I think it would a good next step.

I am happy to work on a movit-based deinterlacer link. But first I would like to finish my "forward speed filter" work that I have in progress.

> But help me understand the priority between filter and link for the same purpose. Of course, if you are not using a chain then a link is not used. 

This is the "magic" that allows a link to tell a downstream normalizing filter that it does not need to do anything:
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/modules/xine/link_deinterlace.c#L42
Essentially, you have to manipulate the frame and get_image() parameters so that the downstream normalizers do not think that anything needs to be done.
This unfortunately means that normalization is needed after the deinterlacer link. This is a work-around for now:
https://github.com/bmatherly/mlt/blob/deinterlace_link/src/modules/xine/link_deinterlace.c#L268
But eventually I will make normalizer links for all normalization types and then we will not need normalizers on the encapsulated producer.

> This along with chains and links in general needs a document in the miscellaneous section of https://mltframework.org/docs/. Do not need to go into huge depth but cover the basic concepts, purposes, and ways they get included including API, melt, and XML.

That is a good idea. I will add that to my todo list.

bmatherly:
> Are you planning a change in this PR to use https://github.com/mltframework/mlt/pull/879?

This is done now. Unfortunately, I had to rebase and force-push so that I could get the mlt_producer_probe() function.

> Are you planning in this PR to remove yadif from the deinterlace filter and the "_need_previous_next" mlt_service property? I guess that could be detrimental to the apps (flowblade?) and users (melt without -chain and old projects/scripts) not yet using chains. Maybe we indicate it is deprecated and remove it in the future.

After we merge this PR, I think we can mark them deprecated. I am not exactly sure where to mark them. Maybe in the YAML file.

As eager as I am, I think we should preserve "_need_previous_next" for a couple of releases before we remove it.

I have more aspirations for the deinterlacer links:
* Test enough to make the ffmpeg deinterlacer the default
* Frame doubling
* Improve expectation setting so that the filter does not have to work so hard to detect when it should be engaged

But I think this PR is in pretty good shape and could be a good step forward for the next release if we do not get to some of the other improvements in time.

src/modules/xine/link_deinterlace.c
```diff
@@ -0,0 +1,331 @@
...
+		mlt_filter_process( pdata->rescale_filter, *frame );
+	}
+	if ( !pdata->resize_filter )
+	{
+		pdata->resize_filter = mlt_factory_filter( mlt_service_profile( MLT_LINK_SERVICE( self ) ), "resize", NULL );
+	}
+	if ( pdata->resize_filter )
+	{
+		mlt_filter_process( pdata->resize_filter, *frame );
+	}
```
ddennedy, line 284: It is not good that a link needs to codify the normalizing filters that the loader producer is responsible for. What happens with crop and/or movit?

bmatherly: I agree. It is an unfortunate work-around.

Movit was not a factor when I started this project - I did not know you were going to resurrect it :)

Crop Source still works since there is nothing defeating the crop and pad normalizers.

Maybe I should prioritize making corresponding normalizing links. They would just be very simple wrappers around the existing filters.

bmatherly:
> Maybe I should prioritize making corresponding normalizing links. They would just be very simple wrappers around the existing filters.

OK, I feel like I have a fairly robust solution for this now. I'm glad I took the time to make a more comprehensive solution. We also have a convenient feature now that we can easily convert any existing filter into a link if we want.

> Movit was not a factor when I started this project - I did not know you were going to resurrect it :)

Movit normalizers work with this solution as well. Timeremap does not work with movit - but that is not a regression since it does not work in master, either.

ddennedy:
> Timeremap does not work with movit 

I have made a project in Shotcut (qt6 branch, MLT master branch) where it is working. I will check it again with this branch.

src/framework/mlt.vers
```diff
@@ -663,4 +663,8 @@ MLT_7.14.0 {
   global:
     mlt_producer_probe;
     mlt_chain_attach_normalizers;
+    mlt_deinterlacer_name;
+    mlt_deinterlacer_id;
+    mlt_link_filter_init;
+    mlt_link_filter_metadata;
```
ddennedy, line 669: Please change these to be in a new block `MLT_7.16.0` derived from `MLT_7.14.0`.

src/framework/chain_normalizers.ini
```diff
@@ -6,4 +6,16 @@
 # The names of the services on the right dictate the preference used (if unavailable
 # the second and third are applied as applicable).
 
+# image links
+deinterlace=deinterlace,avdeinterlace
```
ddennedy, line 10: Let's consider moving avdeinterlace ahead of deinterlace. For fast preview, avdeinterlace onefield and linearblend are almost as fast. And for high quality, avdeinterlace yadif is much faster and also offers bwdif.

src/framework/mlt.vers
bmatherly: done

src/framework/chain_normalizers.ini
bmatherly: That is my plan. But I need more testing/integration first. I do not know the consequences of images being processed out of order when more than one thread is used.

These are my next steps for this project:
1) Test/debug to make sure avdeinterlace works with parallel threads
2) Implement a way for the framework to pass down expectations from the consumer before the first frame is requested
3) Implement frame doubling in avdeinterlace

I would be happy for this branch to be merged now since it is in a workable state with parity performance. Or I am willing to keep working in this branch until all the steps are complete. I would trust your judgement on when to merge.

BTW: I do not plan to implement a movit deinterlace link. However, I am available to help with integration and debugging if someone else decides to make a proof-of-concept.

ddennedy: It is OK to not have a movit.deinterlace yet because deinterlace is first in the normalizers list letting it run immediately after decode and before uploading to OpenGL.

ddennedy:
I found a problem when testing in Shotcut with GPU Effects on. Turn off Settings > Player > Progressive and open an interlaced clip. It looks fine in the Source player. Add it to the Timeline, and the video is garbage. I have not yet found a way to reproduce in melt. Actually, I have this problem even with progressive videos. The only combo that works after adding clip to timeline is interlaced with player progressive turned on. I will try to look into to too over the next couple of days starting tomorrow.

bmatherly:
Thanks for your testing, Dan. Your testing actually found two problems:

1) Frames are corrupted whenever the deinterlacer is not engaged. I wrote the deinterlace links to be a complete pass-through when they are not needed. Their get_image function is not even called. But for some reason, the frames are corrupted when the get_image stack goes directly from movit to avformat with nothing in between.

A simple way to test this is to comment out the first two normalizers in chain_normalizers.ini. The image will always be corrupted. We need to fix it so that movit works when these two normalizer links are not installed:

#deinterlace=deinterlace,avdeinterlace
#fieldorder=fieldorder

2) Regardless of the player "Progressive" checkbox, frames from the player always have "consumer.progressive" set to 1. The deinterlace filters rely on consumer.progressive to know if the consumer is interlaced or not. But with the Shotcut player, frames are always flagged as progressive.

ddennedy:
Take a look at `mlt_tractor.c:producer_get_image()`. Frames only reliably get the "consumer."-prefixed properties when get_image is called, not get_frame (when it calls `mlt_properties_copy(frame_properties, properties, "consumer.");`). It is best to not try to be so aggressive about skipping get_image and simply skip the extra work in your get_image function when desirable (but you still need to call `mlt_frame_get_image()` inside it).

bmatherly:
The root problem was that normalizers were being added twice when the chain was constructed from XML. That fixed most of the problems. I also made the interlace detection more robust at the cost of some efficiency. In the future, in order to support frame doubling, the chain will need to know the interlace status before frames are requested so that the links can be reconfigured to handle normal or double frame rate. So we should get the efficiency back at that time.

ddennedy:
I have an interlaced DV file that I often use for testing deinterlace, and it has 32 KHz audio that is silent. When I add it to the timeline, it has a loud repetitive ticking sound (but not in source player). Conversely, when I open a 44.1 Khz music file (Ogg Vorbis), it is very noisy in source player, but when I add it to the timeline it sounds OK. GPU Effects is Off, and this does not happen in the 23.03 alpha.

ddennedy:
FWIW I uploaded that DV file to our shared_media folder on S3 at `shared_media/dv/bouncing_ball.dv`

bmatherly:
> I have an interlaced DV file that I often use for testing deinterlace, and it has 32 KHz audio that is silent. When I add it to the timeline, it has a loud repetitive ticking sound (but not in source player). 

Thank you for all your testing. I really appreciate it. The audio is working much better for me now.

> FWIW I uploaded that DV file to our shared_media folder

Thanks. I have added this to my test checklist

Also, the time remap filter is working with GPU effects now.

src/modules/xine/common.c
```diff
@@ -0,0 +1,180 @@
...
+{
+	if ( !dst || !src || !dst->data || !src->data )
+	{
+		return 1;
+	}
+
+	if ( method >= mlt_deinterlacer_yadif_nospatial &&
+		 ( !prev || !next || !prev->data || !next->data ) )
+	{
+		method = mlt_deinterlacer_linearblend;
```
ddennedy, line 113: It looks like this is the same behavior as the existing filter, but I think now upon review while testing onefield is a better looking fallback as its appearance is much closer to what yadif produces.

bmatherly: I had thought the same thing. But I didn't care to break from the tradition since I am replacing it with the ffmpeg based deinterlacers. I suppose it only occurs on the first frame (when there is no previous) and the last frame (when there is no next). That will not be an issue with FFMpeg based deinterlacers.

ddennedy: Another way to deal with missing frames is to repeat the correct frame.

PR #879, opened 2023-02-20
Add mlt_producer_probe()

bmatherly:
I need a way for links to be able to query the source producer before they have received a frame. This idea adds a function that will cause the producer to probe itself if necessary. Subsequent calls can skip the probe if the data is current. This implementation skips the entire mlt frame stack so that normalizing filters are skipped.

I have a concept patch here that uses the probe function to simplify the avformat producer widget. I think this might avoid the problem with the GPU thread. But I have not  tested yet.
https://github.com/bmatherly/shotcut/commit/00be4019aeb7ca3a9471676cba8bca07d8182b88

Do you think this is promising/interesting? If so, I will do more testing.

### DIFF
--- a/src/framework/mlt_producer.c
+++ b/src/framework/mlt_producer.c
@@ -1291,8 +1291,11 @@ void mlt_producer_set_creation_time( mlt_producer self, int64_t creation_time )
 
 int mlt_producer_probe( mlt_producer self )
 {
-	if ( self && self->probe )
+	if ( self )
 	{
-		return self->probe( self );
+		int ( *probe )( mlt_producer ) = mlt_properties_get_data( MLT_PRODUCER_PROPERTIES(self), "_probe", NULL );
+		if ( probe )
+			return probe( self );
 	}
+	return 0;
 }
--- a/src/framework/mlt_producer.c
+++ b/src/framework/mlt_producer.c
@@ -1279,3 +1279,20 @@ void mlt_producer_set_creation_time( mlt_producer self, int64_t creation_time )
 	mlt_properties_set( MLT_PRODUCER_PROPERTIES( parent ), "creation_time", datestr);
 	free( datestr );
 }
+
+/** Probe the producer to publish metadata properties.
+ *
+ * After this call the producer will publish meta.media properties
+ *
+ * \public \memberof mlt_producer_s
+ * \param self a producer
+  * \return true on error
+ */
+
+int mlt_producer_probe( mlt_producer self )
+{
+	if ( self && self->probe )
+	{
+		return self->probe( self );
+	}
+}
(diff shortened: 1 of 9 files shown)


ddennedy:
> This implementation skips the entire mlt frame stack so that normalizing filters are skipped.

One can use the `abnormal` producer today to use loader producer without normalizing filters attached.

src/framework/mlt_producer.h
```diff
@@ -97,6 +97,12 @@ struct mlt_producer_s
 	 */
 	int ( *set_in_and_out )( mlt_producer, mlt_position, mlt_position );
 
+	/** Probe for metadata.
+	 *
+	 * \param mlt_producer a producer
+	 * \return true if there was an error
+	 */
+	int ( *probe )( mlt_producer );
```
ddennedy, line 105: This breaks ABI, which is not acceptable.

bmatherly:
> One can use the abnormal producer today to use loader producer without normalizing filters attached.

It seems heavy weight to instantiate an entire new producer when the code currently has a handle to a producer that is already open. For example, I expect to use this function in a link so it can query an upstream producer.

src/framework/mlt_producer.h
bmatherly: Does this mean it is impossible to add a new member to the producer struct? I guess I could store it in a property.

ddennedy: It is temping to simply move the new member to the end of struct, and any theoretical existing built code that is allocating the struct and/or copying it cannot benefit from the new feature anyways. However, there are other structs that embed this struct first, and then it changes the offset of its members that follow! Maybe instead of using a property use a mlt_event.

src/framework/mlt_producer.c
```diff
@@ -1291,8 +1291,11 @@ void mlt_producer_set_creation_time( mlt_producer self, int64_t creation_time )
 
 int mlt_producer_probe( mlt_producer self )
 {
-	if ( self && self->probe )
+	if ( self )
 	{
-		return self->probe( self );
+		int ( *probe )( mlt_producer ) = mlt_properties_get_data( MLT_PRODUCER_PROPERTIES(self), "_probe", NULL );
```
ddennedy, line 1296: To make this property name slightly cleaner while still mitigate name conflict:
a) The leading underscore is not needed or that useful on data properties that have no serializer. These are naturally already hidden.
b) Prefix the property name like we do for programming: `mlt_producer_probe` 
What do you think?

bmatherly: That is a good idea. Done.

bmatherly:
I created a pull request to use this in the qt6 branch. I tested with GPU mode, but I do not know what might trigger a crash. I wonder if you could test it to be sure it does not cause problems with GPU mode.

PR #880, opened 2023-02-21
Fix consumer: resources incorrectly prefixed with root

j-b-m:
The `consumer` producer is currently not correctly recognized when loading xml through the producer_xml. This means that the document root is prepended if you have a resource like: `consumer:path_to_resource`, breaking the producer.

This can be reproduced with this command line (timewarping an mlt playlist) :
`melt timewarp:4:consumer:playlist.mlt -consumer xml:playlist-fast.mlt`

Trying to play the newly created file fails with:
`[producer_xml] failed to load producer "4:/$PATH_TO_SEQUENCE.MLT/consumer:playlist-fast.mlt"

This is because the `consumer` producer is not currently recognized by producer_xml, like `colour`.
`

### DIFF
--- a/src/modules/xml/producer_xml.c
+++ b/src/modules/xml/producer_xml.c
@@ -262,7 +262,8 @@ static inline int is_known_prefix(const char* resource)
 			"udplite",
 			"unix",
 			"color",
-			"colour"
+			"colour",
+			"consumer"
 		};
 		size_t i, n = prefix - resource;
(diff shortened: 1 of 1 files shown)


ddennedy:
Please be aware that using timewarp with MLT XML and consumer producer is not tested. Also, this scenario will mask the higher frame rate in sources. timewarp is most useful with high frame rate sources in a normal frame rate project.

PR #881, opened 2023-02-28
Use same argument name in mlt_factory.h as mlt_factory.c

Cj-bc:
# Background

While reading [doxgen document for `mlt_factory_consumer` function](https://www.mltframework.org/doxygen/mlt__factory_8h.html#a26c79428c0d8fb51c001dbec152da83a), I found that 2nd argument name is different between "Functions" section and actuall individual document.

In "Functions" section, it says 2nd argument is "name"
image

whereas actuall individual document says it is "service"
image

I thought this is really confusing, because the name "name" sounds like I can give arbitrary name to it (instead of valid service name). so I wrote this PR.

I am not familiar with C language, so actually I am not sure if I can rename it or not.
I am glad if you let me know the reason if it is impossible to change it.

# What this PR does
Just renamed some arguments of some functions in `mlt_factory.h`. It makes it clear that those functions should receive service name, not arbitrary name.
Even though the first motivation was to fix `mlt_factory_consumer` declaration, I fixed some other functions as well.


ddennedy:
Yes, it is OK to rename C function parameters. We do not use the headers for our function documentation but rather the `.c` file. It is not clear why you would be caring about the parameter names in the header if you are not a C programmer. Is it affecting some SWIG-generated language binding? Most of those are based on the C++ declarations, which use `id` for this parameter to the respective `Mlt::Factory` methods.  MLT does not have the notion of an object name or ID. If you want to name an object you basically need to set a custom property with a prefix to name-space it like "cjbc:id". Then, you can traverse the object tree using `mlt_parser` or `Mlt::Parser` to find an object by its custom property.

Cj-bc:
Thank you for your kind response :)

> It is not clear why you would be caring about the parameter names in the header if you are not a C programmer. Is it affecting some SWIG-generated language binding?

No. I am planning to try making bindings for Rust/Haskell/Golang(not decided yet) as a part of my hobby project, and I just wanted to try native API to know how it works.
I went to doxgen document because I encounted some problems while reading and trying codes on [Framework design document on www.mltframework.org](https://www.mltframework.org/docs/framework/).

PR #882, opened 2023-03-03
consumer_avformat: interlace aware chroma conversion 422 -> 420

Jolyk:
Substitutes width and height to look like side-by-side before conversion. Expects height is multiple of 4, width is multiple of 8 (system dependent). Covers standard resolution for PAL and NTSC.
It looks like works well with rgb source too.
Doesn't affect picture in editor.
Fixes: https://github.com/mltframework/shotcut/issues/1270

### DIFF
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -1901,8 +1901,7 @@ static void *consumer_thread( void *arg )
 						mlt_frame_get_image( frame, &image, &img_fmt, &img_width, &img_height, 0 );
 
 						// Interlaced 420 correction
-						if ( mlt_properties_get( frame_properties, "progressive" )
-						     && mlt_properties_get_int( frame_properties, "progressive" ) == 0
+						if ( !mlt_properties_get_int( frame_properties, "progressive" )
 						     && pix_fmt == AV_PIX_FMT_YUV420P  // dst
 						     && img_fmt == mlt_image_yuv422    // src. It looks like rgb and 444 go as 422 too.
 						     && height % 4 == 0		       // because reducing twice
--- a/src/modules/avformat/consumer_avformat.c
+++ b/src/modules/avformat/consumer_avformat.c
@@ -1896,7 +1896,25 @@ static void *consumer_thread( void *arg )
 					if ( mlt_properties_get_int( frame_properties, "rendered" ) )
 					{
 						AVFrame video_avframe;
+						int is_interlaced_chroma_correction = 0;
+
 						mlt_frame_get_image( frame, &image, &img_fmt, &img_width, &img_height, 0 );
+
+						// Interlaced 420 correction
+						if ( mlt_properties_get( frame_properties, "progressive" )
+						     && mlt_properties_get_int( frame_properties, "progressive" ) == 0
+						     && pix_fmt == AV_PIX_FMT_YUV420P  // dst
+						     && img_fmt == mlt_image_yuv422    // src. It looks like rgb and 444 go as 422 too.
+						     && height % 4 == 0		       // because reducing twice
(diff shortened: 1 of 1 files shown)


src/modules/avformat/consumer_avformat.c
```diff
@@ -1896,8 +1896,26 @@ static void *consumer_thread( void *arg )
 					if ( mlt_properties_get_int( frame_properties, "rendered" ) )
 					{
 						AVFrame video_avframe;
+						int is_interlaced_chroma_correction = 0;
+
 						mlt_frame_get_image( frame, &image, &img_fmt, &img_width, &img_height, 0 );
 
+						// Interlaced 420 correction
+						if ( mlt_properties_get( frame_properties, "progressive" )
+						     && mlt_properties_get_int( frame_properties, "progressive" ) == 0
```
ddennedy, line 1905: No need for `mlt_properties_get()`. These two conditions should be simplified to
`!mlt_properties_get_int( frame_properties, "progressive" )`
since getting a property that does not exist as integer returns 0.

src/modules/avformat/consumer_avformat.c
```diff
@@ -1896,8 +1896,26 @@ static void *consumer_thread( void *arg )
...
+						     && pix_fmt == AV_PIX_FMT_YUV420P  // dst
+						     && img_fmt == mlt_image_yuv422    // src. It looks like rgb and 444 go as 422 too.
```
ddennedy, line 1907: > It looks like rgb and 444 go as 422 too.

`mlt_consumer` has a render thread that defaults to `mlt_image_yuv422`. If you want to change it to RGB, you need to set consumer property `mlt_image_format=rgb` (or `rgba`). There is no YUV 4:4:4 in MLT.
Is it necessary to check the MLT image format (`img_fmt`) here? You are manipulating only `converted_avframe` which is based on `pix_fmt`.

ddennedy (changes requested):
Thanks for your interesting contribution!

src/modules/avformat/consumer_avformat.c
Jolyk: I know that it will return 0, but I wanted to be sure, that this property exists. I don't want frame without this property considered as interlaced. May be property is just missed (or it is impossible?). But if you still want I will make changes.

src/modules/avformat/consumer_avformat.c
Jolyk: I don't want change to RGB. I just noticed side-effect, that RGB file on input works too with this fix. 
About img_fmt: I tried to check as much as possible to not to break other formats. It is just couple of assembly instructions. What if input format would be already 420. My correction would be wrong. I would prefer to do this check.

src/modules/avformat/consumer_avformat.c
ddennedy: > I don't want frame without this property considered as interlaced.

But this is the way MLT works. The default is interlaced.

src/modules/avformat/consumer_avformat.c
ddennedy: I understand your concern, and you can instead change that line to
`img_fmt != mlt_image_yuv420p`
I let you decide; it is not a required change.

src/modules/avformat/consumer_avformat.c
Jolyk: I made changes. But I'm not sure should I press button "Resolve conversation" or something else. It is my first pull request on github.

src/modules/avformat/consumer_avformat.c
Jolyk: I prefer to stay "as is" here.

PR #883, opened 2023-03-05
profile description and colorspace corrupted with consumer prod

j-b-m:
Project profile's description and colorspace were overwritten by the profile from the consumer.
Reproducible with:
melt -profile dv_pal explicit=1 color:red -track hd-playlist.mlt -consumer xml

The output xml's profile would now use the description of the profile embedded in hd-playlist.mlt

### DIFF
--- a/src/modules/core/producer_loader.c
+++ b/src/modules/core/producer_loader.c
@@ -145,6 +145,7 @@ static mlt_producer create_producer( mlt_profile profile, char *file )
 			profile->sample_aspect_num = backup_profile->sample_aspect_num;
 			profile->width = backup_profile->width;
 			profile->colorspace = backup_profile->colorspace;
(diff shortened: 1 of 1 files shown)


src/modules/core/producer_loader.c
```diff
@@ -144,6 +144,8 @@ static mlt_producer create_producer( mlt_profile profile, char *file )
 			profile->sample_aspect_den = backup_profile->sample_aspect_den;
 			profile->sample_aspect_num = backup_profile->sample_aspect_num;
 			profile->width = backup_profile->width;
+			profile->colorspace = backup_profile->colorspace;
+			profile->description = strdup( backup_profile->description );
```
bmatherly, line 148: Does description need to be freed to avoid memory leak? I am not sure.

Issue #741, opened 2021-10-15
Cropped FLAC produces audio artifacts (crackles, pops) 


tuffnerdstuff:
### Description
When rendering an MLT file referencing a FLAC audio file, the rendered output will contain audible artifacts (crackles, pops) if the producers in offset is greater than zero (beginning is cropped), otherwise the audio is clean.

### Reproduction
1. Generate a 2 second sine wave FLAC audio using ffmpeg: `ffmpeg -f lavfi -i "sine=frequency=1000:duration=2" sine.flac`
2. Render a WAV 0 seconds in (duration 2 seconds): `melt -profile atsc_1080p_60 sine.flac in=00:00:00.000 out=00:00:02.000 -consumer avformat:sine_clean.wav`
3. Render a WAV 1 second in (duration 1 second): `melt -profile atsc_1080p_60 sine.flac in=00:00:01.000 out=00:00:02.000 -consumer avformat:sine_crackle.wav`

You will notice that sine_clean.wav is fine, while sine_crackle.wav is distorted.

### Workaround
If you use uncompressed WAV audio instead of FLAC there are no issues whatsoever.

### Technical Info
* MLT version: `7.0.1`
* FFmpeg version: `n4.4`
* OS: `5.10.68-1-MANJARO #1 SMP PREEMPT Wed Sep 22 12:29:47 UTC 2021 x86_64 GNU/Linux`

### Additional Info
I also tried different FLAC bit-depths (16 and 24 bit) as well as sampling-rates (44,1kHz and 48kHz), but the problem stays the same. If I do the same using ffmpeg, there is no distortion:
`ffmpeg -i sine.flac -ss 00:00:01.000 sine_ffmpeg.wav`

tuffnerdstuff:
Thanks for the quick reply! I will try using FLAC in a container (e.g. Matroska) and report back if I find another workaround. I'll also try to reproduce the problem using FFmpeg only. Is there a way to somehow log what MLT inputs into FFmpeg?

frdbr:
In Kdenlive we've noticed crackles/pops when using 60fps projects. It doesn't happen with 24/30 fps though... Don't know if this is relevant but hope it helps find a fix:  

https://bugs.kde.org/show_bug.cgi?id=410726

ddennedy:
@frdbr The problem with the kdenlive bug you reference is that it is more about general audio distortions, many of which have been fixed or significantly improved, while this is specifically about FLAC as a source.

[...]

PR #885, opened 2023-03-10
Fix cracking sound with FLAC audio

bmatherly:
Account for samples already in the buffer when calculating PTS synchronization.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -2705,6 +2705,7 @@ static int decode_audio( producer_avformat self, int *ignore, const AVPacket *pk
 
 	int channels = codec_context->channels;
 	int audio_used = self->audio_used[ index ];
+	int audio_used_at_start = audio_used;
 	int ret = 0;
 	int discarded = 1;
 	int sizeof_sample = sample_bytes( codec_context );
@@ -2763,6 +2764,7 @@ static int decode_audio( producer_avformat self, int *ignore, const AVPacket *pk
 		int n = FFMIN( audio_used, *ignore );
 		*ignore -= n;
 		audio_used -= n;
+		audio_used_at_start -= n;
 		memmove( audio_buffer, &audio_buffer[ n * channels * sizeof_sample ],
 				 audio_used * channels * sizeof_sample );
 	}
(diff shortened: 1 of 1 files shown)


tuffnerdstuff:
I tested your fix with my setup described in #741 and the audio is clear now!

ddennedy:
The description is misleading because the change is not limited to FLAC. Since this affects everything with audio and especially audio/video synchronization this needs a lot of testing across a variety of inputs that stress seeking (multiple splits) while assessing sync. Also, needs paying attention to continuity of audio at clip split points.

bmatherly:
> The description is misleading because the change is not limited to FLAC

I tried to make a change that was specific to FLAC but it was not sufficient.

> this needs a lot of testing

Agreed. And with this more accurate PTS calculation, I wonder if we want to tighten up the tolerances before we do all that testing. For example, maybe we want the ahead_threshold ot be in units of samples or seconds instead of video frames (since video frame duration changes based on frame rate which is why we get different results at different frame rates).

ajyoon:
For what it's worth, I was going crazy trying to fix crackles and pops in kdenlive with WAV pcm s24le audio persistently appearing at clip boundaries in preview and renders, and this branch seems to have fixed it. Could be a coincidence though if this really is FLAC specific.

ddennedy:
@ajyoon What were the sample rates of your WAV and that of the render output?

ajyoon:
I'm working in 24bit signed WAV, which I think may have contributed to the bugs since [the kdenlive docs recommend 16bit](https://docs.kdenlive.org/en/troubleshooting/windows_issues.html#audio-pops-and-ticks-in-render)

ddennedy:
@ajyoon Sample rate aka frequency, not sample format, please.

ajyoon:
oh my bad - 48k (and 4 channels if that's relevant)

ddennedy:
Yes, that helps a lot, thanks!

ddennedy:
@bmatherly Since you wrote "I wonder if we want to tighten up the tolerances" I assume this PR is not ready to merge, but you did not create it as a draft PR. So, I added the pending label. Remove the pending label when you think it is ready, and in the future you can create a PR as a draft when it is a proposal and still testing. A draft PR shows a "Ready for Review" button you can click when it is really ready.

bmatherly:
> Since you wrote "I wonder if we want to tighten up the tolerances" I assume this PR is not ready to merge, but you did not create it as a draft PR

This is good to go from my perspective. I only suggested tightening the tolerance, because from a math perspective, there is no reason that ahead_threshold could not be 1. But we would have to do a LOT of testing to have confidence in that.

This change will improve any audio format that has a packet duration that is 3 video frames or longer. In the case of 60fps video and 48kHz audio, one video frame is the duration of 800 audio samples. For FLAC, one audio packet is 4096 frames - That is more than 5 video frames. AC3 audio typically has 1536 samples per packet. So that is under the 3 frame threshold. I do not know the typical number of samples per packet for other formats.

PR #889, opened 2023-03-21
Astyle

bmatherly:
Add an astyle target to cmake and apply it to all the code.

I started with the coding style guide as the basis for the configuration file:
https://mltframework.org/docs/codingstyleguide/

Then, I tweaked options to try to minimize the number of changes across the files (follow the majority consensus of style).

Some areas were inconsistent and I had to "choose" a winner. For example, the MLT++ code indents classes and the tests do not. So I chose to indent and all the test files had to change.

Unfortunately, I could not find a style that will break the namespace brace to a newline.


ddennedy:
I don’t really like the documented style anymore that puts spaces inside parentheses and all curly braces on a new line. I was gradually moving away from that. And if I were to reformat all code then maybe also spaces for indent instead of tabs.

ddennedy:
Before you change this based on my feedback, what is your opinion? Probably this approach minimizes the changes. Should I conform to the convention, or is a more typical style preferable?

bmatherly:
The goal of this patch was to minimize changes while making the code consistent. I do not have a strong style preference - but I do have a strong preference to have an automated tool so that I can ignore formatting while I edit, and then run the tool to make the code compliant before I commit.

Personally, I prefer the [Shotcut style](https://github.com/mltframework/shotcut/blob/master/.astylerc). And it would be handy if I could configure my IDE for one style that would work for both.

But matching the Shotcut style would cause a whitespace change to every single line - which would make GitHub Blame feature useless for quite a while. There are CLI/Desktop Git clients that are able to ignore whitespace in their blame display. But as far as I know, the GitHub blame will show whitespace changes.

ddennedy:
> which would make GitHub Blame feature useless for quite a while. There are CLI/Desktop Git clients that are able to ignore whitespace 

Qt Creator has ignore whitespace for blame and diff views and also provides blame previous revision.

bmatherly:
> Qt Creator has ignore whitespace for blame and diff views and also provides blame previous revision.

Sublime merge also has an option to ignore whitespace changes in blame. I think it is a common feature. Just not in GitHub.

Would you be interested to see the change set if I apply the Shotcut formatting?

ddennedy:
To clarify, clang does not have a format. Rather there is command line utility named “clang-format”. It supports a number of presets:

>    - Set coding style. <string> can be:
>                                   1. A preset: LLVM, GNU, > Google, Chromium, Microsoft,
>                                      Mozilla, WebKit.

Shotcut’s style is the same as “Qt” in Qt Creator, and Qt is not listed or it might be exactly the same as another. So, now there is not only a decision to make about style but also the formatting tool, not to mention the color of the bike shed. Sometimes in these situations I simply do nothing, and life is better.

eszlari:
> Rather there is command line utility named [clang-format](https://clang.llvm.org/docs/ClangFormat.html).

To clarify, the Qt Creator built-in ClangFormat plugin (not "Beautifier") doesn't use the clang-format command line utility, but instead https://github.com/llvm/llvm-project/tree/main/clang/lib/Format directly and supports format while typing: https://www.qt.io/blog/2019/04/17/clangformat-plugin-qt-creator-4-9

bmatherly:
> not to mention the color of the bike shed

I was worried about analysis paralysis when I submitted this request. I think that automated formatting is a step forward in maturity for any project and promotes increased contributions. I do not care what format nor what tool. I choose the current tool and formatting in this patch since it represents the smallest code change possible and the highest match to existing style. I volunteer to do all the work to make the chosen tool match the chosen format. I just hope you will accept this request (or a variation of it) so I can think about formatting this one last time and then never again.

eszlari:
> Shotcut’s style is the same as “Qt” in Qt Creator, and Qt is not listed or it might be exactly the same as another.

Maybe just copy their file: https://github.com/qt-creator/qt-creator/blob/master/.clang-format

ddennedy:
Brian please redo this with clang-format and that configuration file Pete showed us.

bmatherly:
> Brian please redo this with clang-format and that configuration file Pete showed us.

Done. See what you think.

Not sure if the formatting broke the build. That would be strange. But if we like this, I can figure it out. I would also make the target return an error so that we can use it for pipeline style check.

ddennedy:
I suggest to add `SortIncludes: Never` to try to fix the build. Changing these is dangerous considering some are dependent upon on macros.

bmatherly:
> I suggest to add SortIncludes: Never to try to fix the build. Changing these is dangerous considering some are dependent upon on macros.

I think that in an ideal world the order would not matter. For this case, I fixed the missing include.

If there is a case where the order is very important, I hope it will be documented. As part of that documentation process, the author can add comments to instruct clang-format to ignore that section of code:

```
// clang-format off
// foo must proceed bar for compilation to work
#include foo.h
#include bar.h
// clang-format on
```

ddennedy:
>  I would also make the target return an error so that we can use it for pipeline style check.

Considering different versions of clang-format gives different results given my recent commit, this is not a good idea. I am using version 14 that comes in Pop!_OS (Ubuntu) 22.04

bmatherly:
> Considering different versions of clang-format gives different results given my recent commit, this is not a good idea. I am using version 14 that comes in Pop!_OS (Ubuntu) 22.04

I added a target called clang-format-check that returns an error.
https://github.com/mltframework/mlt/blob/master/CMakeLists.txt#L530
But I did not tie it into the build automation because of this reason you mention.

I installed 5 versions of clang-format simultaneously. It is not hard to install different versions. Eventually, I will do more research to see if newer versions can be made compatible with older versions to adding/changing some of the configuration options. In the mean time, We can just run clang-format version 14 occasionally in case someone forgets to run it before they commit.

eszlari:
@bmatherly I guess it would make sense to switch the Shotcut source to clang-format too?

bmatherly:
> @bmatherly I guess it would make sense to switch the Shotcut source to clang-format too?

Probably eventually. But I want to make sure it is serving us well for MLT before I do that. I am disappointed that clang-format changes it's formatting from one release to the next. Shotcut already has astyle and it has been more consistent.

Issue #877, opened 2023-02-16
Make build work with -Wall -Werror


Tjoppen:
Currently the build spews warnings, which makes it difficult to catch new warnings. On gcc (and maybe clang?) -Wall and -Werror should be added and all warnings fixed, and any exceptions to this should be explicitly marked in C(XX)FLAGS.

The following addition to CMakeLists.txt improves the situation and makes explicit what warnings currently exist, some of which are quite serious, for example -Wno-restrict:

```cmake
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  set(GCC_FLAGS "-Wall -Werror -Wno-unused-variable -Wno-unused-function -Wno-switch -Wno-parentheses -Wno-sequence-point -Wno-deprecated-declarations -Wno-return-type -Wno-restrict -Wno-unused-but-set-variable -Wno-format -Wno-format-truncation")
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GCC_FLAGS}")
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${GCC_FLAGS} -Wno-discarded-qualifiers")
endif ()
```

bmatherly:
Using your proposal as an inspiration, I have a proposed resolution here: https://github.com/mltframework/mlt/pull/893

Please test it and comment.

bmatherly:
A version of this is committed now.

PR #893, opened 2023-03-26
Warnings as errors

bmatherly:
Proposed resolution for https://github.com/mltframework/mlt/issues/877

In this PR, I only enabled -Wall and -Werror when built for debug. I am open minded about making it always apply.

I think that deprecation warnings should always be ignored. We will always have deprecation warnings in avformat and it make it hard to see other messages.


src/modules/avformat/link_avdeinterlace.c
```diff
@@ -62,6 +62,13 @@ static mlt_image_format validate_format(mlt_image_format format)
     case mlt_image_yuv420p10:
     case mlt_image_yuv444p10:
         ret_format = format;
+        break;
+    case mlt_image_none:
+    case mlt_image_movit:
+    case mlt_image_opengl_texture:
+    case mlt_image_invalid:
+        ret_format = mlt_image_yuv422;
```
ddennedy, line 70: Why `mlt_image_yuv422` here and not `mlt_image_rgba` per my recent fix in b365ce645 ?

src/modules/plusgpl/filter_rotoscoping.c
```diff
@@ -327,7 +327,7 @@ static int filter_get_image(mlt_frame frame,
     mlt_properties frame_properties = MLT_FRAME_PROPERTIES(frame);
     int b_width = mlt_properties_get_int(frame_properties, "meta.media.width");
     int b_height = mlt_properties_get_int(frame_properties, "meta.media.height");
-    if (b_height == 0) {
+    if (b_width == 0 || b_height == 0) {
```
ddennedy, line 330: I was not confident about this change, but b_width and b_height are actually not used for anything.

ddennedy:
In general, it is safe to merge pending these 2 small things.

src/modules/avformat/link_avdeinterlace.c
bmatherly: This change follows the existing behavior. Previously, it would have returned the value set on line 55 above. My goal was resolve the warnings and not speculatively optimize. When I did the initial implementation, FFMpeg always seemed to want to convert to YUV for the deinterlacers. So I thought that might avoid some conversions. Maybe for Movit/OpenGL we should have FFMpeg do the YUV to RGBA conversion. I can make that change if you think that will be best.

ddennedy: No, don’t change at this time. Oversight on my review.

bmatherly:
Do you have an opinion on setting -Wall -Werror for all builds or only Debug builds? Currently it is only enabled for Debug builds.

src/modules/plusgpl/filter_rotoscoping.c
bmatherly: Done

ddennedy:
I think it should be on for all builds. There are several presets now defined by cmake plus unspecified, any of which a contributor might have used leading to a surprise.

PR #896, opened 2023-03-31
Fix compilation: parameter type mismatch

j-b-m:
Fix our PPA compilation:
./src/modules/avformat/factory.c:35:17: error: type of ‘link_swresample_init’ does not match original declaration [-Werror=lto-type-mismatch]
./src/modules/avformat/link_swresample.c:340:10: note: type mismatch in parameter 2
./src/modules/avformat/link_swresample.c:340:10: note: ‘link_swresample_init’ was previously declared here


ddennedy:
Is the PPA working again, and this is everything? We want to know if to keep this open for possibly more.

Issue #899, opened 2023-04-10
Huge memory consumption of mlt-7.14.0 runs oom with 08:40 minutes project on 64+8 GB laptop


Yamakuzure:
System : Gentoo Linux
CPU    : Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
GPU    : Quadro T2000
RAM    : 64GB + 8GB swap
Driver : 515.105.01
CUDA   : 11.7
Qt      : 5.15.8
MLT    : 7.14.0
Shotcut: 22.12.21
ffmpeg : 5.1.2

Main MLT file: (The project)
playlists : 8
producers : 21
chains    : 458
tractors  : 5

Render MLT file: (To render the project as a track for additional filters)
playlists  : 3
producers : 3
chains     : 0
tractors   : 1

--------

Symptoms in Shotcut-22.12.21 with MLT-7.14.0:

- Main project needs about 5 minutes to load.
- The render project takes about 15 minutes to load.
- Rendering breaks off around 68% with oom message.
- Reaching this point took about 25 minutes

`top` reported at about 62%:

````
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
14968 sed       23   3  100,8g  50,2g 722056 S  1088  80,5 134:36.87 melt-7
````

At that time shotcut had spawned 5,258 threads and melt-7 had spawned 1,436 threads

--------

Behavior after downgrading to shotcut-22.09.23 with MLT-7.12.0:

- Main project loads in ~3 seconds
- The render project loads in under 1 second.
- Rendering finishes successfully after 00:23:04

`top` reported at about 62%:

````
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
23266 sed       23   3   38,0g   9,0g 833456 S  1138  14,4 123:38.82 melt-7
````

At that time shotcut had spawned 714 threads and melt-7 had spawned 523 threads.

--------

Yesterday I highjacked https://github.com/mltframework/shotcut/issues/1391 because I thought it was shotcut's fault due to the long loading times and sluggish behavior. But after I checked `top` it more likely seems to be an mlt issue.

I have not yet tried shotcut-22.12.21 with mlt-7.12.0, though.

--------

_Edit_ : I now have tested Shotcut-22.12.21 with MLT-7.12.0: (I made a mistake and had to render again anyway)

Loading time of both main and render project is around 4-5 seconds.

[...]

PR #902, opened 2023-04-13
Fix two data races and a potential deadlock

Yamakuzure:
While testing melt-7 with shotcut for issue #899 (the idea was to check for thread leaks), the following two data races and one cycle in lock order issues were found:

* Lock order:
`producer_set_up_video()` versus `producer_get_frame()`
* Data race:
`consumer_worker_thread()` versus `first_unprocessed_frame()`
* Data race:
`mlt_events_fire()` versus `mlt_event_unblock()`

Details to the individual ThreadSanitizer reports are in the respective commit descriptions.

### DIFF
--- a/src/framework/mlt_consumer.c
+++ b/src/framework/mlt_consumer.c
@@ -42,6 +42,11 @@
  */
 pthread_mutex_t mlt_sdl_mutex = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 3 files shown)


ddennedy:
Thanks! These look good on my reading review. Will do some basic run testing soon prior to merge.

PR #903, opened 2023-04-14
Audio cache

bmatherly:
Proposed fix for https://github.com/mltframework/mlt/issues/899

Also implements https://github.com/mltframework/mlt/pull/744/ and resolves comments from that review

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -671,4 +671,8 @@ MLT_7.16.0 {
     mlt_deinterlacer_id;
     mlt_link_filter_init;
     mlt_link_filter_metadata;
+    mlt_cache_put_frame_audio;
+    mlt_cache_put_frame_image;
+    mlt_frame_clone_audio;
+    mlt_frame_clone_image;
 } MLT_7.14.0;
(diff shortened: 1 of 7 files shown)


ddennedy:
While not performing as well as v7.12 (e.g. 5.7 GB vs 4.9, or 4:33 render time vs. 3:34) this certainly brings the performance much closer. After all, in whole, these changes address a fundamental audio problem, and you do not always get something for nothing. The technical approach is good too. Maybe I would have made `mlt_cache_put_frame` return the cloned frame and then clear certain properties, but I like yours better.

bmatherly:
> While not performing as well as v7.12 (e.g. 5.7 GB vs 4.9, or 4:33 render time vs. 3:34) 

This last two commits to master should help with the memory usage - especially if you have any interlaced clips.

https://github.com/mltframework/mlt/commit/47971dc05532ecb4c8b63223076ab4076222fea4
https://github.com/mltframework/mlt/commit/d0063e6705b8ce2e4c4f67279fc811e43bf54790

bmatherly:
I am done  testing this. Let me know if anyone wants any changes or sees any problems.

PR #909, opened 2023-04-24
Fix #1428: Chroma bug for interlaced 4:2:0

bmatherly:
When converting YUV 4:2:0 interlaced to RGB, convert each field separately to avoid bleeding chroma from one field to the other.

This conforms to "Nominal vertical and horizontal sampling locations of 4:2:0 samples in top and bottom fields" as defined in ITU Rec. H.264. Section 6.2 and similar MPEG standards.

This does not conform to alternative chroma locations such as chroma_sample_loc_type_top_field and chroma_sample_loc_type_bottom_field defined in the VUI SEI message in H.264.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1623,21 +1623,62 @@ static void convert_image_rgb(producer_avformat self,
                               int dst_full_range)
 {
     int flags = mlt_get_sws_flags(width, height, src_pix_fmt, width, height, dst_pix_fmt);
-    struct SwsContext *context = sws_getContext(
-        width, height, src_pix_fmt, width, height, dst_pix_fmt, flags, NULL, NULL, NULL);
     uint8_t *out_data[4];
     int out_stride[4];
-    av_image_fill_arrays(out_data, out_stride, buffer, dst_pix_fmt, width, height, IMAGE_ALIGN);
-    // libswscale wants the RGB colorspace to be SWS_CS_DEFAULT, which is = SWS_CS_ITU601.
-    mlt_set_luma_transfer(context, self->yuv_colorspace, 601, self->full_range, 1);
-    sws_scale(context,
-              (const uint8_t *const *) frame->data,
-              frame->linesize,
-              0,
-              height,
-              out_data,
-              out_stride);
-    sws_freeContext(context);
+
+    if (src_pix_fmt == AV_PIX_FMT_YUV420P && frame->interlaced_frame) {
+        // Perform field-aware conversion for 4:2:0
+        int field_height = height / 2;
+        const uint8_t *in_data[4];
+        int in_stride[4];
+        struct SwsContext *context = sws_getContext(width,
+                                                    field_height,
+                                                    src_pix_fmt,
+                                                    width,
+                                                    field_height,
+                                                    dst_pix_fmt,
+                                                    flags,
+                                                    NULL,
+                                                    NULL,
+                                                    NULL);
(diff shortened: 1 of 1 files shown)


ddennedy:
Did you check if libswscale flag `SWS_FULL_CHR_H_INT` is the culprit? I see a comment in the code of `mlt_get_sws_flags()`:
> Does not cause chroma bleeding when used with SWS_POINT

What about filter `avcolor_space`?

bmatherly:
> Did you check if libswscale flag SWS_FULL_CHR_H_INT is the culprit? I see a comment in the code of mlt_get_sws_flags():

SWS_FULL_CHR_H_INT is unrelated. I tested those permutations with/without.

Handling the fields separate for conversion is the correct way to do it. FFMpeg gives an example in vf_scale:
https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/vf_scale.c#L887

Also, our own slice scaling method has been field aware for years - although it takes some inspection to realized how it handles fields separately:
https://github.com/mltframework/mlt/commit/aec8b753e0237cdf1c68f77b42eacfa81f558588

As a test, I forced the conversion to use the slice method and the chroma did not bleed between fields.

Do you know why we do not use the slice method for all conversions? Why do we have convert_image_yuvp() and convert_image_rgb()?

> What about filter avcolor_space? Doesn't it need this change too?

Yes. However, I think that as a practical matter, we would never ask it to convert from yuv420 because none of our other services operate on 420. So if 420 is converted, it is converted immediately in the producer.

I wonder if I should factor out the slice code and share it with the producer and avcolor_space filter. Unless there is a reason for them to work differently.

ddennedy:
> Do you know why we do not use the slice method for all conversions?

Never got around to it as it did not seem as beneficial as adding it to other things. FFmpeg 5 added support for slice-based processing when using the AVFrame-based API added to libswscale. This is used in the swscale filter, and I did try adding it to the avcolor_space filter in 47828d3, but I reverted it a couple of weeks later in 44f7eb1. I guess that caused some problem at a bad time while, or maybe related to the range conversions that were added shortly after the revert. When this became available in FFmpeg, I thought it would be better to utilize it there. However, with recent review around thread pools, since there is no thread pool for FFmpeg, now I do not.

> Why do we have convert_image_yuvp() and convert_image_rgb()?

`convert_image()` was getting too repetitive with respect to yuv handling after adding 10-bit. So, I simply refactored that function to call upon these 2 subroutines. They are slightly different, and the goal of refactoring is to not change behavior. They could probably have been consolidated the addition of a conditional around the destination colorspace. However, after this change, it is better they are different.

> I wonder if I should factor out the slice code and share it with the producer and avcolor_space filter. Unless there is a reason for them to work differently.

The decision to use vertical slices instead of horizontal was called into question recently since that is less CPU memory cache friendly. Should it use vertical slices for interlace and horizontal for progressive?

bmatherly:
> Never got around to it

That is a valid reason. And it is risky to change large amounts of code if you are not also prepared to throw a wide array of test cases at it.

> Should it use vertical slices for interlace and horizontal for progressive?

I never understood why it was decided to use vertical slices. Where there artifacts using horizontal slices? It seems the SWS uses horizontal slices for all cases.

I think it would be good to try to move to the AVFrame-based API eventually. But we should make sure we are committed to all the testing that would be required.

max-verem:
> I never understood why it was decided to use vertical slices. Where there artifacts using horizontal slices? It seems the SWS uses horizontal slices for all cases.

horizontal slicing generated artifacts on resulting image in all my tests (it seems SWS specific), that why vertical slicing was an only working solution to spread color space conversion among cpu cores.

PR #912, opened 2023-05-04
Fix full range override

bmatherly:
Partial fix for:
https://forum.shotcut.org/t/23-04-color-range-regression/38609

The "force_full_range" parameter was not being taken into account because it is only calculated when the producer is opened.

Additionally, workaround an issue that full range conversion does does not work for 420P->420P.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1586,6 +1586,16 @@ static int convert_image_yuvp(producer_avformat self,
                               int dst_full_range)
 {
     int result = self->yuv_colorspace;
+
+    // Workaround: swscale does not convert the range if the src & dst are both 420P
+    // Use the deprecated formats to force the conversion.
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -1586,6 +1586,16 @@ static int convert_image_yuvp(producer_avformat self,
                               int dst_full_range)
 {
     int result = self->yuv_colorspace;
+
+    // Workaround: swscale does not convert the range if the src & dst are both 420P
```
bmatherly, line 1590: I think this was recently uncovered because the deinterlace filter is no longer requesting 422 if deinterlacing is not required. I was not able to find a way to get the swscale API to convert between full and mpeg ranges if both source and destination are 420P.

ddennedy:
There were some things I did not like in this PR, but it very much inspired my commit 97dcafd that does not require a Shotcut change.

PR #917, opened 2023-05-14
Fix crash when changing video mode with count producer

bmatherly:
As reported here:
https://forum.shotcut.org/t/release-candidate-version-23-04-20/38515/22

My proposal for the short term is that we should not ever use movit for the encapsulated services in the count producer.

The count producer was written before the generic text filter. Now it should be refactored to use the text filter instead of two producers and a transition. This would make it more efficient and it would not have any use for movit conversions. But this larger change should wait until after a release.

### DIFF
--- a/src/modules/plus/producer_count.c
+++ b/src/modules/plus/producer_count.c
@@ -36,7 +36,6 @@
 #define OUTER_RING_RATIO 90
 #define INNER_RING_RATIO 80
 #define TEXT_SIZE_RATIO 70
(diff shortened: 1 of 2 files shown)


ddennedy:
OK, I guess when reloading this producer, it was not receiving the "loader-nogl" argument?

bmatherly:
> OK, I guess when reloading this producer, it was not receiving the "loader-nogl" argument?

Yeah. For example, the XML producer does not know to use the nogl loader. Maybe other places as well.

PR #922, opened 2023-06-02
transition_affine: Fix of memory overconsumption.

Seriy-MLGamer:
When I montaged my cool MLG-montage using Shotcut, I noticed that: when I zoom using "affineSizePosition" filter, memory consumption increases dramatically and it eats all of my RAM and swap file! And it's also very very slow!
But c'mon, that shouldn't eat so much. And I thought that this filter just allocates a huge buffer with the size of the zoomed frame instead of displaying buffer of normal size with applied scale. And there was exactly that thing.
Now we can zoom infinitely! And now it's also fast!
And thank you for the art freedom!

### DIFF
--- a/src/modules/plus/transition_affine.c
+++ b/src/modules/plus/transition_affine.c
@@ -399,6 +399,8 @@ static int transition_get_image(mlt_frame a_frame,
     mlt_image_format b_format = mlt_image_rgba;
     int b_width = mlt_properties_get_int(b_props, "meta.media.width");
     int b_height = mlt_properties_get_int(b_props, "meta.media.height");
+    int b_real_width = b_width;
+    int b_real_height = b_height;
     double b_ar = mlt_frame_get_aspect_ratio(b_frame);
     double b_dar = b_ar * b_width / b_height;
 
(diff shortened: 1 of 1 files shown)


ddennedy:
It used to work the way you have it, but it was changed for a reason: image quality. The scaling quality of affine transform is very poor compared to  a dedicated image scalar such as libswscale.

ddennedy:
Rejected because your change simply circumvents a bunch of logic (without removing the code) and regresses image quality. Also, if the source image is very large (we have a bug report here where someone is using an 21600x10800 image), then your approach also uses a large amount of memory. When working with video, some things will inherently require more memory without making a compromise somewhere else. That might mean you need more memory, limit project resolution, or limit the amount of zoom to work within your system constraints.

Seriy-MLGamer:
We need to fix this if we want to make high quality **free** software. Proprietary video editors don't have issues with this **basic** thing. But I need freedom.
I've tested this right now with interpolation. Quality is fine, interpolation works.
Ok, I can remove some redundant code. But please, merge this pull request or fix that issue somehow else.
Can you reopen it?

PR #929, opened 2023-07-15
Use proper deallocation function for AVCodecContext

ravemo:
[FFmpeg documentation for reference.](https://www.ffmpeg.org/doxygen/trunk/group__lavc__core.html#gaf4daa92361efb3523ef5afeb0b54077f)
This fixes a memory leak in Shotcut that consumed all my memory when editing projects with a large number of cuts.


ddennedy:
Thank you

PR #932, opened 2023-07-23
Fix #931 Blank producer in playlist does not have audio normalization

### LINKED ISSUE #931: Blank producer in playlist does not have audio normalization
Example command:
melt color:red in=0 out=25 -blank 25 color:blue in=0 out=25 -consumer avformat:test.mp4

The command will fail after processing frame 26 because the audio can not be converted to the requested format.

bmatherly:
* Add new blank producer
* Use the blank producer in the playlist so that blank frames have normalization filters attached.

One side effect of this is that a playlist must be initialized with a profile. Some of the normalization filters depend on it.


src/framework/mlt_playlist.h
```diff
@@ -71,7 +71,7 @@ typedef struct playlist_entry_s playlist_entry;
 struct mlt_playlist_s
 {
     struct mlt_producer_s parent;
-    struct mlt_producer_s blank;
+    struct mlt_producer_s blank; // Deprecated
```
ddennedy, line 74: This comment should have 3 slashes so it is picked up by doxygen for API docs.

src/modules/core/producer_blank.yml
```diff
@@ -0,0 +1,12 @@
...
+identifier: blank
+title: Blank
+version: 1
+copyright: Meltytech, LLC
+license: LGPLv2.1
+language: en
+tags:
+  - Video
+  - Audio
+description: A blank frame generator.
```
ddennedy, line 12: "blank" is too redundant in this documentation. What is the audio/video user experience of "blank" - black silence? Also, since this line is not a sentence it should not end with a period.

src/modules/core/producer_blank.c
```diff
@@ -0,0 +1,69 @@
...
+#include <framework/mlt_producer.h>
+
+#include <stdlib.h>
+
+static int producer_get_frame(mlt_producer self, mlt_frame_ptr frame, int index)
+{
+    // Generate a frame
+    *frame = mlt_frame_init(MLT_PRODUCER_SERVICE(self));
+
+    if (*frame != NULL) {
```
ddennedy, line 30: No need to compare with NULL  - not consistent with the majority of the code.

src/modules/core/producer_blank.c
```diff
@@ -0,0 +1,69 @@
...
+    free(producer);
+}
+
+mlt_producer producer_blank_init(mlt_profile profile,
+                                 mlt_service_type type,
+                                 const char *id,
+                                 char *arg)
+{
+    mlt_producer self = calloc(1, sizeof(struct mlt_producer_s));
+    if (self != NULL && mlt_producer_init(self, NULL) == 0) {
```
ddennedy, line 58: No need to compare with NULL and zero.

ddennedy (changes requested):
I think `mlt_playlist_init()` should be deprecated. `Playlist::Playlist()` depends on it, but it could call `mlt_playlist_new(nullptr)` as I am not sure we can remove that constructor.

src/modules/core/producer_blank.yml
bmatherly: I agree that the word is generic. But I am reusing the word as it is already defined in the framework:
https://github.com/mltframework/mlt/blob/b2787f9069004bb89af8d3d15d7761b024671197/src/framework/mlt_producer.c#L227

Here is a proposal for a more detailed description. Let me know what you think:

> A "Blank" producer returns frames that are transparent and silent. These frames are suitable for creating blank spaces in a playlist. mlt_producer_is_blank() will return 1 on this producer.

ddennedy: "transparent" is not accurate because it is only considered transparent when the image format is rgba. Let's just say silent white. Otherwise, I like it's reference to `mlt_producer_is_blank`.

src/framework/mlt_playlist.h
bmatherly: done

src/modules/core/producer_blank.yml
bmatherly: done

src/modules/core/producer_blank.c
bmatherly: done

src/modules/core/producer_blank.c
bmatherly: done

bmatherly:
> I think mlt_playlist_init() should be deprecated. Playlist::Playlist() depends on it, but it could call mlt_playlist_new(nullptr) as I am not sure we can remove that constructor.

Done

ddennedy (approved):

PR #933, opened 2023-07-23
Ensure qtblend/cairoblend pass the correct frame properties

j-b-m:
When using an optimization that bypassed the get_image method on the a_frame in qtblend/cairoblend transitions, the "top_field_first" property and some other were not passed from between the 2 frames of the transition.
This broke field order and caused interlaced rendering to always be bottom field first.

### DIFF
--- a/src/framework/mlt_tractor.c
+++ b/src/framework/mlt_tractor.c
@@ -345,11 +345,6 @@ mlt_producer mlt_tractor_get_track(mlt_tractor self, int index)
     return mlt_multitrack_track(mlt_tractor_multitrack(self), index);
 }
 
-const char *mlt_tractor_passthrough_properties()
-{
-    return "progressive,progressive,distort,colorspace,full_range,force_full_luma,top_field_first,color_trc";
-}
-
 static int producer_get_image(mlt_frame self,
                               uint8_t **buffer,
                               mlt_image_format *format,
@@ -386,7 +381,7 @@ static int producer_get_image(mlt_frame self,
     mlt_properties_set_int(properties, "format", *format);
     mlt_properties_set_double(properties, "aspect_ratio", mlt_frame_get_aspect_ratio(frame));
     // Pass all required frame properties
-    mlt_properties_pass_list(properties, frame_properties, mlt_tractor_passthrough_properties());
+    mlt_properties_pass_list(properties, frame_properties, "progressive,distort,colorspace,full_range,force_full_luma,top_field_first,color_trc");
 
     mlt_properties_set_data(properties,
                             "movit.convert.fence",
--- a/src/framework/mlt_tractor.c
+++ b/src/framework/mlt_tractor.c
@@ -345,6 +345,11 @@ mlt_producer mlt_tractor_get_track(mlt_tractor self, int index)
     return mlt_multitrack_track(mlt_tractor_multitrack(self), index);
 }
 
+const char *mlt_tractor_passthrough_properties()
+{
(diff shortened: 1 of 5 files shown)


src/framework/mlt_tractor.c
```diff
@@ -345,6 +345,11 @@ mlt_producer mlt_tractor_get_track(mlt_tractor self, int index)
     return mlt_multitrack_track(mlt_tractor_multitrack(self), index);
 }
 
+const char *mlt_tractor_passthrough_properties()
+{
+    return "progressive,progressive,distort,colorspace,full_range,force_full_luma,top_field_first,color_trc";
```
bmatherly, line 350: Progressive is in this list twice.

src/framework/mlt_tractor.c
```diff
@@ -345,6 +345,11 @@ mlt_producer mlt_tractor_get_track(mlt_tractor self, int index)
     return mlt_multitrack_track(mlt_tractor_multitrack(self), index);
 }
 
+const char *mlt_tractor_passthrough_properties()
```
bmatherly, line 348: Since this is a public function, it will require documentation. I am not sure what the documentation should say.  In fact, since this function is such a nuanced detail, I think that we should not add it. We can just duplicate the properties string everywhere it needs to be used.

src/framework/mlt_tractor.c
```diff
@@ -380,25 +385,9 @@ static int producer_get_image(mlt_frame self,
...
-                           mlt_properties_get_int(frame_properties, "full_range"));
-    mlt_properties_set_int(properties,
-                           "force_full_luma",
-                           mlt_properties_get_int(frame_properties, "force_full_luma"));
-    mlt_properties_set_int(properties,
-                           "top_field_first",
-                           mlt_properties_get_int(frame_properties, "top_field_first"));
-    mlt_properties_set(properties, "color_trc", mlt_properties_get(frame_properties, "color_trc"));
+    // Pass all required frame properties
+    mlt_properties_pass_list(properties, frame_properties, mlt_tractor_passthrough_properties());
```
bmatherly, line 389: Duplicate the properties string here.

src/modules/frei0r/transition_frei0r.c
```diff
@@ -68,9 +68,13 @@ static int transition_get_image(mlt_frame a_frame,
         // Check if the alpha channel is entirely opaque.
         && mlt_image_rgba_opaque(images[1], *width, *height)) {
         if (invert)
+        {
             error = mlt_frame_get_image(a_frame, image, format, width, height, 0);
-        else
+        } else {
+            // Pass all required frame properties
+            mlt_properties_pass_list(a_props, b_props, mlt_tractor_passthrough_properties());
```
bmatherly, line 75: Duplicate the properties string here.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -188,6 +188,8 @@ static int get_image(mlt_frame a_frame,
             }
         }
         if (!hasAlpha) {
+            // Pass all required frame properties
+            mlt_properties_pass_list(properties, b_properties, mlt_tractor_passthrough_properties());
```
bmatherly, line 192: Duplicate the properties string here.

bmatherly (changes requested):
After looking at the patch in context, I suggest that we should not add the new function. Just duplicate the properties string wherever it is needed.

bmatherly (approved):

bmatherly:
LGTM

PR #934, opened 2023-07-25
GPS text: added grade, cadence, temperature + some small updates

dany123:
I have a quick note about the re-formatting rules: in the long if..else if chains with multiple lines of code inside each if block it absolutely needs a newline somewhere between each if block otherwise the entire chain looks like a huge blob of text and finding something is harder than it needs to be (see the code below line 281 in the first diff below). 
For short blocks it does look nice and cozy but those are readable no matter how you write them.

### DIFF
--- a/src/modules/qt/filter_gpstext.cpp
+++ b/src/modules/qt/filter_gpstext.cpp
@@ -191,23 +191,25 @@ static void gps_point_to_output(mlt_filter filter,
     }
 
     /* for every keyword: we first check if the "RAW" keyword is used and if so, we print the value read from file (or --)
-	   then we check if there's a format keyword and apply the necessary conversion/format.
-	   The format must appear after the main keyword, RAW and format order is not important (strstr is used)
-	*/
+       then we check if there's a format keyword and apply the necessary conversion/format.
+       The format must appear after the main keyword, RAW and format order is not important (strstr is used)
+    */
     if (!strncmp(keyword, "gps_lat", strlen("gps_lat")) && crt_point.lat != GPS_UNINIT) {
         if (strstr(keyword, "RAW")) {
             if (raw.lat == GPS_UNINIT)
                 return;
             snprintf(gps_text, 10, "%3.6f", raw.lat);
-        } else
+        } else {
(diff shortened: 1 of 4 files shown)


ddennedy:
> looks like a huge blob of text

Maybe this is a sign that you need to refactor into more functions. You can also add comment lines as headings for sections of code perhaps followed by an empty line.

dany123:
Comments to group some of these does sound a pretty good workaround; I'm not sure refactoring further would be a good idea, the code would be too scattered (in this particular case).

src/modules/qt/filter_gpstext.cpp
```diff
@@ -252,7 +254,31 @@ static void gps_point_to_output(mlt_filter filter,
             snprintf(gps_text, 4, "%s", bearing_to_compass(raw.bearing));
         } else
             snprintf(gps_text, 4, "%s", bearing_to_compass(crt_point.bearing));
-    } else if (!strncmp(keyword, "gps_vdist_up", strlen("gps_vdist_up"))
+    }
+	else if (!strncmp(keyword, "gps_cadence", strlen("gps_cadence")) && crt_point.cad!=GPS_UNINIT) {
+		if (strstr(keyword, "RAW")) {
```
ddennedy, line 259: This is formatted badly. We do not use tabs any more for indent. We never really wanted expressions crammed together (spaces before and after operators). Please reformat it using `clang-format` version 14.

src/modules/qt/filter_gpstext.cpp
```diff
@@ -252,7 +254,31 @@ static void gps_point_to_output(mlt_filter filter,
...
+			if (raw.cad == GPS_UNINIT) return;
+			snprintf(gps_text, 10, "%.0f", raw.cad);
+		}
+		else
+			snprintf(gps_text, 10, "%.0f", crt_point.cad);
```
ddennedy, line 264: One thing clang-format does not enforce is matching usage of curly braces. I prefer that if the `if` block has curly braces so should each other block such as `else` here.

ddennedy (changes requested):

PR #936, opened 2023-08-01
FIx install of yml file for audioseam filter

j-b-m:
The audioseam yml file currently doesn't install because the CMakeLists.txt is broken

### DIFF
--- a/src/modules/core/CMakeLists.txt
+++ b/src/modules/core/CMakeLists.txt
@@ -75,46 +75,6 @@ set_target_properties(mltcore PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${MLT_MODULE_
 install(TARGETS mltcore LIBRARY DESTINATION ${MLT_INSTALL_MODULE_DIR})
 
 install(FILES
-  consumer_multi.yml
-  filter_audiomap.yml
-  filter_audiowave.yml
-  filter_autofade.yml
-  filter_box_blur.yml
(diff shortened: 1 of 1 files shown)


src/modules/core/CMakeLists.txt
```diff
@@ -75,46 +75,6 @@ set_target_properties(mltcore PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${MLT_MODULE_
...
-  producer_melt.yml
-  producer_noise.yml
-  producer_timewarp.yml
-  producer_tone.yml
-  transition_composite.yml
-  transition_luma.yml
-  transition_matte.yml
-  transition_mix.yml
-  loader.dict
-  loader.ini
```
bmatherly, line 118: Should loader.dict and loader.ini stay in the list? Or do they get installed some other way?

ddennedy: Thanks Brian those do need to be installed, and this commit broke that. And these are much more critical data files than YAML! Can you fix it please @j-b-m ? I’m on vacation.

ddennedy:
Those are target dependencies and make the files appear in IDEs such as Qt Creator. There is a separate install command where these are not included. I also checked last daily (macos) build log here in Actions, and see there is no longer lines showing these two files get installed. Fix it!!

j-b-m:
Fixed, sorry for this.

PR #941, opened 2023-08-22
add astream and vstream properties to avformat producer

ddennedy:
I am still testing this, and there are Shotcut changes required to take advantage of it that shows it basically working. But I want to give you a chance to review it. This is usable from the command line using `astream=1` on the melt command line, for example, to choose the second audio stream (much easier than figuring out or guessing the absolute stream index when you do not know the order of streams - or in the case of a Google Pixel video where the first stream is a data stream). The Pixel video breaks exporting with a proxy in Shotcut because the absolute stream indexes between the source and proxy do not match since the proxy transcode process drops the data stream.

I added this private Pixel video to our shared S3 folder in `Google Pixel` folder. I also added a video with 2 video streams and 2 audio streams (as well as some data streams, as created by a GoPro MAX 360 video camera) to a folder `multiple video and audio streams`.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -2830,9 +2830,10 @@ static int video_codec_init(producer_avformat self, int index, mlt_properties pr
     return self->video_index > -1;
 }
 
-static int pick_video_stream(producer_avformat self, int absolute_index)
+static int pick_video_stream(producer_avformat self)
 {
     mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
+    int absolute_index;
 
     if (self->video_format && mlt_properties_get(properties, "vstream")) {
         // Get the relative stream index
@@ -2883,7 +2884,7 @@ static void producer_set_up_video(producer_avformat self, mlt_frame frame)
                       0);
         context = self->video_format;
     }
-    index = pick_video_stream(self, index);
+    index = pick_video_stream(self);
 
     // Exception handling for video_index
     if (context && index >= (int) context->nb_streams) {
@@ -3563,10 +3564,11 @@ static int audio_codec_init(producer_avformat self, int index, mlt_properties pr
     return self->audio_codec[index] && self->audio_index > -1;
 }
 
-static int pick_audio_stream(producer_avformat self, int absolute_index)
+static int pick_audio_stream(producer_avformat self)
 {
     AVFormatContext *context = self->audio_format;
     mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
+    int absolute_index;
 
     if (context && mlt_properties_get(properties, "astream")) {
         // Get the relative stream index
@@ -3656,7 +3658,7 @@ static void producer_set_up_audio(producer_avformat self, mlt_frame frame)
         context = self->audio_format;
         self->probe_complete = 0;
     }
-    index = pick_audio_stream(self, index);
+    index = pick_audio_stream(self);
 
     // Update the audio properties if the index changed
     if (context && self->audio_index > -1 && index != self->audio_index) {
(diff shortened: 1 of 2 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -2792,6 +2830,27 @@ static int video_codec_init(producer_avformat self, int index, mlt_properties pr
     return self->video_index > -1;
 }
 
+static int pick_video_stream(producer_avformat self, int index)
```
bmatherly, line 2833: Suggestion to rename "index" to "absolute_index" for clarity.

src/modules/avformat/producer_avformat.c
```diff
@@ -2806,8 +2865,10 @@ static void producer_set_up_video(producer_avformat self, mlt_frame frame)
     // Fetch the video format context
     AVFormatContext *context = self->video_format;
 
-    // Get the video_index
-    int index = mlt_properties_get_int(properties, "video_index");
+    // Get the video stream index
+    int index = mlt_properties_get(properties, "vstream")
+                    ? mlt_properties_get_int(properties, "vstream")
```
bmatherly, line 2870: Should this be:

```
absolute_stream_index(self->video_format,
                                      AVMEDIA_TYPE_VIDEO,
                                      mlt_properties_get_int(properties, "vstream")
```

?

src/modules/avformat/producer_avformat.c
```diff
@@ -3511,13 +3640,10 @@ static void producer_set_up_audio(producer_avformat self, mlt_frame frame)
...
-    // Get the audio_index
-    int index = mlt_properties_get_int(properties, "audio_index");
-
-    // Handle all audio tracks
-    if (self->audio_index > -1 && mlt_properties_get(properties, "audio_index")
-        && !strcmp(mlt_properties_get(properties, "audio_index"), "all"))
-        index = INT_MAX;
+    // Get the audio index
+    int index = mlt_properties_get(properties, "astream")
+                    ? mlt_properties_get_int(properties, "astream")
```
bmatherly, line 3645: Should this  be absolute?

bmatherly (changes requested):
I did not download/test this change. But the approach looks sound to me. Some minor nits for your consideration. Feel free to ignore.

src/modules/avformat/producer_avformat.c
ddennedy: agreed, done

src/modules/avformat/producer_avformat.c
ddennedy: No, at this point `index` is not qualified as relative or absolute. And to convert relative to absolute it uses the AVFormatContext, which could be NULL. So, at this point, we only need to see if the stream is disabled to skip opening the context. Otherwise, the context is reopened if needed, and then it can use `pick_video_stream()` to convert a relative stream index to absolute.

src/modules/avformat/producer_avformat.c
ddennedy: Unknown, same as above comment.

src/modules/avformat/producer_avformat.c
```diff
@@ -2792,6 +2830,27 @@ static int video_codec_init(producer_avformat self, int index, mlt_properties pr
     return self->video_index > -1;
 }
 
+static int pick_video_stream(producer_avformat self, int absolute_index)
```
bmatherly, line 2833: The absolute_index parameter for this function is not necessary. The value passed in by the caller is never used. It will be overwritten by the IF or the ELSE  before it is used.

ddennedy: Correct, an artifact of refactoring and debugging. Changed.

bmatherly (approved):

PR #942, opened 2023-09-01
Fix audio index broken for audio only files

j-b-m:
The recent introduction of aStream/vStream broke playback of audio only files.
This patch seems to make it work as usual.


ddennedy:
thanks

PR #943, opened 2023-09-03
On Chain init, don't unnecessarily fetch source frame

j-b-m:
Currently, on Chain initialization, MLT always fetches a frame from the source producer. While this is ok when creating a new Chain, it is problematic for us on project load.

When attempting to load a project file from xml, like:
std::unique_ptr<Mlt::Producer> xmlProd(new Mlt::Producer(profile, "xml-string", playlist_string);

(where playlist_string is the .mlt project xml data)
Simply attempting to create this producer will try to fetch a frame for each avformat producer that is included in a chain, taking a long time and eating a lot of memory in the process.

My proposal would be to avoid fetching a frame when the meta properties have already been set (as is the case in Kdenlive project files). This makes projects open a lot faster.

### DIFF
--- a/src/framework/mlt_chain.c
+++ b/src/framework/mlt_chain.c
@@ -128,8 +128,10 @@ void mlt_chain_set_source(mlt_chain self, mlt_producer source)
         // Save the native source producer frame rate
         base->source_profile = mlt_profile_clone(mlt_service_profile(MLT_CHAIN_SERVICE(self)));
         mlt_frame frame = NULL;
-        mlt_service_get_frame(MLT_PRODUCER_SERVICE(source), &frame, 0);
-        mlt_frame_close(frame);
+        if (!mlt_properties_exists(source_properties, "meta.media.frame_rate_num") || !mlt_properties_exists(source_properties, "meta.media.frame_rate_den")) {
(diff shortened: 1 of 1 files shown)


bmatherly:
This is good timing. I was just thinking about this optimization yesterday. I am willing to help refine and test this change.

This would be a good time to replace the get_frame() with a call to probe().

Can you please try
1) Replace get_frame() with probe()
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_producer.h#L147
2) Perhaps add the same optimization in avformat
https://github.com/mltframework/mlt/blob/master/src/modules/avformat/producer_avformat.c#L3815
I am not sure how the optimization check should interact with the "probe_complete" flag. Maybe probe_complete can be set on init if the meta.meda properties are detected.

bmatherly:
Actually, I just remembered this prior change:
https://github.com/mltframework/mlt/commit/643c86494ed2d5b4f2384d55d49c4fdb6910966c

We do not want to risk that an image might be created outside of the image thread. 

I approve this  change as-is.

There is still an optimization that could be made for the probe() function. But that can be done outside of this change.

PR #944, opened 2023-09-03
Fix a possible crash in avformat probe

bmatherly:
Probe was accessing the cached avformat context from a store pointer (self->child) rather than through the cache. This can result in accesssing released memory if the cache deleted the context.

The context pointer should not be saved at all and should only be accessed through the cache system.

This change includes a rewrite of the video index change detection so that probe() will not have to access the avformat context at all.

This also adds an optimization that if metadata is loaded from XML, and nothing else has changed, then the probe can be skipped.

In my testing, this resolves the crash reported here:
https://forum.shotcut.org/t/beta-version-23-09-now-available-to-test/40558/3

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -96,7 +96,6 @@ struct producer_avformat_s
     int video_index;
     int64_t first_pts;
     atomic_int_fast64_t last_position;
-    int probe_complete;
     int video_seekable;
     int seekable; /// This one is used for both audio and file level seekability.
     atomic_int_fast64_t current_position;
@@ -1505,12 +1504,13 @@ static int pick_av_pixel_format(int *pix_fmt, int full_range)
 
 static void property_changed(mlt_service owner, producer_avformat self, char *name)
 {
-    if (self && name) {
+    if (self && name && self->parent) {
+        mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
         if (!strcmp("color_range", name)) {
             if (self->video_codec
                 && !av_opt_set(self->video_codec,
                                name,
-                               mlt_properties_get(MLT_PRODUCER_PROPERTIES(self->parent), name),
+                               mlt_properties_get(properties, name),
                                AV_OPT_SEARCH_CHILDREN)) {
                 if (self->full_range != (self->video_codec->color_range == AVCOL_RANGE_JPEG)) {
                     self->full_range = self->video_codec->color_range == AVCOL_RANGE_JPEG;
@@ -1518,7 +1518,6 @@ static void property_changed(mlt_service owner, producer_avformat self, char *na
                 }
             }
         } else if (!strcmp("force_full_range", name) || !strcmp("set.force_full_luma", name)) {
-            mlt_properties properties = MLT_PRODUCER_PROPERTIES(self->parent);
             if (self->full_range != mlt_properties_get_int(properties, name)) {
                 self->full_range = mlt_properties_get_int(properties, name);
                 self->reset_image_cache = 1;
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -3741,7 +3744,6 @@ static int producer_get_frame(mlt_producer producer, mlt_frame_ptr frame, int in
     // If cache miss
     if (!self) {
         self = calloc(1, sizeof(struct producer_avformat_s));
-        producer->child = self;
```
ddennedy, line 3744: Do you recall why this was removed? This was trying to reconnect these in the same manner as `mlt_producer_init(producer, self)` does in `producer_avformat_init()`. It seems like the correct thing to do even tho a brief analysis tells me this may never be used.

src/modules/avformat/producer_avformat.c
```diff
@@ -3814,33 +3816,31 @@ static int producer_get_frame(mlt_producer producer, mlt_frame_ptr frame, int in
...
+    if (mlt_properties_get_int(properties, "_probe_complete")) {
         return error;
     }
 
+    if (!mlt_properties_exists(properties, "_probe_complete")) {
+        // If metadata was loaded from XML, no need to probe
+        int video_in_use = mlt_properties_get_int(properties, "vstream") > -1;
+        if (video_in_use && mlt_properties_exists(properties, "meta.media.progressive")) {
+            return error;
+        } else if (!video_in_use && mlt_properties_exists(properties, "meta.media.0.codec.name")) {
```
ddennedy, line 3831: On a Google Pixel phone (there is shared file for this in our S3 folder), there is no "meta.media.0.codec.name" property. In fact, stream 0 is elided because it is not audio or video, and avformat producer only cares about those types of streams. One might be inclined to test for a non-stream-specific meta.media property such as "meta.media.sample_aspect_num". However, that will not be present on audio-only file (`!video_in_use`). So, I think the only reliable way is to get "meta.media.nb_streams" and enumerate the streams and look for "meta.media.%d.stream.type".

ddennedy (changes requested):

src/modules/avformat/producer_avformat.c
bmatherly: It is never used - and it can never be used because the pointer might be deleted by the cache if the cache limit is hit. It essentially  becomes a dangling pointer when the cache destroys it. It was being used by probe because I did not realize the cache could delete it when I wrote the probe function. So I remove this line so that nobody else will try to use it.

src/modules/avformat/producer_avformat.c
bmatherly: Actually, the presence of  "meta.media.nb_streams" might be good enough to indicate that the metadata properties have been loaded by XML. It is interesting to know that all the audio metadata is loaded when the file is opened. It is only video metadata that might require a frame to be pulled, or maybe even an image. I will change it to use nb_streams and do some testing.

ddennedy: >  "meta.media.nb_streams" might be good enough to indicate that the metadata properties have been loaded by XML

Yes, that is true. The only exception would be a file with no usable audio or video, which is not only very rare, but I think would cause no harm here anyways.

bmatherly: I did make the change to nb_streams. But I am not able to trigger that optimization with Shotcut. I notice that we do not store meta.media properties in the XML. Is that on purpose? It seems that all clips will at least be opened when an MLT project is opened (not not necessarily decoding a frame). Maybe there is an opportunity for a future optimization.

ddennedy: Yes, the meta properties are omitted primarily to keep project file size smaller and make it easier to read when necessary. Now, I worry what will happen with proxy editing if they are included. They may initially be based on either the source or the proxy initially, but often converted to the proxy while editing. Then, when switching to the source, they will still represent the proxy, which might be different than source. However, they are recomputed when the producer is actually opened (in first get_frame when using avformat-novalidate). I do not know how that interplay's with chain's needs. That needs analysis and testing.

> It seems that all clips will at least be opened when an MLT project is opened

The point of the avformat-novalidate producer a long while ago was to avoid that, but that does not need meta properties provided. The non-chain producer approach only needs a few things like length and stream indices.  It seems your switch to chain for everything unwound that optimization.

ddennedy: It looks like chain is mainly interested in frame rate, and proxy frame rate is the same as source. The problem that may exist when relying on this info is kind of the same that exists today: when source is variable frame rate (or something strange with many digits of precision) the proxy is a constant frame rate approximate. I will do the testing around this to make the decision for Shotcut.

bmatherly: Your analysis is correct. A chain wants to know the producers "native" frame rate so that it can have the highest temoral resolution if it is doing speed tricks. Maybe there is a different (or less complex) way for the chain to know the frame rate without the needs to probe or pull frames.

PR #947, opened 2023-09-06
Fix #945 MLT::Chain leaking memory

### LINKED ISSUE #945: MLT::Chain leaking memory
Hi all,

While debugging an important memory usage in Kdenlive, I came upon a strange behavior and I think MLT::Chain is not properly closing the producer's data. I tested with the simple code below.

This is a simple loop, creating a producer, asks for an image then closing the producer. If I run this app, I get a rather stable memory usage between 380-410M.

But if I replace Mlt::Producer and instead use Mlt::Chain to build the producer, the app quickly consumes several Gigabytes of RAM. I am not sure exactly how to fix this, but any help is welcome as this issue prevents users to open recent project files because of the memory usage.

```
std::unique_ptr<Mlt::Repository> repository(Mlt::Factory::
Mlt::Profile profile("atsc_1080p_25");
int ow = 320;
int oh = 240;
mlt_image_format format = mlt_image_rgba;
for (int i = 0; i < 100; i++) {
    Mlt::Producer prod(profile, "avformat", "/media/video/4kSample.MP4");
    Mlt::Frame *frame = prod.get_frame();
    const uchar *imagedata = frame->get_image(format, ow, oh);
    delete frame;
}
```

bmatherly:
Ensure the source producer is closed. Memory can lean if it is never closed.

### DIFF
--- a/src/framework/mlt_chain.c
+++ b/src/framework/mlt_chain.c
@@ -394,6 +394,8 @@ void mlt_chain_close(mlt_chain self)
         for (i = 0; i < base->link_count; i++)
             mlt_link_close(base->links[i]);
         free(base->links);
+        mlt_producer_close(base->source);
+        mlt_properties_close(base->source_parameters);
         mlt_profile_close(base->source_profile);
         free(base);
         self->parent.close = NULL;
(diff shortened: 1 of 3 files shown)


src/framework/mlt_chain.c
```diff
@@ -395,6 +395,7 @@ void mlt_chain_close(mlt_chain self)
             mlt_link_close(base->links[i]);
         free(base->links);
         mlt_profile_close(base->source_profile);
+        mlt_producer_close(base->source);
```
ddennedy, line 398: And what about source_parameters?

bmatherly: Good catch. I fixed that with a force push.

PR #956, opened 2023-09-24
Avoid cusp and overshoot in keyframe_type_smooth

bmatherly:
Change the catmull-rom interpolation method to be "centripetal". The current implementatoin uses the "unity" method. The centripetal method ensures there will be no cusps in the spline. Additionally, modify the interpolation to avoid overshoot.

These changes ensure that a line will interpolate directly from one value to another without allowing the interpolated values to reverse along the way.

This implementation changes the existing keyframe_type_smooth behavior. If we prefer to make a new keyframe type, instead, then this patch can easily be adapted to add a new type. The old  behavior can be easily restored by changing alpha to 0 in the new function.

Here is a comparison of the old vs new implementation - as well as with linear. This comparison uses the keyframe sequence from the unit test that was added: "50~=0; 60~=100; 100~=110; 150~=200; 200~=110; 240~=100; 260~=50; 300~=200; 301~=10; 350~=11"
image
In the case of the last overshoot, I do not think that the user would have expected the value to actually go negative. In this case, the user would probably be forced to "coerce" the value by adding some linear keyframes in between. But this proposed method does not overshoot - which I think would be more in line with expectations.

### DIFF
--- a/src/framework/mlt_animation.c
+++ b/src/framework/mlt_animation.c
@@ -3,7 +3,7 @@
  * \brief Property Animation class definition
  * \see mlt_animation_s
  *
- * Copyright (C) 2004-2021 Meltytech, LLC
+ * Copyright (C) 2004-2023 Meltytech, LLC
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -26,6 +26,8 @@
 #include "mlt_properties.h"
 #include "mlt_tokeniser.h"
 
+#include <float.h>
+#include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -55,6 +57,10 @@ struct mlt_animation_s
 };
 
 static void mlt_animation_clear_string(mlt_animation self);
+static int interpolate_item(mlt_animation_item item,
+                            mlt_animation_item p[],
+                            double fps,
+                            mlt_locale_t locale);
 
 /** Create a new animation object.
  *
@@ -81,8 +87,7 @@ void mlt_animation_interpolate(mlt_animation self)
         animation_node current = self->nodes;
         while (current) {
             if (!current->item.is_key) {
-                double progress;
-                mlt_property points[4];
+                mlt_animation_item points[4];
                 animation_node prev = current->prev;
                 animation_node next = current->next;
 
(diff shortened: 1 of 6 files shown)


src/framework/mlt_animation.c
```diff
@@ -973,3 +963,289 @@ void mlt_animation_clear_string(mlt_animation self)
...
+ *  at which the value will be interpolated. The value for "keyframe_type" determines which
+ *  interpolation will be used.
+ * \param p a sequential array of 4 animation items. The frame value for item must lie between the
+    frame values for p[1] and p[2].
+ * \param fps the frame rate, which may be needed for converting a time string to frame units
+ * \param locale the locale, which may be needed for converting a string to a real number
+  * \return true if there was an error
+ */
+
+static int interpolate_item(mlt_animation_item item,
```
bmatherly, line 1067: Why did I move this function? The centripetal method requires the position of all 4 points for non-uniform points. So I need the position of each property that is being used to interpolate. Instead of passing the positions to the property interpolation function, I just used the animation items (which already have the frame position included). I could move it back and add more parameters if we prefer that - it would need a new name to not break ABI.

src/framework/mlt_property.c
```diff
@@ -1213,6 +1215,7 @@ static inline double catmull_rom_interpolate(double y0, double y1, double y2, do
 
 /** Interpolate a new property value given a set of other properties.
  *
+ * \deprecated
```
bmatherly, line 1218: I propose to deprecate this function and keep the new function private in mlt_animation so that we can change it if necessary. I doubt any applications are using it. It just needed to be public so that mlt_animation could call it.

src/tests/test_properties/test_properties.cpp
```diff
@@ -1273,12 +1273,12 @@ private Q_SLOTS:
         QCOMPARE(p.anim_get_color("key", 50).g, 255);
         QCOMPARE(p.anim_get_color("key", 50).b, 171);
         QCOMPARE(p.anim_get_color("key", 50).a, 223);
-        QCOMPARE(p.anim_get_color("key", 60).a, 237);
-        QCOMPARE(p.anim_get_color("key", 70).a, 239);
+        QCOMPARE(p.anim_get_color("key", 60).a, 223);
```
bmatherly, line 1276: I think this is a great example of unexpected behavior for the existing implementation. The keyframe on both sides of these frames has a value of 223. But the algorithm would add "ripple" and interpolate different values for intermediate frames.

Merlimau:
Beside this new function: Would it be possible that the user can create his own curve type with Bezier and save it for further use?

bmatherly:
Have a look at this idea: https://github.com/bmatherly/mlt/commit/ade3ce7d04b2a45ab27bdcc3e62df01bea5b4c57
It will be very easy to make more keyframe types. For backwards compatibility, we could just make this a new keyframe type. Or, we could go with your idea to make the algorithm tied to some environment variable/version detection.

Merlimau:
> For backwards compatibility, we could just make this a new keyframe type.

I would go with a new type, for backwards compatibility.

Merlimau:
> we could go with your idea to make the algorithm tied to some environment variable/version detection.

I’m not sure if you understand my idea of “Bezier” keyframe types. Here a screenshot from Blender with interpolation (Bezier) and easing modes, this is what I meant by additional keyframe types.  
image

ddennedy:
@Merlimau This is not about Bezier, and that is out-of-scope for this PR. MLT has no capability to support keyframes that have parameters.

bmatherly:
Superseded by https://github.com/mltframework/mlt/pull/957

PR #957, opened 2023-09-30
New keyframe types

bmatherly:
Add new keyframe types: smooth_natural and smooth_tight . Both new types use centripetal catmull-rom interpolation.
The centripetal method ensures there will be no cusps in the spline. Additionally, modify the interpolation to avoid overshoot. These changes ensure that a line will interpolate directly from one value to another without allowing the interpolated values to reverse along the way.

The previous "smooth" type is aliased to "smooth_loose" and maintained for backwards compatibility.

This chart shows the difference in the curves:
image

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -682,7 +682,7 @@ MLT_7.18.0 {
     mlt_audio_free_data;
 } MLT_7.16.0;
 
-MLT_7.20.0 {
+MLT_7.22.0 {
   global:
     mlt_property_is_color;
     mlt_property_is_numeric;
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -681,3 +681,10 @@ MLT_7.18.0 {
   global:
     mlt_audio_free_data;
 } MLT_7.16.0;
+
+MLT_7.20.0 {
+  global:
+    mlt_property_is_color;
+    mlt_property_is_numeric;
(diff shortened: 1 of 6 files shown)


src/framework/mlt.vers
```diff
@@ -681,3 +681,10 @@ MLT_7.18.0 {
   global:
     mlt_audio_free_data;
 } MLT_7.16.0;
+
+MLT_7.20.0 {
```
ddennedy, line 685: Version 7.20.0 was just release. Please change this to `MLT_7.22.0` but it is still based on `MLT_7.18.0` below since 7.20.0 has no new symbols.

j-b-m:
I just want to say a big :+1:  from my side as I have seen many cases where the current smooth interpolation produces very inappropriate interpolation, and the new mlt_keyframe_smooth_natural seems to fix these cases, so thanks!

bmatherly:
Added new keyframe types implementing popular easing functions:

Ease In
image

Ease Out
image

Ease In Out
image

I'm going to start to integrated this into Shotcut. Maybe I will find some bugs along the way. But I think this is ready for a thorough review if anyone wants to take the time.

src/framework/mlt.vers
bmatherly: Done

frdbr:
Thanks for adding this Brian, all MLT users are going to be very happy. Cheers.

Issue #769, opened 2022-02-10
Timewarp Producer Issue: Ignores framerate settings in mlt_profile, reports wrong frame on callback.


Eoin-ONeill-Yokai:
I've been trying to use a timewarp producer to allow for fast-forward and slow-motion features in my current application and noticed some issues after hot swapping my existing producer (`avformat`) with a `timewarp` producer wrapping my original producer.

Firstly, I noticed that my callback `consumer-frame-show` no longer respects my current frame rate settings (i.e. when I change my profile to 4 frames per second, I still get 24 frame show callbacks a second that I have to deal with) which is causing some parts of my application to not respect the target framerate settings. I've even tried to reassign the profile after changes to frame-rate, but it would still always callback 24 times a second.

Secondly, I want to know if there's a good way to discern the position of the frame relative to the internal producer (the `avformat` producer) from within the `consumer-frame-show` callback. I've noticed that the timewarp obviously changes the frame's "position" to report the position of the timewarp producer. While I could perhaps discern the position using a simple multiplication (timewarp warp_speed value * frame position) I can't help but wonder if there's a better way to do this.

ddennedy:
> when I change my profile to 4 frames per second, I still get 24 frame show callback

Really? That seems impossible. The consumer fires this event according to its framerate, which can be selected by using a profile. It does not know about the timewarp producer. It sounds like your profile was adversely affected. (Not sure why you want 4 fps?!?) I do not believe we see this problem in Shotcut. Did you try using Shotcut?

bmatherly:
> when I change my profile to 4 frames per second, I still get 24 frame show callbacks a second that I have to deal with

Are you using that new profile to initialize all of your services? Or just the producer?


You should be able to query the "original_position" property on the frame:

https://github.com/mltframework/mlt/blob/9c3bb7d997a78fed3142c2ef95988623d21dce15/src/framework/mlt_frame.c#L183

ddennedy:
Also, if you change the profile after you start the consumer, it will not respond to those changes. You need to restart the consumer.

ddennedy:
> You should be able to query the "original_position" property on the frame:

See also `mlt_frame_original_position()`

Eoin-ONeill-Yokai:
@ddennedy 

> It sounds like your profile was adversely affected. (Not sure why you want 4 fps?!?) I do not believe we see this problem in Shotcut.

[...]

PR #961, opened 2023-11-29
Upstreaming Krita Animation Plugin

emmetoneillpdx:
Hey there, I'm Emmet. I work on animation systems and other stuff for [Krita](https://krita.org/). 

A few months back we merged a [pretty big refactor to our animation playback system that uses MLT in order to improve our audio-video synchronization when playing back animation files](https://invent.kde.org/graphics/krita/-/merge_requests/1323). We recently shipped this new animation system in [Krita 5.2](https://krita.org/en/krita-5-2-release-notes/) and I'm happy to report that it seems to be working relatively well so far for our users who are animating to audio. MLT being designed with tight audio-video synchronization in mind seems to have made it a good fit for Krita's needs. :+1: 

Anyway... I'm here to see if we can possibly _upstream our small module as well as some minor changes_ that we've had to make to get things building and working across our various target platforms (Linux, Windows, Mac, and Android). 

(There are a couple small memory fixes in there too.)


src/framework/mlt_repository.c
```diff
@@ -336,12 +336,14 @@ void *mlt_repository_create(mlt_repository self,
...
-    mlt_properties_close(&self->parent);
-    free(self);
+	mlt_properties_close( self->consumers );
+	mlt_properties_close( self->filters );
+	mlt_properties_close( self->links );
+	mlt_properties_close( self->producers );
+	mlt_properties_close( self->transitions );
+	mlt_properties_close( &self->parent );
+
+	free( self );
```
ddennedy, line 346: Please remove this superficial change that does not conform to our clang-format standard.

src/modules/krita/producer_krita.c
```diff
@@ -0,0 +1,224 @@
...
+
+
+
+typedef struct
+{
+	int frame_start;
+	int frame_end;
+	int limit_enabled;
+	double speed;
+	mlt_producer producer_internal;
```
ddennedy, line 43: Please format all files in the `krita` subfolder using `clang-format` 14 and our `.clang-format` file.

src/modules/krita/producer_krita.c
```diff
@@ -0,0 +1,224 @@
...
+
+	return r;
+}
+
+static void producer_property_changed( mlt_service owner, mlt_producer self, mlt_event_data event_data)
+{
+	const char *name = mlt_event_data_to_string(event_data);
+	if (!name) return;
+
+	if (strcmp(name, "start_frame") || strcmp( name, "end_frame" )){
```
ddennedy, line 117: `strcmp()`  returns 0 for equality, which is boolean false in C. So, all string equality tests using it should be preceded with `!` like `if (!strcmp(name, "start_frame") ...`. I am not sure how you get this working for you as-is!

src/modules/krita/producer_krita.yml
```diff
@@ -0,0 +1,33 @@
+schema_version: 0.3
+type: producer
+identifier: krita
+title: Krita
+version: 1
+copyright: Eoin O'Neill, Emmet O'Neill
```
ddennedy, line 6: Consider adding a `description` because it is not clear at all the purpose of this producer. At a casual glance, one thinks it can read a `.kra` file to produce video frames in MLT!

src/modules/krita/producer_krita.yml
```diff
@@ -0,0 +1,33 @@
...
+creator: Eoin O'Neill, Emmet O'Neill
+license: LGPLv2.1
+language: en
+tags:
+  - Video
+  - Audio
+parameters:
```
ddennedy, line 13: The `resource` property is not documented. It gets a `argument: yes` attribute. Also, the `speed` property is not documented.

ddennedy (changes requested):

src/modules/krita/producer_krita.c
```diff
@@ -0,0 +1,224 @@
+/*
+ * filter_freeze.c -- simple frame freezing filter
```
bmatherly, line 2: Fix this line to match the file name

src/modules/krita/producer_krita.c
bmatherly: Alternately, consider removing this function altogether, remove the properties from the pdata struct and just read the properties directly when you need them. The private data struct is more commonly used to store internal state, not externally set properties.

bmatherly (changes requested):
Great job putting this together!

In addition to the code comments, I have these general comments.

Please explain why a new producer is needed and you can't modify one of these similar producers to add the features you need:
https://github.com/mltframework/mlt/blob/master/src/modules/core/producer_hold.c
https://github.com/mltframework/mlt/blob/master/src/modules/kdenlive/producer_framebuffer.c
https://github.com/mltframework/mlt/blob/master/src/modules/core/producer_timewarp.c

Why should  there be a new Krita module for this producer? Does it do something specific for Krita that no other application would want? Generally, the modules exist because the services in the module have some common dependency. Having a module named "krita" would imply that the services depend on some library from Krita. If that is not the case, I recommend to move this service into an existing module.

src/framework/mlt_repository.c
```diff
@@ -336,12 +336,13 @@ void *mlt_repository_create(mlt_repository self,
...
-    mlt_properties_close(self->transitions);
-    mlt_properties_close(&self->parent);
-    free(self);
+	mlt_properties_close( self->consumers );
+	mlt_properties_close( self->filters );
+	mlt_properties_close( self->links );
+	mlt_properties_close( self->producers );
+	mlt_properties_close( self->transitions );
+	mlt_properties_close( &self->parent );
+	free( self );
```
ddennedy, line 345: Indentation of this is still wrong; it uses 4 spaces now instead of tabs. Our `.clang-format` file covers that.

src/modules/krita/producer_krita.c
```diff
@@ -0,0 +1,231 @@
...
+    return error;
+}
+
+static int producer_get_frame(mlt_producer producer, mlt_frame_ptr frame, int index)
+{
+    private_data *pdata = (private_data *) producer->child;
+
+    const int position = mlt_producer_position(pdata->producer_internal);
+    if (is_valid_range(pdata) && is_limit_enabled(pdata)) {
+        mlt_properties_set_position(MLT_PRODUCER_PROPERTIES(pdata->producer_internal),
```
bmatherly, line 98: Does mlt_producer_seek() work here instead of modifying a private property directly?

src/modules/krita/producer_krita.c
```diff
@@ -0,0 +1,231 @@
...
+        pdata->limit_enabled = 1;
+        pdata->speed = 1.0;
+
+        // Initialize property values for start / end frames.
+        mlt_properties_set_int(producer_properties, "start_frame", pdata->frame_start);
+        mlt_properties_set_int(producer_properties, "end_frame", pdata->frame_end);
+        mlt_properties_set_int(producer_properties, "limit_enabled", pdata->limit_enabled);
+        mlt_properties_set_double(producer_properties, "speed", pdata->speed);
+
+        // Create a producer for the clip using the false profile.
```
bmatherly, line 202: This copied comment is incorrect because this is not using a "false" profile. It uses the same profile as the parent. You could probably just remove the comment to avoid future confusion.

bmatherly:
> @bmatherly Initially, I had the same thought. But then I realized this is something special for media playback for Krita since A) it uses the abnormal loader producer to skip normalizing filters and B) it does some special things for audio speed handling.

The code is the same as the timewarp producer but with the false profile trick removed and a start/end parameter added. I expect the start/end parameters could be avoided by using in/out points. But I do not know exactly how this is integrated into Krita. I agree this is unique to Krita and unlikely to be useful to anyone else. Thanks for humoring me on the discussion.

ddennedy:
> I expect the start/end parameters could be avoided by using in/out points.

I had that thought as well! Also there is general producer ignore _points property that probably could have been used.

bmatherly:
>  I kinda like giving them their own little isolated space to work in. As for a separate module, I would turn it off for our Shotcut app bundles. But it should default on so distributions have it ready for their Krita builds.

I agree with this.

ddennedy:
Please create a PR for only the fixes (or let me if you want me to do them). I doubt we will accept this new producer. You can use or extend timewarp, or you can copy your producer into Krita and register it at runtime using mlt_repository_register() - no need to register metadata as that is purely supplemental.

bmatherly:
I applied the two memory leak fixes from [f829e56](https://github.com/mltframework/mlt/pull/961/commits/f829e5661f29d101e7aba1a609aa9b3f6bce76e6) to master

emmetoneillpdx:
Hey again @ddennedy and @bmatherly, sorry for the delayed response to your comments. I've been a bit preoccupied!

> But then I realized this is something special for media playback for Krita since A) it uses the abnormal loader producer to skip normalizing filters and B) it does some special things for audio speed handling. The updated description in the yml mentions that.

That's my fault! I wanted to be concise but I probably wasn't detailed enough in my description of the context of this PR... 

Let me elaborate a bit. :)

## Background

Krita's a Qt/KDE program for art and animation, and until our latest refactor we had been using QtMultimedia as the back end for our animation audio playback (like an animator might use when animating a scene to dialogue or background music). QtMultimedia had a pretty basic interface that didn't allow for fine control of samples or any simple mechanism for synchronization. Naturally, this lead to a lot of bugs involving audio/animation synchronization, both inside Krita and in the final exported animation video files. 

So we started looking for other options, and looking at KDenLive we discovered MLT which, having an emphasis on frame-by-frame AV synchronization, seemed like it might be a good candidate.

## Krita and MLT

Since the refactor, we're now using MLT to drive playback of animations when audio is attached, so the producer in this patch is at the center of that.

While our main goal with Krita and MLT was to improve synchronization of animation frames and attached audio, we had a few other requirements for keeping feature parity with our previous system:

- Dynamic playback speed adjustment. (Krita has a "speed" slider that can be adjusted as the animation is playing back, and so we need to have speed changes apply in real time.")
- Sub-region looping. (If an animator selects a slice of frame somewhere in their animation, Krita will loop playback to within that sub-region.)
- Pull and push modes. (We wanted to be able to surrender control to the audio system for better synchronization during playback and rendering, but also push some number of samples/frames out while "scrubbing" and editing.)

_To handle playback speed_, we were [originally looking into using the `timewarp` producer](https://github.com/mltframework/mlt/issues/769), but we ran into problems doing it dynamically without rebuilding the pipeline. It's been... quite a long time... since we did that initial research, but iirc our conclusion was that the timewarp producer (at the time at least) was not made for dynamic speed adjustment and that we should use it as a template to do something similar while allowing for speed changes to be made during playback without rebuilding the pipeline.

_For sub-region looping_, we had originally wanted to implement a separate "filter" element/service which sat between the consumer(s) and producer and simply restricted (by modulo wrapping, for our purposes) the frame index within a given region (for example, whatever slice of frames the user had selected in Krita). We explored this but for whatever reason we just could not make it work, so ultimately we ended up merging it into our main producer (the `restrict_range` function now handles that).

(_Push and pull_ we've handled by using a pair of consumers. If there's a better and more dogmatic way of handling that please let me know. Though it's working alright so far!)

> Please create a PR for only the fixes (or let me know if you want me to do them). I doubt we will accept this new producer. You can use or extend timewarp, or you can copy your producer into Krita and register it at runtime using mlt_repository_register() - no need to register metadata as that is purely supplemental.

Alright, no worries. 

If timewarp can be made to work with real time speed changes in the way that we require, then I don't see any reason why wouldn't be able to just use that. Especially if we also have some mechanism to restrict/wrap/loop frames within a specific sub-region of indices.

Even if upstreaming our bespoke producer _as-is right now_ isn't something that you want to do, I think we'd be happy to find other, more general ways to incorporate the features we need into upstream MLT however possible. I'm not exactly an experienced C programmer (can you tell? :laughing:) and we ended up jumping somewhat blindly into MLT, but the results so far have been pretty solid and so I hope we can keep a good relationship and line of communication with you guys in upstream MLT. :)

Anyway, I'll look into timewarp again. I'm also definitely interested in learning more about `mlt_repository_register` if that can give us an alternative to maintaining/shipping a full fork, like we're doing now.

ddennedy:
> I applied the two memory leak fixes from [f829e56](https://github.com/mltframework/mlt/pull/961/commits/f829e5661f29d101e7aba1a609aa9b3f6bce76e6) to master

Looks like you need to push them.

ddennedy:
> I'm also definitely interested in learning more about `mlt_repository_register`

1. Copy `producer_krita.c` to krita source tree
2. Put `extern "C" ` before `producer_krita_init()`
3. Add `producer_krita.c` to `CMakeLists.txt`
4. Add `extern "C" void* producer_krita_init(mlt_profile, mlt_service_type, const char*, const void*)` to `KisPlaybackEngineMLT.h`
5. After your call to `Mlt::Factory::init()` add `repository->register_service(mlt_service_producer_type, "krita", producer_krita_init);`
6. Add casts to fix C++ compile errors

I verified this works using Shotcut.

bmatherly:
> Looks like you need to push them.

Oops. Pushed now.

ddennedy:
Please make a new branch and PR with your win32 .def fixes, and put the new producer in your app.

PR #964, opened 2023-12-06
Add a loop range feature to MLT Producer

bmatherly:
This allows an application to request a producer loop between a start and end frame.

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -688,3 +688,8 @@ MLT_7.22.0 {
     mlt_property_is_numeric;
     mlt_property_is_rect;
 } MLT_7.18.0;
+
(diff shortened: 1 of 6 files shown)


ddennedy:
Is this redundant with the properties in, out, and eof?

bmatherly:
> Is this redundant with the properties in, out, and eof?

I tried to start with a variation of in/out/eof. But I didn't think of a way to make it work. If you change in and out, you inherently change the length of the clip. It also changes the meaning of the frame number that you requested to be the loop point.

I am willing to try out some other ideas if you think of another way to accomplish this.

ddennedy:
Do you still want to add this?

bmatherly:
No longer needed

PR #974, opened 2024-03-10
Avoid FAST_FAIL_RANGE_CHECK_FAILURE during preset enumeration

hmartinez82:
Fixes https://github.com/mltframework/shotcut/issues/1522


ddennedy:
It is not entirely clear what "FAST_FAIL_RANGE_CHECK_FAILURE" is as it is Windows-specific, but change is good and correct. Thanks!

hmartinez82:
It does seem to be Windows specific indeed and only when using MSVCRT. That function is full of SAL annotations in its parameter.

PR #983, opened 2024-05-11
Initial Implementation of LV2 module

joinlaw:
Can be tested with melt eg:

$ melt ~/path/to/audio/file.wav -attach "lv2.http<//kxstudio.sf.net/carla/plugins/audiogain_s" 5=0.3 -consumer avformat:processed.mp3 acodec=libmp3lame

(:) in http://kxstudio.sf.net/carla/plugins/audiogain_s replaced with (<) because melt and/or mlt doesn't accept (:) in plugin name and also uri should be prefiexed with "lv2."

currently only this extensions are supported:
http://lv2plug.in/ns/ext/buf-size#boundedBlockLength
http://lv2plug.in/ns/ext/urid#map
http://lv2plug.in/ns/ext/urid#unmap
http://lv2plug.in/ns/ext/options#options

https://lv2plug.in/pages/host-compatibility.html


src/modules/lv2/factory.c
```diff
@@ -0,0 +1,261 @@
+/*
+ * factory.c -- the factory method interfaces
+ * Copyright (C) 2003-2022 Meltytech, LLC
```
ddennedy, line 3: This new code file should only have the year 2024.

src/modules/lv2/factory.c
```diff
@@ -0,0 +1,261 @@
...
+#include <lv2/port-groups/port-groups.h>
+#include <lv2/port-props/port-props.h>
+#include <lv2/presets/presets.h>
+#include <lv2/resize-port/resize-port.h>
+#include <lv2/ui/ui.h>
+#include <lv2/worker/worker.h>
+
+#include <lilv/lilv.h>
+
+#ifdef GPL
```
ddennedy, line 42: Please avoid a multiple license situation. The code you borrowed for this had to for a special reason.

src/modules/lv2/factory.c
```diff
@@ -0,0 +1,261 @@
...
+    }
+#endif
+
+    return result;
+}
+
+MLT_REPOSITORY
+{
+
+#ifdef GPL
```
ddennedy, line 222: Here we can clearly see there are no MLT services without "GPL." So, if GPLv2 is required, then can be easily enforced in the top-level `CMakeLists.txt` and no need to clutter the code in this module with these conditionals.

src/modules/lv2/factory.c
```diff
@@ -0,0 +1,261 @@
...
+      s = calloc(1, strlen("lv2.") + strlen (desc->uri) + 1);
+
+      sprintf(s, "lv2.%s", desc->uri);
+
+      char *str_ptr = strchr(s, ':');
+      while (str_ptr != NULL)
+	{
+	  *str_ptr++ = '<';
+	  str_ptr = strchr(str_ptr, ':');
+	}
```
ddennedy, line 240: Regarding the plugin name, since it must be altered, please drop the broken scheme and just make it like "lv2//kxstudio.sf.net/carla/plugins/audiogain_s". Then, we can just say it is the schema-less URI. Notice I did not include a period after "lv2" because the "//" is enough delimiters but still following some URI grammar. Or, since we are breaking URI grammar use  "lv2.kxstudio.sf.net/carla/plugins/audiogain_s". Your choice between the two, but I would rather not hear from people trying to use "http:" either in support issues or as bug reports.

src/modules/lv2/filter_lv2.c
```diff
@@ -0,0 +1,279 @@
...
+        mlt_properties_set_data(properties,
+                                "lv2rack",
+                                lv2rack,
+                                0,
+                                (mlt_destructor) lv2_rack_destroy,
+                                NULL);
+
+        if (resource)
+	  // Load LV2 Rack XML file
+	  lv2_rack_open_file(lv2rack, resource);
```
ddennedy, line 62: Does this really support a "LV2Rack" file? I found there is an unpopular and abandoned "lv2rack" project, but JACK Rack never did support LV2. If it is legitimate, what GUI tool creates it? But also, in `factory.c` there is no LV2 rack XML file filter registered.

src/modules/lv2/filter_lv2.yml
```diff
@@ -0,0 +1,25 @@
+schema_version: 7.0
+type: filter
+identifier: lv2
+title: LV2
+version: 1
+license: GPLv2
+language: en
+url: http://www.lv2.org/
+creator: Dan Dennedy
```
ddennedy, line 9: Remove my name

src/modules/lv2/filter_lv2.yml
```diff
@@ -0,0 +1,25 @@
...
+description: Process audio using LV2 plugins.
+notes: >
+  Automatically adapts to the number of channels and sampling rate of the consumer.
+bugs:
+  - Some effects have a temporal side-effect that may not work well.
+
+parameters:
+  - identifier: resource
+    argument: yes
+    title: JACK Rack XML file
```
ddennedy, line 21: Does it really handle a JACK Rack XML file?

src/modules/lv2/producer_lv2.yml
```diff
@@ -0,0 +1,13 @@
+schema_version: 7.0
+type: producer
+identifier: ladspa
+title: LADSPA
```
ddennedy, line 4: Should be LV2

ddennedy (changes requested):
The approach of copying the jackrack module and modifying it is not acceptable. That brings in a lot of baggage. jackrack code base was chosen at a time when I needed to support a rack XML file made with its GUI tool for a customer. This was a time when there were only some toy GUI tools for MLT. As a bonus it added support for JACK. I later modified it to support a single named LADSPA plugin for Kdenlive and friends. This ought to be a cleaner implementation without all of that baggage. I guess you can get there by continuing to modify, but I am not in a situation where I want to go through many PR reviews to get there. Alternatively, can the existing `jackrack` module be extended to support LV2?

src/modules/lv2/filter_lv2.c
joinlaw: No it doesn't this is first time I heard about the "lv2rack" project and actually what I did as you know is to heavily modify jackrack module to create new module and this was copied and renamed in the process without me knowing for what it is used.

src/modules/lv2/filter_lv2.yml
joinlaw: No

joinlaw:
> The approach of copying the jackrack module and modifying it is not acceptable. That brings in a lot of baggage. jackrack code base was chosen at a time when I needed to support a rack XML file made with its GUI tool for a customer. This was a time when there were only some toy GUI tools for MLT. As a bonus it added support for JACK. I later modified it to support a single named LADSPA plugin for Kdenlive and friends. This ought to be a cleaner implementation without all of that baggage. I guess you can get there by continuing to modify, but I am not in a situation where I want to go through many PR reviews to get there. Alternatively, can the existing `jackrack` module be extended to support LV2?

At first I tried to extend the jackrack module and even put both LADSPA and LV2 plugins in the same plugin manager but I fail to get working code for months until I tried copying and I get here.

Actually I have to think about the approach and take time to be more familiar with filter design in mlt. The problem is it was hard for me to use same structure of plugin/process/manager so I have to heavily alter it and only `lock_free_fifo.c`, `lock_free_fifo.h` files copied without modification.

src/modules/lv2/factory.c
joinlaw: As we said `:` can't be used in filter name for some reason.

Not all lv2 URIs contain http:// for example this one don't:
urn:distrho:vitalium

and `<` is disallowed in URI sytax so it can't conflict I there is chance there is another plugin URI with similar name but different delimiter and mlt/melt accept it so I choose it.

but still this is ugly there should be another solution.

joinlaw:
with this commit [aba53e9](https://github.com/mltframework/mlt/pull/983/commits/aba53e9452ef79e26b6e0f840f7d0d3672098415) now parameters metadata is more accurate and not causing issues and be readonly

kdenwork

And still I don't get dropdown options (enumerations) to work.

I am investigating ways to integrate this module in `jackrack`.

joinlaw:
By now this feature is reimplemented as part of jackrack module and the old implementation moved to `lv2-module-old-way` https://github.com/joinlaw/mlt/tree/lv2-module-old-way branch.

src/modules/jackrack/filter_lv2.c
```diff
@@ -0,0 +1,279 @@
+/*
+ * filter_ladspa.c -- filter audio through LADSPA plugins
+ * Copyright (C) 2004-2018 Meltytech, LLC
```
ddennedy, line 3: Change year in copyright for all files. For modified files the range end year is 2024. For new files, the Meltytech copyright line should be only the year 2024.

ddennedy (changes requested):

src/modules/jackrack/lv2_plugin.c
```diff
@@ -272,8 +274,19 @@ static int lv2_plugin_instantiate(const LilvPlugin *plugin,
...
+  if (lv2context_can_ui != NULL)
+    {
+      /* Video editors and other hosts that support custom GUI should use mlt_environment_set ("lv2context_can_ui", "1")
+	 to inform mlt lv2 plugin manager and set UI features and extensions if not set.
+       */
+      if (lv2context_can_ui[0] == '1')
+	{
+	  //WIP: if support UI
+	}
+    }
```
joinlaw, line 287: For hosts (ie. Video Editors) that want to support LV2 Custom GUIs they should set the mlt environment variable "lv2context_can_ui" to "1" and the the lv2 plugin instance will be instantiated with LV2 UI features and extensions and then they can get "global_lv2_world" handle of LilvWorld* and and retrieve the UI URI and handles and create ui support with suil library (http://drobilla.net/software/suil.html).

This is the plan for supporting Custom GUIs.

ddennedy:
Do not force push to a PR that has a lot of changes and already had some review. When you force push I need to start over the review because I can not see what you changed from I left off. And when there is a lot of code, it makes the reviewer more reluctant to start over.

CMakeLists.txt
```diff
@@ -250,6 +251,10 @@ if(MOD_JACKRACK)
   list(APPEND MLT_SUPPORTED_COMPONENTS jackrack)
 endif()
 
+if(USE_LV2)
+  pkg_check_modules(lilv IMPORTED_TARGET lilv-0)
```
ddennedy, line 255: I get a compile error because I did not install a lv2 dev package, and the new option defaults to ON. I think we need to turn the option off when the pkg module fails.
```
/home/ddennedy/src/mlt/src/modules/jackrack/plugin_desc.h:38:10: fatal error: lv2.h: No such file or directory
   38 | #include <lv2.h>
      |          ^~~~~~~
```
This worked for me:
```
--- CMakeLists.txt
+++ CMakeLists.txt
@@ -253,6 +253,9 @@ endif()
 
 if(USE_LV2)
   pkg_check_modules(lilv IMPORTED_TARGET lilv-0)
+  if(NOT lilv_FOUND)
+    set(USE_LV2 OFF)
+  endif()
 endif()
 
 if(MOD_KDENLIVE)
```

Next, I installed `liblilv-dev` and ran `cmake -D USE_LV2=ON`, but now I get a compile errors:
```
/home/ddennedy/src/mlt/src/modules/jackrack/lv2_context.c:305:13: error: ‘saved_rack_destroy’ defined but not used [-Werror=unused-function]
  305 | static void saved_rack_destroy(saved_rack_t *saved_rack)
      |             ^~~~~~~~~~~~~~~~~~
/home/ddennedy/src/mlt/src/modules/jackrack/lv2_context.c:286:22: error: ‘saved_rack_new’ defined but not used [-Werror=unused-function]
  286 | static saved_rack_t *saved_rack_new(lv2_context_t *lv2_context, const char *filename, xmlDocPtr doc)
      |                      ^~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```
Debug builds (`CMAKE_BUILD_TYPE=Debug`) enable strict compilation options that PRs must pass.

ddennedy (changes requested):

src/modules/lv2/factory.c
ddennedy: The problem with `<` is that it is not friendly on the command line, and it is reserved character in XML. These are the 2 main protocols for MLT! How about `^`?

src/modules/jackrack/CMakeLists.txt
```diff
@@ -40,6 +45,13 @@ if(GPL AND TARGET PkgConfig::xml AND TARGET PkgConfig::glib AND ladspa_h_FOUND)
     target_sources(mltjackrack PRIVATE filter_jackrack.c)
     install(FILES filter_jackrack.yml DESTINATION ${MLT_INSTALL_DATA_DIR}/jackrack)
   endif()
+
+  if(USE_LV2 AND TARGET PkgConfig::lilv)
+    target_link_libraries(mltjackrack PRIVATE PkgConfig::lilv)
+    target_sources(mltjackrack PRIVATE filter_lv2.c producer_lv2.c lv2_context.c lv2_plugin.c lv2_process.c lv2_plugin_settings.c)
```
ddennedy, line 51: Please add the new header files to this as well as my IDE uses this to show the project files. You can see all of our CMakeLists.txt do that including other `target_sources` in the same file.

ddennedy:
Build target `clang-format-check` fails with many errors. I noticed problems immediately when I start editing files,  and `git diff` showed many unrelated changes because my IDE is setup to follow our clang-format rules and reformats some of the code. One of the biggest issues is using tab characters for indent. `clang-format` should be following our include `.clang-format` config file.

ddennedy:
Not wanting to be a total downer, I did run it (with my change to use `^`), and it is working including things like `melt -query` and serializing to XML with `melt -consumer xml  my.mp4 -attach lv2.http^//calf.sourceforge.net/plugins/Crusher 20=100`

```
...
    <filter id="filter0">
      <property name="mlt_service">lv2.http^//calf.sourceforge.net/plugins/Crusher</property>
      <property name="20">100</property>
    </filter>
...
```

I also ran the above for several frames through `valgrind`, and it did not complain about any memory problems in this new code. So, just these few things to tidy up, and it is ready!

src/modules/lv2/factory.c
joinlaw: yeah I got the point it would be good for xml and commandline and also (^) is in url forbidden characters so it should be fine.

https://url.spec.whatwg.org/#host-miscellaneous
http://www.faqs.org/rfcs/rfc1738.html

PR #991, opened 2024-06-04
Sparse intra only decode

### LINKED ISSUE #986: producer_avformat performs useless work when source is intra-only and consumer framerate is lower than the source's framerate
This may or may not be considered a bug, but with ffmpeg one can set `-thread_type frame -threads 16` to speed up intra-only decoding, especially for slow codecs like JPEG2000. `producer_libavformat.c` happily passes `thread_type` along, but this is purely detrimental. libavformat decodes multiple frames at once as instructed, but producer_libavformat only ever uses the first one. Playing more nicely with libavformat's threading would enable better handling of heavy to decode formats like JPEG2000. I see that recently there has been done some work to make melt would work better with intra-only, and supporting frame threading would be one way to make it work even better.

Tjoppen:
This fixes #986. It also brings the handling of rational numbers more in line with how ffmpeg does things, avoiding the use of floating point timestamp calculations.

With this change, if the output fps is lower than the source fps and the source is intra-only, then only packets that are actually necessary to decode are sent to the decoder. Frame threading is still done as before. This is most dramatic when decoding expensive formats like JPEG2000. If the output fps is the same or higher than the source fps then the behavior is as before.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -2324,6 +2324,8 @@ static int producer_get_image(mlt_frame frame,
             // We only deal with video from the selected video_index
             if (self->pkt.stream_index == self->video_index) {
                 int64_t pts = best_pts(self, self->pkt.pts, self->pkt.dts);
+                // default to decoding all intra-only frames from req_position to the future
+                int should_decode = int_position >= req_position;
                 if (pts != AV_NOPTS_VALUE) {
                     if (!self->video_seekable && self->first_pts == AV_NOPTS_VALUE)
                         self->first_pts = pts;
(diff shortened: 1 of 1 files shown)


ddennedy:
@Tjoppen  This makes things broken for me when reducing the frame rate of intra-only  video. Broken means the image is frequently not updating for each frame. I am testing with Ut Video and DNxHR. Also, when I add some logging in the new block to further determine `should_decode` at
```
                    if (self->pkt.duration != AV_NOPTS_VALUE) {
                        // compute the interval this intra-only packet covers
```
it is showing that it gets evaluated for non-intra-only too. Except  in the those cases, the image is updating. Where is the intra-only test condition?
My test video is 60000/1001 fps, and this works on intra-only with `melt -profile atsc_1080p_2997 ...` but not `melt -profile atsc_1080p_30 ...`

Also, the first commit to change calculations to use `AVRational` is [breaking frame accurate seeking](https://bugs.kde.org/show_bug.cgi?id=492692) as well as subsequent frame updates (not changing frame rate) in some formats such as MPEG-2.

CC: @j-b-m

PR #992, opened 2024-06-05
Fix double free crash

bmatherly:
If the caller frees the returned object (which they should), then a double free error will occur because the encapsulated mlt_profile will be freed by the Profile and by the origional service that owned it.

### DIFF
--- a/src/mlt++/MltService.cpp
+++ b/src/mlt++/MltService.cpp
@@ -117,7 +117,7 @@ Service *Service::consumer()
 
 Profile *Service::profile()
 {
-    return new Profile(mlt_service_profile(get_service()));
+    return new Profile(mlt_profile_clone(mlt_service_profile(get_service())));
 }
 
 mlt_profile Service::get_profile()


ddennedy:
Where is this other free? Services do not own the profile.

ddennedy:
`Mlt::FilteredProducer::detach()` constructs a `Profile` whose destructor frees the `mlt_profile` via `mlt_profile_close()` but the `Mlt::Profile` is not required due to a `Mlt::Producer` constructor that takes a `mlt_profile`:
```
--- src/mlt++/MltFilteredProducer.cpp
+++ src/mlt++/MltFilteredProducer.cpp
@@ -66,8 +66,7 @@ int FilteredProducer::detach(Filter &filter)
             Service *consumer = it->consumer();
             if (consumer->is_valid())
                 consumer->connect_producer(*producer);
-            Profile p(get_profile());
-            Producer dummy(p, "colour");
+            Producer dummy(get_profile(), "colour");
             dummy.connect_producer(*it);
             if (last->get_service() == it->get_service()) {
                 delete last;
```

bmatherly:
> Where is this other free? Services do not own the profile.

Presumably, whoever owns the mlt_profile that was passed to the constructor will either still try to use it (access freed memory) or free it (double free).

I wonder if we should deprecate Service::profile() to discourage its use. Mlt::Properties does not follow the same pattern as other Mlt classes. Other classes that take an mlt_properties based object in their constructor work properly because mlt_properties are reference counted.

PR #995, opened 2024-06-19
src/modules/core/transition_mix.c: Increase MAX_CHANNELS to 32

Tjoppen:
This matches filter_audiomap.c

Perhaps this should be a project-wide CMake option? Conform them for now.

### DIFF
--- a/src/modules/core/transition_mix.c
+++ b/src/modules/core/transition_mix.c
@@ -26,7 +26,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define MAX_CHANNELS (6)
+#define MAX_CHANNELS (32)
 #define MAX_SAMPLES (192000)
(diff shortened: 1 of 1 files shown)


bmatherly:
Did you consider an approach to change this service to remove the static definition and dynamically allocate and loop based on the number of actual channels? I wonder how hard that would be to do.

Tjoppen:
Oh and to address your comment in the other PR, this one has been tested and is known to work with filter_audiomap.c. Perhaps rather than doing what #996 does, we a common API for reporting channel limits could be introduced? That way there's a way to query it.

ddennedy:
> doesn't hurt

It increases memory usage on projects with multiple mix transitions. Using my example from the other PR, the project with 100 mix transitions needs 2.3 GiB RAM just for the mix transitions. Do your requirements permit a lower value such 8 - 16?

PR #996, opened 2024-06-19
Introduce project-wide MAX_CHANNELS in CMakeLists.txt

Tjoppen:
This is an alternative to PR #995

### DIFF
--- a/src/modules/core/transition_mix.c
+++ b/src/modules/core/transition_mix.c
@@ -26,7 +26,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define MAX_CHANNELS (6)
 #define MAX_SAMPLES (192000)
 #define SAMPLE_BYTES(samples, channels) ((samples) * (channels) * sizeof(float))
 #define MAX_BYTES SAMPLE_BYTES(MAX_SAMPLES, MAX_CHANNELS)
--- a/src/modules/plus/ebur128/ebur128.c
+++ b/src/modules/plus/ebur128/ebur128.c
@@ -398,12 +398,11 @@ void ebur128_get_version(int* major, int* minor, int* patch) {
   *patch = EBUR128_VERSION_PATCH;
 }
 
-#define VALIDATE_MAX_CHANNELS (64)
 #define VALIDATE_MAX_SAMPLERATE (2822400)
 
 #define VALIDATE_CHANNELS_AND_SAMPLERATE(err)                                  \
   do {                                                                         \
-    if (channels == 0 || channels > VALIDATE_MAX_CHANNELS) {                   \
+    if (channels == 0 || channels > MAX_CHANNELS) {                            \
       return (err);                                                            \
     }                                                                          \
                                                                                \
--- a/src/modules/rubberband/filter_rbpitch.cpp
+++ b/src/modules/rubberband/filter_rbpitch.cpp
@@ -38,8 +38,6 @@ typedef struct
     uint64_t out_samples;
 } private_data;
 
-static const size_t MAX_CHANNELS = 10;
-
 static int rbpitch_get_audio(mlt_frame frame,
                              void **buffer,
                              mlt_audio_format *format,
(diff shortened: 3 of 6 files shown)


Tjoppen:
I wasn't able to properly test that filter_ambisonic-decoder.cpp compiles due to its C++ API having changed (or Debian having an ancient version), but not removing MAX_CHANNELS in its leads to an earlier compilation failure as expected.

src/modules/plus/ebur128/ebur128.c
```diff
@@ -398,12 +398,11 @@ void ebur128_get_version(int* major, int* minor, int* patch) {
   *patch = EBUR128_VERSION_PATCH;
 }
 
-#define VALIDATE_MAX_CHANNELS (64)
```
bmatherly, line 401: I do not want to change this file because I want to keep it synchronized with the upstream project
https://github.com/jiixyj/libebur128

src/modules/core/transition_mix.c
```diff
@@ -26,7 +26,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define MAX_CHANNELS (6)
```
bmatherly, line 29: This change will result in higher heap allocation and higher loop counts in the code below. Rather than increasing the static value, should we change the code below to better handle a dynamic number of channels? I think we would just need to delay the heap allocations until after the audio has been received and the actual number of channels is known.

src/modules/rubberband/filter_rbpitch.cpp
```diff
@@ -38,8 +38,6 @@ typedef struct
     uint64_t out_samples;
 } private_data;
 
-static const size_t MAX_CHANNELS = 10;
```
bmatherly, line 41: I checked the rubberband library and it does not not have a limit on the channel count. I set this to 10 initially because that is the highest number of channels that I ever tested.

bmatherly (changes requested):
I do not feel strongly about this pull request or the alternate that was submitted. However, I would lean slightly toward our current approach which is to allow each service to define their own channel count limit based on practical limitation or real limitations passed down by other libraries. Also, if a service has a limit defined, it would be good if someone had tested the service at that limit - which would not be the case with this patch. Users could submit pull requests to increase the limit for a specific service when they have tested it at that limit.

src/modules/plus/ebur128/ebur128.c
Tjoppen: Is the limit in libebur128 defined in a publicly accessible header? If so then we should use that rather than defining our own.

src/modules/core/transition_mix.c
Tjoppen: Allocation isn't a problem thanks to overcommit. As for loop counts, they only increase of channels_a or channels_b is greater than 6, which is broken in master anyway at the moment.

Supporting arbitrarily large numbers of channels would be nice, but increasing the setting in CMakeLists.txt and allocating upfront is far easier and again basically free thanks to overcommit. The only snag is the allocation being done by calloc() which may cause the allocation to actually occur due to touching all of the memory, unless libc's calloc() is clever enough to map all pages to a single page of all zeroes.

Even FFmpeg has a static limit, from libavcodec/internal.h: #define FF_SANE_NB_CHANNELS 512U

src/modules/plus/ebur128/ebur128.c
Tjoppen: Just checked; it's in ebur128.c so the answer is no.

Tjoppen: Oh wait the whole of it is just incorporated as-is. Nevermind then. It seems it will successfully fail if the number of channels is greater than the limit.

src/modules/core/transition_mix.c
ddennedy: The main issue is that most NLE projects use multiple transition objects. In all MLT NLEs, not only there is a mix transition for every track but also every video transition to cross-fade audio. Then, consider a multitrack project with at least one video track with many A/V transitions. With this, the project requires much more memory. For example, let's say there are 100 mix transitions in the project, which is not unreasonable from my user support experience: before 439 MiB, after 4.6 GiB. So, I have to reject this.

PR #999, opened 2024-06-28
Fix OpenBSD support

brad0:
OpenBSD does not have LC_NUMERIC locale support.

### DIFF
--- a/src/framework/mlt_properties.c
+++ b/src/framework/mlt_properties.c
@@ -759,7 +759,7 @@ int mlt_properties_set(mlt_properties self, const char *name, const char *value)
 
             // Determine the value
             if (isdigit(id[0])) {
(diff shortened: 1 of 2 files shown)


ddennedy:
Please change this to draft, add all commits you need to fix the build, and then submit it for review (instead of numerous PRs).

brad0:
> Please change this to draft, add all commits you need to fix the build, and then submit it for review (instead of numerous PRs).

It's the last diff I have. I submitted multiple PRs as most projects want self contained separate diffs instead of one big diff.

ddennedy:
Yeah, well, "fix the OpenBSD" is self-contained IMO, and a PR can hold multiple commits. OK, thanks for the update!

brad0:
> Yeah, well, "fix the OpenBSD" is self-contained IMO, and a PR can hold multiple commits. OK, thanks for the update!

The commit message was poor, I admit that and I was going to update it to something better.

PR #1000, opened 2024-07-02
Add subtitle feed filter, overlay filter and producer.

bmatherly:
Add two new filters:

filter_subtitle_feed can add subtitle properties to a frame. The subtitle data can come from an SRT file, or as SRT string attached to the filter.

filter_subtitle can receive subtitle properties from a frame and burn them into the video

producer_subtitle is a convenience producer that puts a filter_subtitle on top of a color produced frame.

### DIFF
--- a/src/modules/plus/filter_subtitle.yml
+++ b/src/modules/plus/filter_subtitle.yml
@@ -1,0 +1,182 @@
+schema_version: 7.0
+type: filter
+identifier: subtitle
+title: Subtitle
+version: 1
+copyright: Meltytech, LLC
+license: LGPLv2.1
+language: en
+tags:
+  - Video
+description: Overlay subtitles onto the video
+
+parameters:
+  - identifier: resource
+    argument: yes
+    title: Resource
+    type: string
+    description: The path to a text file that contains SRT subtitles
+    required: no
+    readonly: no
+
+  - identifier: text
+    title: Text
+    type: string
+    description: >
+        A string that containes a complete SRT document.
+        This parameter is ignored if resource is set.
+    required: no
+    readonly: no
+
+  - identifier: feed
+    title: Text
+    type: string
+    description: >
+        A string that identifies the feed in the MLT Frame to overlay.
+        This parameter is ignored if resource or text is set.
+    required: no
+    default: 0
+    readonly: no
+
+  - identifier: geometry
+    title: Geometry
+    type: rect
+    description: A set of X/Y coordinates by which to adjust the text.
+    default: 20%/80%:60%x20%:100
+
+  - identifier: family
+    title: Font family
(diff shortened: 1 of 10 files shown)


src/modules/plus/subtitles/subtitles.h
```diff
@@ -0,0 +1,42 @@
+/*
```
bmatherly, line 1: This file is duplicated in Shotcut. We could move this into the MLT framework - maybe as "mlt_subtitle". But it does not have an object - it is just functions to read srt files.

Also, in order to move it into the MLT framework, I would need to rewrite it as C - which I am willing to do.

ddennedy: An SRT helper does not logically belong in the MLT framework. It could be a separate project, but then you would have more to manage. In case there are changes needed for Shotcut, I do not want to require a new version of MLT. Distros do not build Shotcut the way we do, and may use a slightly older version. They have more to manage. So, leave it duplicated as-is for now.

src/modules/plus/filter_subtitle.yml
```diff
@@ -0,0 +1,182 @@
...
+    type: integer
+    description: The weight of the font.
+    minimum: 100
+    maximum: 1000
+    default: 400
+    readonly: no
+    mutable: yes
+    widget: spinner
+
+  - identifier: fgcolour
```
ddennedy, line 91: I am not in a position to do a very detailed review, and I am going to merge this and let you fix things in master as needed. One thing I just noticed is that none of your parameters in YAML include `animation: yes` but I see keyframes in the screenshots of the Shotcut PR. (`mutable` is not the same thing as `animation`.)

bmatherly: > I am not in a position to do a very detailed review, and I am going to merge this and let you fix things in master as needed.

I think that is fine. This is just adding two services. So the regression risk is low. As long as you are in agreement with the concept/approach, we can fix things later.

> One thing I just noticed is that none of your parameters in YAML include animation: yes

That is a good observation and a common omission among all the filters that use the underlying "text" filter: gps_text, dynamictext, timer. The problem is that the underlying "text" filter might use QT or might fall back to Pango. The qtext filter has implemented animation, but the pango_text filter has not. So none of the YML for any of those services specifies "animation: yes" - even the qtext filter which has implemented it. I suggest that we add it for the qtext animation and omit it for all the filters that might fallback to pango (like this one). It can be an undocumented feature that if your application is guaranteed to use qtext, it can also use animation for those parameters (like Shotcut).

ddennedy: Oh yeah, except it is in parameters of qtext filter. This is fine.

PR #1006, opened 2024-07-21
Add power data to gps_parser and filter_gpstext

augustoborn:
Added power meter data to the GPS filter, tested and working on a GPX file extracted from Strava ([gpxexample.zip](https://github.com/user-attachments/files/16323523/gpxexample.zip)
).

### DIFF
--- a/src/modules/qt/filter_gpstext.yml
+++ b/src/modules/qt/filter_gpstext.yml
@@ -40,6 +40,7 @@ parameters:
         * #gps_dist_flat#     - total distance travelled on flat area so far
         * #gps_cadence#       - (bike) cadence, in revolutions per minute
         * #gps_temperature#   - ambient temperature (<atemp> in .gpx)
+        * #gps_power#         - power in watts
         * #gps_grade_percentage# - gradient of GPS track (computed from GPS and elev)
         * #gps_grade_degrees# - same as above but converted to degrees
       An extra word "RAW" (uppercase!) can be added to the keyword to display the 
--- a/src/modules/qt/gps_parser.cpp
+++ b/src/modules/qt/gps_parser.cpp
@@ -816,6 +816,8 @@ gps_point_proc weighted_middle_point_proc(gps_point_proc *p1,
                                                max_gps_diff_ms);
     crt_point.atemp
         = weighted_middle_double(p1->atemp, p1->time, p2->atemp, p2->time, new_t, max_gps_diff_ms);
+    crt_point.power
+        = weighted_middle_double(p1->power, p1->time, p2->power, p2->time, new_t, max_gps_diff_ms);
     return crt_point;
 }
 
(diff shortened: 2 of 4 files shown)


ddennedy:
@dany123 should review and approve this since he wrote that code.

src/modules/qt/gps_parser.cpp
```diff
@@ -937,6 +940,22 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
                 nr_atemp++;
             }
 
+            //power
+            if (gp_r[i].power != GPS_UNINIT) {
+                if (power != GPS_UNINIT && nr_power > 0 && nr_power <= nr_one_minute * 60) {
```
dany123, line 945: > nr_one_minute * 60

Not very important but I recommend removing the "* 60" part and just leave the one minute default allowed gap. While I see that Strava fills in with 0 during pauses, I can imagine some other service not bothering to write anything and you'd have a period of up to 1 hour of not pedaling where the power would be interpolated for no reason.

src/modules/qt/gps_parser.cpp
```diff
@@ -1200,6 +1226,8 @@ void qxml_parse_tcx(QXmlStreamReader &reader, gps_point_ll **gps_list, int *coun
                     crt_point.cad = reader.readElementText().toDouble();
                 } else if (reader.name() == QString("Temperature")) {
                     crt_point.atemp = reader.readElementText().toDouble();
+                } else if (reader.name() == QString("Power")) { // guessing
```
dany123, line 1229: A Shotcut forum user added a [sample tcx ](https://forum.shotcut.org/t/gps-text-video-filter/29278/34?u=daniel47) when requesting power to be added, it seems the Garmin schema uses this (a bit more complex) extension for storing power:

            <Extensions>
              <ns3:TPX>
                <ns3:Speed>0.0</ns3:Speed>
                <ns3:Watts>0</ns3:Watts>
              </ns3:TPX>
            </Extensions>

src/modules/qt/filter_gpstext.yml
dany123, line 1: I think the version (on line 5) should be incremented.

src/modules/qt/gps_parser.cpp
dany123: FYI, if the namespace thing is too annoying to deal with or you don't care about the .gpx section I am fine with approoving this as is.

src/modules/qt/filter_gpstext.yml
ddennedy: Will there be a UI change in Shotcut or Kdenlive that needs to depend on the version?

src/modules/qt/gps_parser.cpp
ddennedy: Garmin is very popular, but we can merge this and then try to extend to tcx. I am getting the impression that we are not going to hear back from the contributor.

src/modules/qt/filter_gpstext.yml
frdbr: FYI: Kdenlive does not have this feature implemented.

dany123: > Will there be a UI change in Shotcut 

For completeness  there should be a new value in the drop-down list of available fields. 
I can make a quick PR for it tomorrow.

PR #1007, opened 2024-07-21
Fix a race condition when starting a video thread

dimula73:
Variable `self->playing` is initialized by a background SDL thread that we start in `consumer_play_audio`. When FPS value is too hight (e.g. 100fps) or when the system is too slow, `self->playing` may be initialized **after** the actual check is done. It will end-up in a deadlock in `consumer_thread` because video_thread will never be started and noone will ever consume the frames from the queue.

The patch just starts the video thread unconditionally, making sure the frame will always be consumed.

See: https://bugs.kde.org/show_bug.cgi?id=489146

### DIFF
--- a/src/modules/sdl2/consumer_sdl2.c
+++ b/src/modules/sdl2/consumer_sdl2.c
@@ -63,7 +63,6 @@ struct consumer_sdl_s
     int width;
     int height;
     int out_channels;
-    atomic_int playing;
     SDL_Window *sdl_window;
     SDL_Renderer *sdl_renderer;
     SDL_Texture *sdl_texture;
@@ -342,9 +341,6 @@ static void sdl_fill_audio(void *udata, uint8_t *stream, int len)
         self->audio_avail = 0;
     }
 
-    // We're definitely playing now
-    self->playing = 1;
-
     pthread_cond_broadcast(&self->audio_cond);
     pthread_mutex_unlock(&self->audio_mutex);
 }
@@ -371,7 +367,6 @@ static int consumer_play_audio(consumer_sdl self, mlt_frame frame, int init_audi
     pcm += mlt_properties_get_int(properties, "audio_offset");
 
     if (mlt_properties_get_int(properties, "audio_off")) {
-        self->playing = 1;
         init_audio = 1;
         return init_audio;
     }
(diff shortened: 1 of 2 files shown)


dimula73:
It might actually be that `consumer_sdl2.c` needs the same change. I'm not very sure, since I cannot test this usecase. If you think I should extrapolate this change to `consumer_sdl2.c`, please tell.

dimula73:
> It might actually be that `consumer_sdl2.c` needs the same change

I've actually copied the fix into `consumer_sdl2.c`

src/modules/sdl2/consumer_sdl2.c
```diff
@@ -804,8 +808,10 @@ static void *consumer_thread(void *arg)
             // Play audio
             init_audio = consumer_play_audio(self, frame, init_audio, &duration);
 
-            // Determine the start time now
-            if (self->playing && init_video) {
+            // Start the video thread unconditionally. We cannot check for self->playing,
+            // because the variable may be initialized with a delay by a backdround SDL
+            // thread and cause race condition
+            if (init_video) {
```
ddennedy, line 814: Since `playing` is now only set and never read, it should be removed along with the comment here that no longer makes sense. 

I think this change is OK even though I did not reproduce it in melt with a 240 fps video. Maybe more repro steps are required. Otherwise, I tested it successfully in melt and Shotcut.

src/modules/sdl2/consumer_sdl2_audio.c
```diff
@@ -593,8 +597,10 @@ static void *consumer_thread(void *arg)
             // Play audio
             init_audio = consumer_play_audio(self, frame, init_audio, &duration);
 
-            // Determine the start time now
-            if (self->playing && init_video) {
+            // Start the video thread unconditionally. We cannot check for self->playing,
+            // because the variable may be initialized with a delay by a backdround SDL
+            // thread and cause race condition
+            if (init_video) {
```
ddennedy, line 603: Since `playing` is now only set and never read, it should be removed along with the comment here that no longer makes sense.

ddennedy (changes requested):

src/modules/sdl2/consumer_sdl2.c
dimula73: Hi, @ddennedy!

To reproduce the issue one needs to start/stop hi-fps playback constantly and very quickly. Like:

1) Start 100fps playback
2) Wait 1000ms
3) Stop for 100ms
4) Start again
5) Repeat

This situation is quite easy to reproduce in Krita's GUI, when the user restarts playback by clicking mouse very quickly.

PS:
I'll implement the technical part of the comment now.

dimula73: The flag is now removed!

PR #1011, opened 2024-08-07
Add support for SMPTE-esque timecodes

palmer-dabbelt:
This reads burned in timecodes from the headers of the two timecode flavors I have floating around (Tentacle and GoPro), and provides them to users.

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -657,3 +657,9 @@ MLT_7.22.0 {
     mlt_property_is_numeric;
     mlt_property_is_rect;
 } MLT_7.18.0;
+
+MLT_7.226.0 {
+  global:
+    mlt_producer_set_timecode;
+    mlt_producer_get_timecode;
+} MLT_7.22.0;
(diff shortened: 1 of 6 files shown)


palmer-dabbelt:
[this was also on the mailing list]

I'm very new to all this video editing stuff, but I'm trying to get multiple devices syced up via timecodes and I think I've mostly got it  working.  I'm using kdenlive and have a patch for that as well, but I  think the MLT side of things is going to need some work as what I have here seems a bit clunky:

* I don't think converting to milliseconds was actually the right thing to do, as I'm just converting everything to/from frames.
* I'm not sure how cut clips are meant to work, but I don't have any code to handle timecode offsets either here or in kdenlive so I'm assuming they're going to be broken.
* I'm just sort of guessing at this metadata based on reading some files  with ffprobe, I didn't read any specs.  So I have no idea if that pattern matching is going to be robust.

That said, I figured I'd just send this for now so at least I could get it out on the list and see if anyone has a better idea...

bmatherly:
Can you describe the use case for this change? For example, the mlt_producer_get_creation_time() was added so that producers (clips) could be sorted by the application in chronological order. Is that the same use for this timecode property?

Typically, when people request timecode in MLT, they want a monotonic code attached to each frame for synchronization. But this patch will only provide a single timecode (presumably for the first frame in the file).

ddennedy:
This code does not belong in the framework and can be performed in the application.

PR #1016, opened 2024-08-22
GPS text updates: .tcx power and minor fixes

dany123:
also fixed #gps_speed RAW# not showing any values

### DIFF
--- a/src/modules/qt/gps_parser.cpp
+++ b/src/modules/qt/gps_parser.cpp
@@ -943,7 +943,7 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
 
             //power
             if (gp_r[i].power != GPS_UNINIT) {
-                if (power != GPS_UNINIT && nr_power > 0 && nr_power <= nr_one_minute * 60) {
+                if (power != GPS_UNINIT && nr_power > 0 && nr_power <= nr_one_minute) {
                     nr_power++;
                     for (j = i; j > i - nr_power; j--) {
                         gp_p[j].power = power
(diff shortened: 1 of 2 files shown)


src/modules/qt/gps_parser.cpp
```diff
@@ -943,7 +943,7 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
 
             //power
             if (gp_r[i].power != GPS_UNINIT) {
-                if (power != GPS_UNINIT && nr_power > 0 && nr_power <= nr_one_minute * 60) {
+                if (power != GPS_UNINIT && nr_power > 0 && nr_power <= nr_one_minute) {
```
ddennedy, line 946: It is time for me to make a new MLT release, and I do not want to change behavior from what is already in the beta and release candidate for Shotcut. What is the impact of this change if it comes after releases?

dany123: This value sets how much time can pass between 2 different readings of power to consider it continous (as in, not paused in between). In this case it was set at 60 minutes (now 1 minute) which is realistically way too much but also doesn't actually matter as the difference will be a number slowly increasing/decreasing from the last read value to the next value instead of showing "--". 

In practice power meters will read at a consant pace (mostly every pedal rotation or some fixed few seconds) and if there's a one hour gap it probably ran out of battery or something so I'd rather it not show any value than maybe showing 123W when you're doing a massive climb of say 1000W.

For proper gpx or tcx files this change has no effect, I tested with the original PR's gpx and with the .tcx from the forum thread.

dany123: I should mention this code is in the interpolation area, that's why I mention slowly increasing/decreasing over long times.

The only candidate for allowing 1h long gaps so far has been the temperature (which is directly above this code and the reason OP used this value) which makes sense that if you read 23 degrees now and 15 degress in 59 minutes it can realistically intrerpret as the temp slowly decreased over the entire hour (but still, 1h was a long stretch, most temperature sensors report at 1 minute intervals).

PR #1017, opened 2024-08-24
Reduce log level of this VST message

bmatherly:
This three line message was spitting out every time I ran a melt command. I propose that this does not need to be a run-time warning. Maybe a compile warning?

### DIFF
--- a/src/modules/jackrack/plugin_mgr.c
+++ b/src/modules/jackrack/plugin_mgr.c
@@ -1018,8 +1018,8 @@ vst2_mgr_t *vst2_mgr_new()
     vst2_mgr_get_path_plugins(pm);
 
     if (!pm->all_plugins)
-        mlt_log_warning(
-            NULL, "No VST2 plugins were found!\n\nCheck your VST_PATH environment variable.\n");
+        mlt_log_info(
+            NULL, "No VST2 plugins were found! Check your VST_PATH environment variable.\n");
     else
         pm->all_plugins = g_slist_sort(pm->all_plugins, vst2_mgr_sort);
 


ddennedy:
I agree it is annoying, and there are similar messages all at warning level, which is the default in `melt`. But Shotcut uses info level by default except debug builds use verbose. So, I would prefer to have it a verbose. Can you change that, as well as for LV2 and LADSPA? It is not possible to log this at compile time because the plugins are discovered at run time and depend on environment variables.

ddennedy:
I was hoping to avoid showing the VST2 plugins message to Shotcut users to avoid getting requests or bug reports about getting VST2 and VST3 plugins to work. Our lack of a UI generator ("generic" on the roadmap, and making one leads to reports about being unable to translate MLT service metadata YAML) rather prevents that, and VST3 is whole 'nother beast MLT doesn't support. However, Shotcut runs melt with `-verbose` regardless of build type. Adding a `-loglevel info` to `melt.c` and changing Shotcut to add those args in `MeltJob` prevents showing some useful information in export logs such as hardware encoder. For example, VA-API shows absolutely nothing at info level but all of this at verbose!
```
[AVHWDeviceContext @ 0x78412c0059c0] Trying to use DRM render node for device 0.
[AVHWDeviceContext @ 0x78412c0059c0] libva: VA-API version 1.14.0
[AVHWDeviceContext @ 0x78412c0059c0] libva: Trying to open /usr/lib/x86_64-linux-gnu/dri/radeonsi_drv_video.so
[AVHWDeviceContext @ 0x78412c0059c0] libva: Found init function __vaDriverInit_1_14
[AVHWDeviceContext @ 0x78412c0059c0] libva: va_openDriver() returns 0
[AVHWDeviceContext @ 0x78412c0059c0] Initialised VAAPI connection: version 1.14
[AVHWDeviceContext @ 0x78412c0059c0] VAAPI driver: Mesa Gallium driver 24.0.3-1pop1~1711635559~22.04~7a9f319 for AMD Radeon Graphics (radeonsi, renoir, LLVM 15.0.7, DRM 3.57, 6.9.3-76060903-generic).
[AVHWDeviceContext @ 0x78412c0059c0] Driver not found in known nonstandard list, using standard behaviour.
[mlt_buffer @ 0x78412c1fd280] w:1280 h:720 pixfmt:nv12 tb:1001/30000 fr:30000/1001 sar:0/1 csp:unknown range:unknown
[h264_vaapi @ 0x78412c001c00] Input surface format is nv12.
[h264_vaapi @ 0x78412c001c00] Using VAAPI profile VAProfileH264Main (6).
[h264_vaapi @ 0x78412c001c00] Using VAAPI entrypoint VAEntrypointEncSlice (6).
[h264_vaapi @ 0x78412c001c00] Using VAAPI render target format YUV420 (0x1).
[h264_vaapi @ 0x78412c001c00] RC mode: CQP.
[h264_vaapi @ 0x78412c001c00] Block Level bitrate control: OFF.
[h264_vaapi @ 0x78412c001c00] RC quality: 23.
[h264_vaapi @ 0x78412c001c00] RC framerate: 30000/1001 (29.97 fps).
[h264_vaapi @ 0x78412c001c00] Driver does not report any additional prediction constraints.
[h264_vaapi @ 0x78412c001c00] Using intra and P-frames (supported references: 1 / 0).
[h264_vaapi @ 0x78412c001c00] Driver does not support some wanted packed headers (wanted 0xd, found 0x1).
[h264_vaapi @ 0x78412c001c00] Using level 3.1.
```
As well as the following only at verbose level when using faststart
```
[mp4 @ 0x78412c000f40] Starting second pass: moving the moov atom to the beginning of the file
```

Still, I can commit my melt `-loglevel` implementation that supports every log level.

bmatherly:
Let me summarize to see if I am understanding your proposal:

- Shotcut default log level - info
- Shotcut debug log level - debug
- melt default log level - warning
- melt Shotcut job log level - verbose or info

Regardless of verbose or info for the job logs, maybe we need to audit the messages and cherry-pick some messages to "promote" if they are helpful for debugging or "demote" if they are distracting. I have to admit that I am sometimes thoughtless about setting log levels in new code and just use "warning" if I want to see it and "debug" if I do not want to see it in the application log.

For example, this message seems to cause a lot of confusions among our users:
`libDeckLinkAPI.so: cannot open shared object file: No such file or directory`

This is another example that causes SPAM and has raised questions in the past:
`[swscaler @ 000001fa94146c40] YUV color matrix differs for YUV->YUV, using intermediate RGB to convert`

Having an option to set melt to any log level would be good. We already have: debug, quiet, verbose and timings (with warning as the default).

Regarding the compile option...
I think that the warning/info message should only be printed if the user was expecting to find plugins. We are obviously not expecting to find any plugins because we did not package any VST plugins. How would you feel about a compile option to disable the VST code? Something like "-DVST_PLUGINS=OFF"? The default could be ON.

ddennedy:
> Shotcut default log level - info
> Shotcut debug log level - debug

No, here is today that I am keeping:
- debug build is verbose log level
- non-debug build is info level

> melt Shotcut job log level - verbose or info

Today is verbose regardless of build type, but anything lower seems to leave out good info based on my experience.

> we need to audit the messages

I am not interested in that.

> libDeckLinkAPI.so: cannot open shared object file: No such file or directory

That comes from the BlackMagic Design code in `DeskLinkAPIDispatch.cpp` that I would rather not modify. It is somewhat useful, and even tho it gets reported it is also easy for many to dismiss.

> [swscaler @ 000001fa94146c40] 

These and many others like it come from FFmpeg libs, obviously. And its log level is set from the MLT log level. Some of these are at verbose level and others less like info or warning. That cannot be helped much.

> "-DVST_PLUGINS=OFF"

`-DUSE_VST2=OFF` already exists, but I chose not to exclude it (as well as other things) in Shotcut's build because the included `melt` is useful outside of Shotcut as well. Well, the new LV2 is not included due to missing dependency, and I am OK to leave VST2 out for now.

I will accept this as-is and extend it to the others. And for now leave Shotcut's MLT log levels alone.

PR #1018, opened 2024-08-25
Fix DTS audio-only does not play

bmatherly:
This specific file was a WMA file with DTS coded audio. The first packet of audio does not have a PTS of zero. Since there is no video in the file, MLT tries to seek to position 0 for the first audio frame and attempts to sync audio frames as if they started at 0. This patch finds a first_pts value if there is no video in the file to look for.

A user sent me this test file privately. If a release is eminent, I suggest waiting until after the release to merge this change.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -1271,6 +1271,25 @@ static void find_first_pts(producer_avformat self, int video_index)
     int64_t prev_pkt_duration = AV_NOPTS_VALUE;
 
     av_init_packet(&pkt);
+
+    if (video_index == -1 && self->audio_index >= 0) {
+        // Special case for audio only files.
+        while (ret >= 0 && pkt_countdown-- > 0 && (self->first_pts == AV_NOPTS_VALUE)) {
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -1271,6 +1271,25 @@ static void find_first_pts(producer_avformat self, int video_index)
     int64_t prev_pkt_duration = AV_NOPTS_VALUE;
 
     av_init_packet(&pkt);
+
+    if (video_index == -1 && self->audio_index >= 0) {
+        // Special case for audio only files.
+        while (ret >= 0 && pkt_countdown-- > 0 && (self->first_pts == AV_NOPTS_VALUE)) {
+            ret = av_read_frame(context, &pkt);
+            if (ret >= 0 && pkt.stream_index == self->audio_index) {
+                mlt_log_verbose(MLT_PRODUCER_SERVICE(self->parent),
```
ddennedy, line 1280: Please change this to debug level, like the video code below.

ddennedy (changes requested):

PR #1024, opened 2024-09-10
fix color shift in stills/JPEG export preset

austincbrooks:
The current JPEG export preset usually produces a color shift when exporting. This is due to typical BT.601/709 mismatch-type stuff. I have added JPEG-standard values to the export preset to ensure that all conversions/signaling/tagging is explicit.

Investigated due to the bug report at https://forum.shotcut.org/t/exported-image-color-is-different-from-the-video/45736

### DIFF
--- a/presets/consumer/avformat/stills/JPEG
+++ b/presets/consumer/avformat/stills/JPEG
@@ -1,3 +1,4 @@
+color_range=pc
 progressive=1
 f=image2
 vcodec=mjpeg
@@ -6,5 +7,7 @@ an=1
 audio_off=1
 g=1
 bf=0
+pix_fmt=yuvj422p
+colorspace=srgb
 
 meta.preset.extension=jpg


presets/consumer/avformat/stills/JPEG
```diff
@@ -6,5 +7,7 @@ an=1
 audio_off=1
 g=1
 bf=0
+pix_fmt=yuvj422p
```
bmatherly, line 10: I did not test this. But I wonder why this is not RGB8 - which would be the native pixel format for JPEG. Does this introduce an extra conversion to eventually get to RGB?

austincbrooks: `ffmpeg -h encoder=mjpeg` only supports YUV pixel formats.  I figured I would incur a conversion penalty if I handed RGB8 to the encoder?

ddennedy: Why do you think JPEG is a RGB pixel format?

bmatherly: I made an incorrect assumption. I learned something new today!

austincbrooks: My understanding was that old JPEGs were BT.601 and new JPEGs were sYCC from the sRGB spec.  Some "[light reading](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-T.871-201105-I!!PDF-E)" seemed to say that's generally true, except when it isn't, as is always the case when JPEG is involved. :)

ddennedy: That’s about color space. My reading of things (not in ffmpeg) say encoding first converts RGB to Y’CbCr. See section 6.2 in the doc Austin linked.

austincbrooks: I think we're all saying the same thing at this point.  When I said BT.601 and sYCC, I was implying that the encoding was YCbCr as well, since those are YCbCr color spaces.

PR #1037, opened 2024-10-10
Check if window size changed and resize it

### LINKED ISSUE #1036: Using consumer `sdl2` with `window_id` wont resize rendered area
When using the `sdl2` consumer with a defined `window_id` (an existing "window"), the rendered area will not be updated when the widget/window is resized.

jonata:
Fixes #1036.

Can I add PySide6/PyQt/PyGtk script sample together with `play.py`?

### DIFF
--- a/src/modules/sdl2/consumer_sdl2.c
+++ b/src/modules/sdl2/consumer_sdl2.c
@@ -573,6 +573,7 @@ static int consumer_play_video(consumer_sdl self, mlt_frame frame)
     int video_off = mlt_properties_get_int(properties, "video_off");
     int preview_off = mlt_properties_get_int(properties, "preview_off");
     int display_off = video_off | preview_off;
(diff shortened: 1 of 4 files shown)


ddennedy:
Yes please do add your player sample to help me verify this before merge.

jonata:
It would not work on Wayland. Are you using X11?

In any case, the problem on my side is what I put on the bug report, where the sdl render rect would not update when the window is resized. `SDL_SetWindowSize` with `window_width` and `window_height` is a workaround for this.

ddennedy:
I am on X11.

> the sdl render rect would not update when the window is resized

Well, that is confirmed working. I just want to head off any future bug report that these pyqt6 examples do not show video. For me, they only do for a fraction of a second because the size decreases very quickly. I can add my workaround to the examples.

jonata:
Sure, yes, I agree. Thanks!

PR #1039, opened 2024-10-17
Add decimals keyword to gps text + move RAW keyword check

dany123:
-added decimals keyword to all numeric keywords: #gps_speed decimals 5# km/h
-refactored the code around the RAW keyword to avoid all those duplicated snprintf

Requested here:
https://forum.shotcut.org/t/about-gpstext-altitude/45566
https://forum.shotcut.org/t/hello-again-gps-dist/43313


ddennedy:
Do you want to include this in Shotcut 24.10 release due by end-of-month even though it will not be tested?

dany123:
Yes, I can do more tests on the nightly to be safe, locally it was all good going through most of the keywords with a random file.

PR #1040, opened 2024-10-17
Remove GPS text guess_offset (to be done properly in qml)

dany123:
The offset update condition was done badly - I initially assumed this would only be done one time at file load but it's being done every time the offset is 0 and gps file is read (including at export time).
I have a better fix directly in filter's ui.qml (still need to test it properly, probably done tomorrow or saturday).


ddennedy:
Am I correct to conclude this is OK to apply even without the UI change since offset is not working as intended?

dany123:
Yes, it's basically a small revert so the default offset will be 0.

PR #1047, opened 2024-11-04
Add support for XML embedded in chains

bmatherly:


src/mlt++/MltChain.cpp
```diff
@@ -34,6 +36,12 @@ Chain::Chain(Profile &profile, const char *id, const char *service)
     }
 
     mlt_producer source = mlt_factory_producer(profile.get_profile(), id, service);
+    if (source
+        && !strcmp("xml", mlt_properties_get(MLT_PRODUCER_PROPERTIES(source), "mlt_service"))) {
+        mlt_producer_close(source);
+        // Always use a producer-consumer for xml so that the profile can be changed
+        source = mlt_factory_producer(profile.get_profile(), "consumer", service);
```
ddennedy, line 43: I noticed when recently debugging a related issue in Shotcut that the XML as Clip already gets a consumer producer if the profiles mismatch (a feature of the loader producer). Is that different when using a chain? It is not very desirable to add knowledge about certain modules within the framework proper. It is not a hard rule but rather a goal to minimize it.

bmatherly: I noticed this too. When encapsulating XML producer in a chain, the problem becomes that sometimes that producer would be wrapped in consumer producer and sometimes it would not. The chain needs to be able to change the profile on the encapsulated producer. The XML producer does not support this very well because it does not propagate the profile change down to the producers inside of it. Forcing it to always use the producer_consumer works around that problem.

>  It is not very desirable to add knowledge about certain modules within the framework proper.

I agree - and I would prefer a generic way to do this. Another option would be to add a private property to producers called "_mutable_profile". If the profile can be changed (like avformat), then the property would be true. If the profile can not be changed (like xml), the property would be false. That could be the condition to decide if a producer_consumer should be used.

src/modules/xml/producer_xml-clip.yml
```diff
@@ -0,0 +1,18 @@
+schema_version: 0.1
```
ddennedy, line 1: I prefer to use the latest version 7.0 for new files,  partly because I feel that since this is a new producer (not a specially named producer for the existing `xml` producer) and should get the 
```
parameters:
  - identifier: resource
    argument: yes
    title: File
    type: string
    description: An XML text file containing MLT XML.
    readonly: no
    required: yes
    mutable: no
    widget: fileopen
```
And `argument` requires a newer schema version than the first 0.1. Instead of analyzing the minimum version just use the latest to not worry.

src/modules/xml/producer_xml-clip.c
```diff
@@ -0,0 +1,226 @@
+/*
+ * producer_xml-native.c -- A wrapper for mlt XML in native profile
```
ddennedy, line 2: This has the wrong filename in the comment header.

src/modules/xml/producer_xml-clip.c
```diff
@@ -0,0 +1,226 @@
...
+                       mlt_properties_get(frame_properties, "consumer.rescale"));
+    mlt_properties_set_int(xml_frame_properties,
+                           "consumer.top_field_first",
+                           mlt_properties_get_int(frame_properties, "consumer.top_field_first"));
+    mlt_properties_set(xml_frame_properties,
+                       "consumer.color_trc",
+                       mlt_properties_get(frame_properties, "consumer.color_trc"));
+    mlt_properties_set(xml_frame_properties,
+                       "consumer.color_range",
+                       mlt_properties_get(frame_properties, "consumer.color_range"));
```
ddennedy, line 96: What about using `mlt_properties_copy(xml_frame_properties, frame_properties, "consumer.")`?

src/modules/xml/producer_xml-clip.yml
bmatherly: Done

src/modules/xml/producer_xml-clip.c
bmatherly: Done

src/modules/xml/producer_xml-clip.c
bmatherly: done

bmatherly:
All great comments. Thanks.

PR #1063, opened 2025-02-02
Color Producer: add "checkerboard" as special keyword

jlskuz:
Do you think this makes sense or should we introduce a strictly separated checkerboard producer for this?

### DIFF
--- a/src/modules/core/producer_colour.c
+++ b/src/modules/core/producer_colour.c
@@ -17,6 +17,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
+#include "framework/mlt_image.h"
 #include <framework/mlt_frame.h>
 #include <framework/mlt_log.h>
 #include <framework/mlt_pool.h>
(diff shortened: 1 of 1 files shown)


jlskuz:
You can test this change with `melt color:checkerboard`

grafik

bmatherly:
I don't feel strongly about this. But I guess I would lean toward making it a separate producer. Maybe other people feel differently.

ddennedy:
Why not use checkerboard.svg?

jlskuz:
It would be difficult to make a static image fit to all the different profile resolutions . Or is there an easy approach I do miss?

ddennedy:
Good question. I think it will work best as a filling pattern. I asked Gemini AI about it, and it believes that this pattern element is supported in Qt 6 even though it is not a part of SVG Tiny 1.2. Gemini wrote:
Yes, you can create custom pattern fills in SVG.  Here's how, along with explanations and examples:

```
<svg width="300" height="200">
  <defs>  <pattern id="my-custom-pattern" x="0" y="0" width="20" height="20" patternUnits="userSpaceOnUse">
      <rect x="0" y="0" width="10" height="10" fill="red" />
      <rect x="10" y="10" width="10" height="10" fill="blue" />
      </pattern>
  </defs>

  <rect x="20" y="20" width="260" height="160" fill="url(#my-custom-pattern)" />
</svg>

```
Explanation:
 * <defs> tag:  The <defs> (definitions) tag is where you define reusable SVG elements, including patterns.  This keeps your main drawing area cleaner.
 * <pattern> tag: This tag defines the pattern itself.  Key attributes:
   * id:  A unique identifier for the pattern so you can reference it later.  (e.g., id="my-custom-pattern")
   * x and y: The starting coordinates of the pattern tile.  Think of this as the top-left corner of the first tile.
   * width and height: The width and height of the pattern tile.  This determines how often the pattern repeats.
   * patternUnits:  Crucial!
     * userSpaceOnUse:  The pattern scales with the element it's applied to.  If the element is scaled, the pattern tiles also scale.  This is the most common and usually the desired behavior.
     * objectBoundingBox: The pattern scales to fit the bounding box of the element.  Less common and often leads to unexpected results if the element is scaled non-uniformly.
 * Pattern Content: Inside the <pattern> tag, you put the SVG elements that make up the repeating tile.  In the example above, we have two rectangles, one red and one blue.  You can use any SVG elements here:
   * <rect>, <circle>, <line>, <path>, <image>, <text>, etc.
   * You can even use other patterns within a pattern!
   * The content is repeated horizontally and vertically to fill the shape to which the pattern is applied.
 * Applying the Pattern: To use the pattern as a fill, you reference it using url(#pattern-id) in the fill attribute of a shape:
   `<rect x="20" y="20" width="260" height="160" fill="url(#my-custom-pattern)" />`

Key Improvements and Considerations:
 * userSpaceOnUse:  Using patternUnits="userSpaceOnUse" is generally recommended.  It makes the pattern behave more intuitively when the filled element is transformed (scaled, translated, etc.).
 * Pattern Tile Size: The width and height of the <pattern> tag control the size of the repeating tile.  Experiment with these values to get the desired effect.
 * Complex Patterns: You can create very complex patterns by nesting SVG elements and using different shapes, colors, and transformations within the pattern definition.
 * Example with a Circle and Text:

```
<svg width="200" height="150">
  <defs>
    <pattern id="circle-text-pattern" x="0" y="0" width="50" height="50" patternUnits="userSpaceOnUse">
      <circle cx="25" cy="25" r="20" fill="green" />
      <text x="25" y="28" text-anchor="middle" dominant-baseline="middle" font-size="12" fill="white">SVG</text>
    </pattern>
  </defs>

  <rect x="10" y="10" width="180" height="130" fill="url(#circle-text-pattern)" />
</svg>

```

This example creates a pattern of green circles with the text "SVG" inside.
Remember to adjust the x, y, width, height, and the contents of the pattern to create the exact fill you need.  This gives you a powerful way to add custom textures and designs to your SVGs.

ddennedy:
The SVG route makes it more customizable. If made a distinct producer, it can creep with options, which can be good or bad (new functions to not break API). So if SVG does not work out, then maybe this keyword approach constrains it.

ddennedy:
There is also frei0r filter that can show the alpha channel as a checkerboard pattern that you can use with a color:0 producer.

jlskuz:
I think you mean the frei0r.alpha0ps filter. That checkerboard has a way too small scale:



I will check what you wrote about the SVG and if it can fit my use case.

ddennedy:
Gemini gave me this to start with. I am on phone while traveling so unable to help in detail for a couple of days.

```
<svg width="100" height="100" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <pattern id="checkerboard" x="0" y="0" width="10" height="10" patternUnits="userSpaceOnUse">
      <rect x="0" y="0" width="5" height="5" fill="white" />
      <rect x="5" y="0" width="5" height="5" fill="lightgray" />
      <rect x="0" y="5" width="5" height="5" fill="lightgray" />
      <rect x="5" y="5" width="5" height="5" fill="white" />
    </pattern>
  </defs>

  <rect x="0" y="0" width="100" height="100" fill="url(#checkerboard)" />

  <rect x="20" y="20" width="60" height="60" fill="rgba(255, 0, 0, 0.5)" />
</svg>

```

jlskuz:
With the SVG you posted above pasted to a file `pattern.svg` I get:

```
melt ./pattern.svg
qt.svg: ./pattern.svg:13:6: Could not resolve property: #checkerboard
qt.svg: ./pattern.svg:13:6: Could not resolve property: #checkerboard
```

and just black.

Interestingly importing the same file to Shotcut or Kdenlive shows up fine.

And still the SVG approach would mean we need to set the `width` and `height` of the SVG from the Kdenlive code based on the current profile / monitor resolution. Or how did you imagine it?

frdbr:
Just an example for inspiration on how this works in Blender

Blender alpha monitor toggle

bmatherly:
I am opposed to this as a keyword for the color producer. I would prefer that it be a new producer so that we can add other parameters over time - like the size of the squares. I am willing to move the code to a new producer if nobody is opposed.

ddennedy:
I don’t want to be able to add parameters. MLT framework is not a canvas API!!! Honestly, there are still other options to explore including expanding the Frei0r filter, Glaxnimate, libavfilter. But I’m a little tired of leading this thread, and this  undocumented exposure of an existing API seems pragmatic.

bmatherly:
> Honestly, there are still other options to explore including expanding the Frei0r filter, Glaxnimate, libavfilter.

I agree with that and I thought the conversation was trending in that direction. So I did not chime in.

> this undocumented exposure of an existing API seems pragmatic.

I guess so. And since we did not update the YAML to include this keyword, it is also undocumented from the filter parameters. It can be a secret for now. And we can deprecate it later if we want to recommend a different way.

Thanks for humoring me.

jlskuz:
Thanks for merging!

Just to add some context: we want to use it as an (optional) monitor background for clips with alpha. Since that uses a color clip already to show different colored background at the users choice the special keyword seemed the best option for me.

> Honestly, there are still other options to explore including expanding the Frei0r filter, Glaxnimate, libavfilter.

I agree and if the frei0r.alpha0ps improves that can definitely be a good replacement. However the advance of the keyword solution is, that it also works if frei0r is not available/installed (but one can argue that it will be in most cases and the checkerboard monitor background feature is not critical).

> I guess so. And since we did not update the YAML to include this keyword, it is also undocumented from the filter parameters. It can be a secret for now. And we can deprecate it later if we want to recommend a different way.

As we do not plan to expose it to the user and only use it for the monitor background as described above that would work without major issues from the MLT user perspective for us.

PR #1064, opened 2025-02-03
Fix various aspect ratio issues with qtblend filter/transition

j-b-m:
This fixes multiple issues when applying several qtblend filters on the same frame.
One issue is that the qtblend filter, like the affine filter/transition, sometimes query an image using the producer's original resolution, not in profile resolution in order to be able to display for example to zoom in a 4k clip with an HD profile, while keeping the 4K resolution.

However this caused issues because in several parts of the code, we calculate the preview scaling by comparing the size of the requested image with the size of the project profile.

Among the several fixes in this patch, we now check if another qtblend filter was applies and use a cached value for consumer scaling.

### DIFF
--- a/src/modules/qt/common.h
+++ b/src/modules/qt/common.h
@@ -21,7 +21,6 @@
 
 #include <framework/mlt.h>
 
-#define MLT_QTBLEND_MAX_DIMENSION (16000)
 
 class QImage;
 
--- a/src/modules/qt/common.h
+++ b/src/modules/qt/common.h
@@ -21,6 +21,8 @@
 
 #include <framework/mlt.h>
 
+#define MLT_QTBLEND_MAX_DIMENSION (16000)
+
 class QImage;
 
 bool createQApplicationIfNeeded(mlt_service service);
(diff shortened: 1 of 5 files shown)


src/modules/qt/common.h
```diff
@@ -21,6 +21,8 @@
 
 #include <framework/mlt.h>
 
+#define MLT_QTBLEND_MAX_DIMENSION (16000)
```
bmatherly, line 24: Why is this defined here if it is only used in filter_qtblend.cpp

src/modules/qt/filter_qtblend.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * filter_lightshow.cpp -- animate color to the audio
+ * filter_qtblend.cpp -- Qt composite filter
  * Copyright (C) 2015 Meltytech, LLC
```
bmatherly, line 3: It would be good to update the copyright date in all of these files that are changed.

bmatherly (changes requested):
I don't have much opinion on the substance of the change. It seems a little strange that a filter would keep track of whether it has been applied already. But I don't have a better idea for how to handle that situation. Maybe there should be a generic way to indicate that the original image has already been scaled (like affine).

src/modules/qt/filter_qtblend.cpp
j-b-m: Done, thanks for looking at it. Since these changes only affect qtblend that is mostly used in Kdenlive, I will merge the changes in a few days if there is no opposition.

bmatherly (approved):

PR #1069, opened 2025-02-24
Check for QT's offscreen platform

j-b-m:
Currently the Qt module (and Glaxnimate) abort when the `DISPLAY` or `WAYLAND_DISPLAY` variables are not set, meaning there is a graphical server running. However, use of Qt modules is also possible using Qt's offscreen rendering, which is enabled by setting the QT_QPA_PLATFORM to `offscreen`.

This patch checks for this condition and allows running the Qt modules in this case.

### DIFF
--- a/src/modules/glaxnimate/producer_glaxnimate.cpp
+++ b/src/modules/glaxnimate/producer_glaxnimate.cpp
@@ -141,12 +141,15 @@ static bool createQApplicationIfNeeded(mlt_service service)
 #endif
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
         if (getenv("DISPLAY") == 0 && getenv("WAYLAND_DISPLAY") == 0) {
-            mlt_log_error(service,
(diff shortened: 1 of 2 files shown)


src/modules/glaxnimate/producer_glaxnimate.cpp
```diff
@@ -141,12 +141,15 @@ static bool createQApplicationIfNeeded(mlt_service service)
 #endif
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
         if (getenv("DISPLAY") == 0 && getenv("WAYLAND_DISPLAY") == 0) {
-            mlt_log_error(service,
+            const char *qt_qpa = getenv("QT_QPA_PLATFORM");
+            if (qt_qpa == 0 || strcmp(qt_qpa, "offscreen") != 0) {
```
ddennedy, line 145: We prefer (!qt_qpa || strcmp(qt_qpa, “offscreen”))  rather than add additional, unnecessary comparisons.

PR #1070, opened 2025-02-27
Fix pixel format detection for VP8/VP9 streams

j-b-m:
FFmpeg's native VP8 / VP9 decoders don't support alpha channel, while the libvpx versions do. A workaround to correctly play the video is implemented in MLT to switch to the vpx versions of the decoders if available:
https://github.com/mltframework/mlt/blob/master/src/modules/avformat/producer_avformat.c#L2715

However the pixel format info is still incorrect: clips with alpha are reported as yuv420p instead of yuva420p, because the native decoder is used when calling `avformat_find_stream_info` , so we get info from the native decoder, see for reference this tread : 
https://stackoverflow.com/questions/66702932/is-there-a-way-to-force-ffmpeg-to-decode-a-video-stream-with-alpha-from-a-webm

This is problematic as for example in Kdenlive we rely on the pixel format to detect alpha.

Below is a sample clip that can be used to reproduce the problem:
https://invent.kde.org/-/project/19/uploads/3104b97f60c00a07dddb9d0343465fc2/Directed_Exit_short.webm

`melt Directed_Exit_short.webm -consumer xml:`
Reports a meta.media.0.codec.pix_fmt as **yuv420p** despite the clip having an alpha channel

With my change, the same command correctly reports **yuva420p**.

Note that I am not so familiar with avformat code so any suggestion to improve or a different way to fix is welcome.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -490,7 +490,7 @@ static mlt_properties find_default_streams(producer_avformat self)
     // Allow for multiple audio and video streams in the file and select first of each (if available)
     for (i = 0; i < context->nb_streams; i++) {
         // Get the codec context
-        AVFormatContext *vpxContext = NULL;
+        AVFormatContext *vpx_context = NULL;
         AVStream *stream = context->streams[i];
         if (!stream)
             continue;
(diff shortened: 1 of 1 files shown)


src/modules/avformat/producer_avformat.c
```diff
@@ -490,13 +490,44 @@ static mlt_properties find_default_streams(producer_avformat self)
...
         // Get the codec context
+        AVFormatContext *vpxContext = NULL;
         AVStream *stream = context->streams[i];
         if (!stream)
             continue;
         AVCodecParameters *codec_params = stream->codecpar;
         const AVCodec *codec = avcodec_find_decoder(codec_params->codec_id);
         if (!codec)
             continue;
+        int switchToVpxCodec = 0;
```
ddennedy, line 501: Symbols do not use camelCase in MLT! This also applies to `vpxContext`.  Otherwise, this looks OK.

ddennedy (changes requested):

PR #1074, opened 2025-03-16
Add mlt_service_set_consumer

bmatherly:
This can be used to restore a consumer in case it is stolen temporarily (for example to run the XML consumer).

### DIFF
--- a/src/framework/mlt.vers
+++ b/src/framework/mlt.vers
@@ -662,3 +662,8 @@ MLT_7.30.0 {
   global:
     mlt_image_full_range;
 } MLT_7.22.0;
+
+MLT_7.32.0 {
+  global:
+    mlt_service_set_consumer;
+} MLT_7.30.0;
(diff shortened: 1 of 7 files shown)


ddennedy:
I try to require a unit test for things touching the framework. Please

bmatherly:
> I try to require a unit test for things touching the framework. Please

Thanks for the reminder. I agree. I also added the Mlt++ API. It is good to be complete.

I understand this might need a release. Maybe we want to wait to merge this until we have more changes that justify a release.

ddennedy:
Yes, it needs a release, but the bigger point is that Shotcut needs to bump to the new release in `CMakeLists.txt`, which is based on the version of MLT needed to compile, not necessarily the recommended. These things greatly impact distros. There are some important fixes to get out for Shotcut, and I would not want them to delay that because they need to update MLT, which might requires some validation with dependents Kdenlive and Flowblade. You see what I mean? Nonetheless, the fix for Shotcut is important too! So, I am on the fence. What do you think?

bmatherly:
There are only a couple of minor changes since the last release. My opinion is that this fix can wait for another release. But I do not know if we will have another API breaking change soon. The last one was in January. But before that it was November 2023!

PR #1103, opened 2025-07-21
Use PThreads4W for MSVC

jlskuz:
Inital implementation for some modules, more to be added


ddennedy:
Does "Inital implementation" mean this is supposed to be a draft PR, not actually ready for review?

jlskuz:
It just means that not all modules are covered yet, but it can be reviewed and merged anyways

ddennedy (approved):

Issue #864, opened 2023-01-27
Alpha blending issue


frdbr:
Alpha blending doesn't work properly leading to ill-blended artifacts in transitions. See example in Kdenlive [bug report](https://bugs.kde.org/show_bug.cgi?id=422918). Alpha blending can't be done properly in 8 bits uint because gamma needs to be decoded for alpha to mean occlusion as in the Porter & Duff paper. More information on [wikipedia](https://en.wikipedia.org/wiki/Alpha_compositing#Gamma_correction) and [youtube](https://www.youtube.com/watch?v=LKnqECcg6Gw).

![image](https://user-images.githubusercontent.com/77309092/215001252-e8ed3194-b057-4775-94db-0347af5e8068.png)

A possible solution would be to:
1. Step one would be to use at least float16
2. Then undo gamma for every RGB
3. Then compose, then redo gamma on RGB

This report is compiled from information provided by @aurelienpierreeng [here](https://t.me/kdenlive/57811) and [here](https://t.me/kdenlive_dev/12877)

ddennedy:
This is a [road map](https://www.mltframework.org/changes/todo/) set of items, works correctly, in Movit, and is not limited to a few places. Please be aware that converting all processing to work in linear RGB will change the behavior of many color-oriented filters essentially breaking compatibility if not done correctly and thus existing projects should not benefit from a change here. Fixing it also has a big impact on performance and so must be coupled with something that works well with GPU. Then to keep performance good while still having a decently large toolset of effects means many things need to be ported. But there is vast landscape of GPU technologies, and there should always be a multi-threaded CPU fallback that is not GPU emulation. So, this is a whole big can of worms.

ddennedy:
Fixed by #1159 for the next version.
It requires the app to opt into by supplying a new consumer property `mlt_color_trc=linear`. This adds a considerable performance overhead, which we hope to improve some, but I would not expect a lot. An app/user could do the interactive work in like a draft mode without this and then turn it on only for export render. Next, upon review of the result, if some things need more precise adjustment turn on linear within the app to fine tune and re-render. Similar to a workflow with preview scaling.

PR #1104, opened 2025-07-24
Add mlt_image_rgba64 format

bmatherly:
This proposed patch adds a new image format: 64bit per pixel RGBA (16 bits per channel).
It also implements some filters (boxblur, qt, and ffmpeg) to use the new format.

Example commands:
melt color:red -consumer sdl2 mlt_image_format=rgba64

melt test.mp4 -filter avfilter.gblur av.sigma=50 -consumer sdl2 mlt_image_format=rgba64

melt test.mp4 -filter qtext:"hello" fgcolour=0x777777ff -consumer sdl2 mlt_image_format=rgba64

Looking for discussion and comments about whether we would want to add this image format to support higher bit depth processing. Comments welcome.

### DIFF
--- a/src/framework/mlt_image.c
+++ b/src/framework/mlt_image.c
@@ -3,7 +3,7 @@
  * \brief Image class
  * \see mlt_mlt_image_s
  *
- * Copyright (C) 2020-2024 Meltytech, LLC
+ * Copyright (C) 2020-2025 Meltytech, LLC
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
--- a/src/framework/mlt_image.c
+++ b/src/framework/mlt_image.c
@@ -195,6 +195,8 @@ int mlt_image_calculate_size(mlt_image self)
         return self->width * self->height * 3;
     case mlt_image_yuv444p10:
         return self->width * self->height * 6;
+    case mlt_image_rgba64:
+        return self->width * self->height * 4 * 2;
     case mlt_image_none:
     case mlt_image_invalid:
         return 0;
@@ -232,6 +234,8 @@ const char *mlt_image_format_name(mlt_image_format format)
         return "yuv420p10";
     case mlt_image_yuv444p10:
         return "yuv444p10";
+    case mlt_image_rgba64:
+        return "rgba64";
     case mlt_image_invalid:
         return "invalid";
     }
@@ -276,7 +280,8 @@ void mlt_image_fill_black(mlt_image self)
     case mlt_image_opengl_texture:
         return;
     case mlt_image_rgb:
-    case mlt_image_rgba: {
+    case mlt_image_rgba:
+    case mlt_image_rgba64: {
         int size = mlt_image_calculate_size(self);
         memset(self->planes[0], 0, size);
         break;
@@ -432,6 +437,20 @@ void mlt_image_fill_checkerboard(mlt_image self, double sample_aspect_ratio)
         memset(self->planes[1], 128, self->height * self->strides[1] / 2);
         memset(self->planes[2], 128, self->height * self->strides[2] / 2);
     } break;
+    case mlt_image_rgba64: {
+        uint16_t *p = (uint16_t *) self->planes[0];
+        for (int i = 0; i < self->height; i++) {
+            for (int j = 0; j < self->width; j++) {
+                uint16_t color = ((((i + oy) / h) % 2) ^ (((j + ox) / w) % 2)) ? gray1 : gray2;
+                color *= 256;
+                p[0] = color;
+                p[1] = color;
+                p[2] = color;
+                p[3] = 0xffff;
+                p += 4;
+            }
+        }
+    } break;
     }
 }
 
@@ -454,7 +473,8 @@ void mlt_image_fill_white(mlt_image self, int full_range)
     case mlt_image_opengl_texture:
         return;
     case mlt_image_rgb:
-    case mlt_image_rgba: {
+    case mlt_image_rgba:
+    case mlt_image_rgba64: {
         int size = mlt_image_calculate_size(self);
         memset(self->planes[0], 255, size);
         break;
@@ -526,6 +546,14 @@ void mlt_image_fill_opaque(mlt_image self)
                 *pLine += 4;
             }
         }
+    } else if (self->format == mlt_image_rgba64 && self->planes[0] != NULL) {
+        for (int line = 0; line < self->height; line++) {
+            uint16_t *pLine = (uint16_t *) self->planes[0] + (self->strides[0] * line) + 3;
+            for (int pixel = 0; pixel < self->width; pixel++) {
+                *pLine = 0xffff;
+                *pLine += 4;
+            }
+        }
     } else if (self->planes[3] != NULL) {
         memset(self->planes[3], 255, self->height * self->strides[3]);
     }
@@ -576,6 +604,10 @@ int mlt_image_format_size(mlt_image_format format, int width, int height, int *b
         if (bpp)
             *bpp = 6;
         return 6 * height * width;
+    case mlt_image_rgba64:
+        if (bpp)
+            *bpp = 8;
+        return 8 * height * width;
     default:
         if (bpp)
             *bpp = 0;
--- a/src/framework/mlt_types.h
+++ b/src/framework/mlt_types.h
@@ -48,6 +48,7 @@ typedef enum {
     mlt_image_yuv422p16, /**< planar YUV 4:2:2, 32bpp, (1 Cr & Cb sample per 2x1 Y samples), little-endian */
     mlt_image_yuv420p10, /**< planar YUV 4:2:0, 15bpp, (1 Cr & Cb sample per 2x2 Y samples), little-endian */
     mlt_image_yuv444p10, /**< planar YUV 4:4:4, 30bpp, (1 Cr & Cb sample per 1x1 Y samples), little-endian */
+    mlt_image_rgba64, /**< 16-bit RGB with alpha channel */
     mlt_image_invalid
 } mlt_image_format;
 
--- a/src/modules/avformat/filter_avcolour_space.c
+++ b/src/modules/avformat/filter_avcolour_space.c
@@ -71,6 +71,9 @@ static int convert_mlt_to_av_cs(mlt_image_format format)
     case mlt_image_yuv444p10:
         value = AV_PIX_FMT_YUV444P10LE;
         break;
+    case mlt_image_rgba64:
+        value = AV_PIX_FMT_RGBA64LE;
+        break;
     default:
         mlt_log_error(NULL,
                       "[filter avcolor_space] Invalid format %s\n",
@@ -187,7 +190,7 @@ static int convert_image(mlt_frame frame,
         uint8_t *output = mlt_pool_alloc(size);
 
         if (out_width == width && out_height == height) {
-            if (*format == mlt_image_rgba) {
+            if (*format == mlt_image_rgba && output_format != mlt_image_rgba64) {
                 register int len = width * height;
                 uint8_t *alpha = mlt_pool_alloc(len);
 
@@ -227,6 +230,20 @@ static int convert_image(mlt_frame frame,
                     }
                     mlt_frame_set_alpha(frame, alpha, len, mlt_pool_release);
                 }
+            } else if (*format == mlt_image_rgba64 && output_format != mlt_image_rgba) {
+                // Extract the alpha mask from the RGBA image
+                int len = width * height;
+                uint8_t *alpha = mlt_pool_alloc(len);
+                if (alpha) {
+                    const uint16_t *s = ((uint16_t *) *image) + 3; // start on the alpha component
+                    uint8_t *d = alpha;
+                    for (int i = 0; i < len; i++) {
+                        *d = lrint((float) *s / 256.0);
+                        d++;
+                        s += 4;
+                    }
+                    mlt_frame_set_alpha(frame, alpha, len, mlt_pool_release);
+                }
             }
         } else {
             // Scaling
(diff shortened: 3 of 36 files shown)


ddennedy:
Thank you for this, Brian. It does not need to be implemented in this PR, but there must be a plan to support a working/intermediate luma (color transfer) other than sRGB with gamma to fix #864. It has been easy to say, well, it is legacy 8-bit and wave a hand about performance. However, it does not make sense to offer higher bit depth to improve quality and not fix that: color management.

While it *might* be nice, it does not necessarily need to use OpenColorIO. The nice things about OpenColorIO IMO are support for many camera Log color profiles and other advanced color inputs, a 3D LUT for the GPU, and the color professionals behind it. However, it can be tricky to integrate for the GPU, and the non-GPU performance is awful. There can be a proof-of-concept to use libswscale to convert to and from linear color. If OpenColorIO is chosen, then that requires GPU integration (i.e. Movit today). If there is some implementation of Vulkan in MLT, then apparently there is a way to convert the shaders generated by OCIO to Vulkan at runtime as Blender is doing that.

I do not think it makes sense to couple the linear color to the new rgba64 type because producers such as qimage and qtext do not output that. There can be a new consumer property like `work_color_transfer` and a frame property `color_transfer` to go along with a new normalizing filter. Next, consumers need a way to convert from, say, `work_color_transfer=linear` to whatever is implied by the other existing `colorspace` and `color_trc` properties. 

In summary, a plan for high bit depth implies a plan for color management, and both of those imply some plan for GPU processing unless... libswscale (and/or [next FFmpeg version's color management system](https://www.phoronix.com/news/FFmpeg-HDR-Work-EOY2024)) proves suitable enough, people can continue to use the avfilter.lut3d for Log footage, and all of this is decent enough performance on modern hardware. I know performance will not be good on CPU by doing these conversions and all processing at this higher bit depth, but will it be acceptable for some?

ddennedy:
I thought a little about how doing some prototype of the linear conversion. I found that something like this works today to convert to linear and back:

`$ melt ~/Videos/IMG_0859.MOV -attach avfilter.colorspace av.ispace=bt709 av.space=bt709 av.iprimaries=bt709 av.primaries=bt709 av.trc=linear -attach avfilter.colorspace av.ispace=bt709 av.space=bt709 av.iprimaries=bt709 av.primaries=bt709 av.itrc=linear av.trc=bt709`

It is not proper because this must come before scaling and deinterlace, but I think it is good enough to use with this PR to evaluate performance.

bmatherly:
Thanks for the comments, Dan. I agree with all our points. I think that we could merge this code, and then move on to handle color management. I've converted this from a draft to a review. Please have a look at the code.

Maybe a baby-step in the right direction could be to add a step to the image convert function to apply the colorspace filter to a frame if necessary:
https://github.com/mltframework/mlt/blob/8b0f73e7363a8032a98d769900f8ed382e96b1a4/src/modules/avformat/filter_avcolour_space.c#L85

src/framework/mlt_types.h
```diff
@@ -48,6 +48,7 @@ typedef enum {
     mlt_image_yuv422p16, /**< planar YUV 4:2:2, 32bpp, (1 Cr & Cb sample per 2x1 Y samples), little-endian */
     mlt_image_yuv420p10, /**< planar YUV 4:2:0, 15bpp, (1 Cr & Cb sample per 2x2 Y samples), little-endian */
     mlt_image_yuv444p10, /**< planar YUV 4:4:4, 30bpp, (1 Cr & Cb sample per 1x1 Y samples), little-endian */
+    mlt_image_rgba64, /**< 16-bit RGB with alpha channel */
```
ddennedy, line 51: The title of the PR says "mlt_image_rgb64" but that is not defined here. The PR title becomes the commit message when squash-merged, which I do since it not only squashes but contains the PR number to refer to discussion around a change.

src/framework/mlt_image.c
```diff
@@ -195,6 +195,8 @@ int mlt_image_calculate_size(mlt_image self)
         return self->width * self->height * 3;
     case mlt_image_yuv444p10:
         return self->width * self->height * 6;
+    case mlt_image_rgba64:
```
ddennedy, line 198: Please update the copyright year in modified files that need it.

src/framework/mlt_image.c
```diff
@@ -432,6 +437,20 @@ void mlt_image_fill_checkerboard(mlt_image self, double sample_aspect_ratio)
...
+        uint16_t *p = (uint16_t *) self->planes[0];
+        for (int i = 0; i < self->height; i++) {
+            for (int j = 0; j < self->width; j++) {
+                uint16_t color = ((((i + oy) / h) % 2) ^ (((j + ox) / w) % 2)) ? gray1 : gray2;
+                color *= 256;
+                p[0] = color;
+                p[1] = color;
+                p[2] = color;
+                p[3] = 0xffff;
+                p += 4;
```
ddennedy, line 450: I thought we prefer array indexing over pointer math (increment) now.

src/modules/avformat/filter_avcolour_space.c
```diff
@@ -227,6 +230,20 @@ static int convert_image(mlt_frame frame,
...
+                // Extract the alpha mask from the RGBA image
+                int len = width * height;
+                uint8_t *alpha = mlt_pool_alloc(len);
+                if (alpha) {
+                    const uint16_t *s = ((uint16_t *) *image) + 3; // start on the alpha component
+                    uint8_t *d = alpha;
+                    for (int i = 0; i < len; i++) {
+                        *d = lrint((float) *s / 256.0);
+                        d++;
+                        s += 4;
```
ddennedy, line 243: Why not use Duff's Device like the code above it?

src/modules/avformat/filter_avcolour_space.c
```diff
@@ -298,8 +313,24 @@ static int convert_image(mlt_frame frame,
...
+                int alpha_size = 0;
+                uint8_t *alpha = mlt_frame_get_alpha_size(frame, &alpha_size);
+                if (alpha && alpha_size >= len) {
+                    const uint8_t *s = alpha;
+                    uint16_t *d = ((uint16_t *) *image) + 3; // start on the alpha component
+                    for (int i = 0; i < len; i++) {
+                        *d = *s << 8;
+                        d += 4;
+                        s++;
+                    }
```
ddennedy, line 327: Duff's Device?

src/modules/avformat/filter_avfilter.c
```diff
@@ -64,6 +64,14 @@ static int animatable_avoption(const AVOption *opt)
 }
 #endif
 
+int is_rgb(int format)
+{
+    if (format == mlt_image_rgb || format == mlt_image_rgba || format == mlt_image_rgba64) {
+        return 1;
+    }
+    return 0;
```
ddennedy, line 72:
```suggestion
    return format == mlt_image_rgb || format == mlt_image_rgba || format == mlt_image_rgba64;
```

src/modules/core/filter_brightness.c
```diff
@@ -72,6 +72,15 @@ static int sliced_proc(int id, int index, int jobs, void *cookie)
                     p += 4;
                 }
             }
+        } else if (ctx->image->format == mlt_image_rgba64) {
+            for (int line = 0; line < slice_height; line++) {
+                uint16_t *p = (uint16_t *) ctx->image->planes[0]
+                              + ((slice_line_start + line) * ctx->image->strides[0]) + 3;
+                for (int pixel = 0; pixel < ctx->image->width; pixel++) {
+                    *p = (*p * m) >> 16;
+                    p += 4;
```
ddennedy: Use array indexing, but `p[pixel]` reads weird  using `pixel` for the index var name. I would change `line` to `row`, `p` to `line`, and `pixel` to `col` => `line[col]`

src/modules/core/filter_pillar_echo.c
```diff
@@ -59,7 +59,7 @@ typedef struct
     mlt_rect rect;
 } scale_sliced_desc;
 
-static int scale_sliced_proc(int id, int index, int jobs, void *data)
+static int scale_sliced_proc_rgb32(int id, int index, int jobs, void *data)
```
ddennedy: This is rgb24 not 32

src/modules/core/producer_colour.c
```diff
@@ -187,6 +188,17 @@ static int producer_get_image(mlt_frame frame,
...
                 }
                 break;
+            case mlt_image_rgba64: {
+                uint16_t *p16 = (uint16_t *) p;
+                const int pixel_count = *width * *height;
+                for (int j = 0; j < pixel_count; j++) {
+                    *p16++ = color.r << 8;
+                    *p16++ = color.g << 8;
+                    *p16++ = color.b << 8;
+                    *p16++ = color.a << 8;
```
ddennedy: Let's use array indexing in new code instead of like `*p16++`: `p16[j + 1]` etc.

src/modules/qt/common.cpp
```diff
@@ -63,13 +63,24 @@ bool createQApplicationIfNeeded(mlt_service service)
...
+#endif
+    return mlt_image_rgba;
+}
+
+void convert_qimage_to_mlt(QImage *qImg, uint8_t *mImg, int width, int height)
 {
 #if QT_VERSION >= QT_VERSION_CHECK(5, 2, 0)
     // QImage::Format_RGBA8888 was added in Qt5.2
-    // Nothing to do in this case  because the image was modified directly.
+    // QImage::Format_RGBA64 was added in Qt5.12
```
ddennedy: We do not need to keep the old code for Qt 5.2. No one uses that anymore. Minimum Qt 5 version we support now is its last, which is also LTS: 5.15

src/modules/qt/common.cpp
```diff
@@ -87,8 +98,19 @@ void convert_qimage_to_mlt_rgba(QImage *qImg, uint8_t *mImg, int width, int heig
...
+    if (format == mlt_image_rgba64) {
+        *qImg = QImage(mImg, width, height, QImage::Format_RGBA64);
+        return;
+    }
+#endif
+
+    Q_ASSERT(format == mlt_image_rgba);
+
 #if QT_VERSION >= QT_VERSION_CHECK(5, 2, 0)
```
ddennedy: Another place to remove before Qt 5.2 code

ddennedy:
I made a little test project with multiple tracks using qtblend, 4K60 10-bit video source, image on V2, text with drop shadow, and fade in/out everything. Then, I ran it like this:
`time melt '/home/ddennedy/Videos/test-rgba64/test-rgba64.mlt' -consumer null terminate_on_pause=1 real_time=-4 mlt_image_format=rgb`

Next, I ran it with `rgba64`, and it was actually faster! I repeated it with the same result.

Finally, I ran it with the 2 `avfilter.colorspace av.threads=8...` in a crude manner. Now, that was a killer: over 2.5x slower, 1m13s to 3m9s.
However, that is a separate project, and the way I have been thinking about it, someone can opt into it for 8-bit as well. FYI, this filter does not have support for RGB, but it does support yuv420p10 and yuv444p10. yuv420p10 was very fast: 1m47s. Also, FFmpeg 8 will likely have some things that might change the approach.

src/framework/mlt_types.h
bmatherly: Done

src/framework/mlt_image.c
bmatherly: Done

src/framework/mlt_image.c
bmatherly: I think the main point is to avoid an addition or multiplication for each operation. I could increment j by 4, and use that. But "j" is used in the color calculation.

src/modules/avformat/filter_avcolour_space.c
bmatherly: The purpose of Duff's Device is manual loop unrolling. But compilers can do that for us now. In addition, a compiler might make other choices like vectorizing. But Duff's Device can make those other optimizations challenging for a compiler. My practice has been to write new code in the most intuitive way, and then only resort to manual optimization (like Duff's Device) if profiling shows that it makes an improvement with a specific compiler. I was on the fence for this one because I also like consistency and it would have been easy to copy/modify the duff's device code above. But in this case, it was very easy to rewrite.

src/modules/avformat/filter_avcolour_space.c
bmatherly: See my comment above.

src/modules/core/filter_brightness.c
bmatherly: Good idea. Done.

src/modules/core/filter_pillar_echo.c
bmatherly: Good catch. It is actually rgba32. Fixed now.

src/modules/core/producer_colour.c
bmatherly: Good idea. Done.

src/modules/qt/common.cpp
bmatherly: Done - in a separate commit in case it causes problems.

src/modules/qt/common.cpp
bmatherly: Done

bmatherly:
Thanks for the thorough review.

> I think the next most critical thing to port to rgba64 (different PR) before it can be used in something like Shotcut is transition_luma because it is not obvious to avoid them, unlike a filter that does not have the #10bit tag.

I agree with that, and I would add the SPR filter. I am eager to start thinking about color management. But I think I might go add rgba64 to those two services first. They can go in a separate PR if I do not finish them in time for this one. Just like you are working to get end-to-end GPU workflows, I am eager to get end-to-end rgba64 workflows. So feel free to call out other services that would benefit. I am on the fence about Frei0r.

ddennedy:
Changes look good, and I run tested it.

ddennedy (approved):

PR #1110, opened 2025-08-06
Allow arbitrary rotations by being able to specify an anchor point [qtblend]

balooii:
Hi!

I think it would be a nice addition for the qtblend filter to be able to set the rotation anchor point so you can rotate around any point you want.

I (mis)used a rect type parameter for this as I didn't find a point-equivalent parameter type and just used x and y component out of it.

If you're interested in this change I'd be happy to change it to something more appropriate if you can guide me :slightly_smiling_face: 

Here are some examples how it could be used:

Rotate around bottom-right corner (1,1)
```
melt -profile hdv_720_25p color:red length=100 -filter qtblend rect="100 100 200 200 1" rotation=90 rotate_anchor="1 1 0 0 0" -consumer avformat:test_bottomright.mp4
```

Rotate around some point outside the rectangle (-0.5, 0.5)
```
melt -profile hdv_720_25p color:red length=100 -filter qtblend rect="100 100 200 200 1" rotation=45 rotate_anchor="-0.5 0.5 0 0 0" -consumer avformat:test_outside.mp4
```

Rotate around center (0.5, 0.5), equivalent to rotate_center set to 1
```
melt -profile hdv_720_25p color:red length=100 -filter qtblend rect="100 100 200 200 1" rotation=45 rotate_anchor="0.5 0.5 0 0 0" -consumer avformat:test_center.mp4
```

Rotate around top-left corner (0, 0), equivalent to rotate_center set to 0
```
melt -profile hdv_720_25p color:red length=100 -filter qtblend rect="100 100 200 200 1" rotation=45 rotate_anchor="0.5 0.5 0 0 0" -consumer avformat:test_topleft.mp4
```

In Inkscape this could be done like so:

[inkscape_rotation_anchor.webm](https://github.com/user-attachments/assets/b87a8136-6bde-4543-bf3e-40ad5696a5e5)

### DIFF
--- a/src/modules/qt/filter_qtblend.cpp
+++ b/src/modules/qt/filter_qtblend.cpp
@@ -166,12 +166,22 @@ static int filter_get_image(mlt_frame frame,
     if (mlt_properties_get(properties, "rotation")) {
         double angle = mlt_properties_anim_get_double(properties, "rotation", position, length);
         if (angle != 0.0) {
-            if (mlt_properties_get_int(properties, "rotate_center")) {
+            if (mlt_properties_get(properties, "rotate_anchor")) {
+                mlt_rect anchor = mlt_properties_anim_get_rect(properties, "rotate_anchor", position, length);
+                // Use custom anchor point (x,y are normalized 0-1 coordinates) where 0, 0 is top left and 1, 1 is bottom right
+                // negative values are allowed so its possible to rotate around a point outside the rectangle
+                double anchor_x = anchor.x * rect.w;
+                double anchor_y = anchor.y * rect.h;
+                transform.translate(anchor_x, anchor_y);
+                transform.rotate(angle);
+                transform.translate(-anchor_x, -anchor_y);
+            } else if (mlt_properties_get_int(properties, "rotate_center")) {
+                // old style rotation (from center) to keep compatibility, equivalent to rotate_anchor = 0.5, 0.5
                 transform.translate(rect.w / 2.0, rect.h / 2.0);
                 transform.rotate(angle);
                 transform.translate(-rect.w / 2.0, -rect.h / 2.0);
             } else {
-                // old style rotation (from top left corner) to keep compatibility
+                // old style rotation (from top left corner) to keep compatibility, equivalent to rotate_anchor = 0, 0
                 transform.rotate(angle);
             }
             hasAlpha = true;
@@ -268,6 +278,10 @@ mlt_filter filter_qtblend_init(mlt_profile profile, mlt_service_type type, const
         filter->process = filter_process;
         mlt_properties properties = MLT_FILTER_PROPERTIES(filter);
         mlt_properties_set_int(properties, "rotate_center", 0);
+        
+        // Initialize rotate_anchor property with center position (0.5, 0.5)
+        mlt_rect default_anchor = {0.5, 0.5, 0.0, 0.0, 0.0};
+        mlt_properties_set_rect(properties, "rotate_anchor", default_anchor);
     } else {
         mlt_log_error(MLT_FILTER_SERVICE(filter), "Filter qtblend failed\n");
 


ddennedy:
> I (mis)used a rect type parameter

That is fine. A lot of rect properties in MLT do not utilize the opacity field, and that is OK. However, an alternative is `anchor_x` and `anchor_y`. I do not have a preference. I want you to explore which is more convenient for Kdenlive and reply with your decision. Thank you

balooii:
Thank you, I have reached out to JBM for another opinion. 

Aesthetically I think I would prefer the two double params than using a rect for a point but don't really have a opinion.
 
I haven't looked at the Kdenlive keyframe/param model too deeply and don't know the one from Shotcut at all so maybe there is a preference based on how it will be used there at some point.

ddennedy:
Previously I wrote that Shotcut does not use qtblend. Well, it does use the transition now for track blending; so, I deleted it. But it does not use it as a transform filter. I recently had to do some work on the Shotcut Corner Pin UI, which is based on 4 points, and Shotcut does not have a point parameter type. The underlying frei0r plugin uses 2 distinct number parameters to represent the point. I used rect params in its UI, which has a visual control. It was quite a bit of ugly extra work to coordinate the rect-as-point with the discrete numbers--unrelated to rect having more fields than necessary. So, my vote is to use rect in case I adopt this into Shotcut or port it to another filter. However, priority goes to Kdenlive devs.

j-b-m:
Thanks Dan for the input. I am also in favor of using a rect to handle position parameters. Reviewing this MR is on my todo list, hopefully I can move on with this in the coming weeks.

luzpaz:
Note: downstream MR https://invent.kde.org/multimedia/kdenlive/-/merge_requests/655 is dependent on this PR

luzpaz:
@j-b-m any chance you can review this ? Thanks :pray:

src/modules/qt/filter_qtblend.yml
j-b-m:
```suggestion
version: 3
```
I would update the filter version so the apps can ensure the feature is available

src/modules/qt/filter_qtblend.cpp
```diff
@@ -268,6 +279,10 @@ mlt_filter filter_qtblend_init(mlt_profile profile, mlt_service_type type, const
         filter->process = filter_process;
         mlt_properties properties = MLT_FILTER_PROPERTIES(filter);
         mlt_properties_set_int(properties, "rotate_center", 0);
+
+        // Initialize rotate_anchor property with center position (0.5, 0.5)
+        mlt_rect default_anchor = {0.5, 0.5, 0.0, 0.0, 0.0};
+        mlt_properties_set_rect(properties, "rotate_anchor", default_anchor);
```
j-b-m:
```suggestion
```

Initializing `rotate_anchor` breaks existing projects as it takes precedence over the older methods. Previous projects that did not have the `rotate_center` property set would break, using  rotate by center instead of top left corner.

src/modules/qt/filter_qtblend.yml
```diff
@@ -56,6 +56,17 @@ parameters:
...
+  - identifier: rotate_anchor
+    title: Rotation anchor point
+    type: rect
+    description: >
+      Keyframable anchor point for rotation. X and Y coordinates define the rotation center.
+      Values are normalized (0-1 for inside rectangle, negative or >1 for outside).
+      Default is center (0.5, 0.5). Takes precedence over rotate_center when set.
+    mutable: yes
+    animation: yes
+    default: "0.5 0.5 0 0 0"
```
j-b-m:
```suggestion
```

Remove default value, it is unset by default

j-b-m:
Unless someone sees an issue with it, I will merge my suggestion to remove the default value for `rotate_anchor`, and merge this MR to master in the next 1-2 days.

PR #1111, opened 2025-08-06
xml parser: remove deprecated entity substitution setup to fix #1083

### LINKED ISSUE #1083: Does not write file to target location if filename contains escaped &
Noticed this on Kdenlive but can reproduce it in Shotcut exactly the same so this appears to be an issue in mlt.

Both applications will create an .mlt file where the consumer target property is set to `/path/Black&amp;White.mp4` when in GUI it was specified as `/path/Black&White.mp4`. When melt does its thing it will write the file to `/path/Black&#38;White.mp4` not `/path/Black&White.mp4`.

Thank you for your work on this project.

balooii:
This fixes #1083 for me

Based on your comment pointing to libxml2:

* I compiled and tested a sample from libxml2 and could not reproduce the problem
* apparently this behavior of replacing &amp; with &  when parsing is called entity substitution
* found this term in producer_xml.c where we were calling  xmlSubstituteEntitiesDefault(1)
* saw that this is marked as deprecated and is not being used in the sample from libxml2
* removed it and it didn't fix the problem
* found that setting replaceEntities fixes it (the docs mention XML_PARSE_NOENT though)

Long story short, I have no idea what I'm doing but at least could confirm your tip regarding libxml2.
I would suggest you may take it from here :slightly_smiling_face:

### DIFF
--- a/src/modules/xml/producer_xml.c
+++ b/src/modules/xml/producer_xml.c
@@ -2065,7 +2065,6 @@ mlt_producer producer_xml_init(mlt_profile profile,
 
     // Setup libxml2 SAX parsing
     xmlInitParser();
-    xmlSubstituteEntitiesDefault(1);
     // This is used to facilitate entity substitution in the SAX parser
     context->entity_doc = xmlNewDoc(_x("1.0"));
     if (is_filename)
@@ -2114,6 +2113,8 @@ mlt_producer producer_xml_init(mlt_profile profile,
         free(sax);
         return NULL;
     }
+
+    xmlcontext->replaceEntities = 1;
 
     // Reset the stack.
     mlt_deque_close(context->stack_service);


src/modules/xml/producer_xml.c
```diff
@@ -2115,6 +2114,8 @@ mlt_producer producer_xml_init(mlt_profile profile,
         return NULL;
     }
 
+    xmlcontext->replaceEntities = 1;
```
ddennedy, line 2117: Use `xmlCtxtUseOptions(xmlcontext, XML_PARSE_NONET)` instead:
https://gnome.pages.gitlab.gnome.org/libxml2/html/parser_8h.html#ae47f76b45b93114d3fc9d6f627f061ef

ddennedy (changes requested):

src/modules/xml/producer_xml.c
ddennedy: Actually, my test shows that is not working, and neither does your change to set the struct member! Basically, the changes libxml2 does not provide smooth backwards-compatibility and we need keep the deprecated function call for the older versions of libxml2. MLT does not aggressively drop support for older versions of dependencies. So, keep the deprecated function call and use  `XML_PARSE_NONET` if that works for you, or keep your `xmlcontext->replaceEntities = 1;` if that does not work.

balooii: As it seems to matter I am on Arch Linux with libxml2 2.14.5-1.

Hmm, if you cannot reproduce it then it's probably not a good idea to merge this anyway. Not confident in this code change.

Will close this PR for now and try to triage this a bit more, maybe with older libxml2 versions and my Ubuntu VM or so. If I have more information will mention it in the issue or maybe create another MR.

Cheers!

ddennedy:
What?! It is a good idea to fix it. Please test with this change only:
```c
--- a/src/modules/xml/producer_xml.c
+++ b/src/modules/xml/producer_xml.c
@@ -2079,6 +2079,7 @@ mlt_producer producer_xml_init(mlt_profile profile,
         free(sax);
         return NULL;
     }
+    xmlCtxtUseOptions(xmlcontext, XML_PARSE_NONET);
 
     // Parse
     sax_orig = xmlcontext->sax;
```
I am tempted to push it as the fix, but a confirmation would be nice.

ddennedy:
OOPS   Should use `XML_PARSE_NOENT` instead of `XML_PARSE_NONET` Geez that's not error prone! LMFAO it still does not work to remove the old function call with older libxml2.

PR #1114, opened 2025-08-10
Use CMake's find_package for fftw3

jlskuz:

~Luckily fftw3 uses CMake itself and hence provides everything need for that~ Edit: it also allows to be build without CMake and does not provide a FFTW3Config.cmake etc. then. So we need to provide a FindFFTW3.cmake ourself


jlskuz:
Draft because I need to test again after the latest changes

jlskuz:
Tested now. The CI failure seems unrelated to my change

ddennedy:
This broke Shotcut's Windows and macOS builds.
Windows: https://github.com/mltframework/shotcut/actions/runs/16985767816/job/48154243595
> ninja: error: '/mingw64/lib/libfftw3.dll.a', needed by 'out/lib/mlt/libmltplus.dll', missing and no known rule to make it

The file `/mingw64/lib/libfftw3.dll.a'` does exist.  After removing `QUIET` in `FindFFTW3.cmake` I do not see any more information.

ddennedy:
On macOS it says:
```
 -- Found PkgConfig: /opt/local/bin/pkg-config (found version "0.29.2")
CMake Error at /opt/local/lib/cmake/fftw3/FFTW3Config.cmake:13 (include):
  include could not find requested file:

    /opt/local/lib/cmake/fftw3/FFTW3LibraryDepends.cmake
Call Stack (most recent call first):
  cmake/FindFFTW3.cmake:5 (find_package)
  CMakeLists.txt:203 (find_package)

```
`/opt/local` is the prefix where MacPorts is installed.
```
ddennedy@Dans-Mac-mini build % pkg-config --modversion fftw3    
3.3.10
```
`/opt/local/lib/cmake/fftw3/FFTW3Config.cmake` reads:
```
# defined since 2.8.3
if (CMAKE_VERSION VERSION_LESS 2.8.3)
  get_filename_component (CMAKE_CURRENT_LIST_DIR ${CMAKE_CURRENT_LIST_FILE} PATH)
endif ()

# Allows loading FFTW3 settings from another project
set (FFTW3_CONFIG_FILE "${CMAKE_CURRENT_LIST_FILE}")

set (FFTW3_LIBRARIES fftw3)
set (FFTW3_LIBRARY_DIRS /opt/local/lib)
set (FFTW3_INCLUDE_DIRS /opt/local/include)

include ("${CMAKE_CURRENT_LIST_DIR}/FFTW3LibraryDepends.cmake")

if (CMAKE_VERSION VERSION_LESS 2.8.3)
  set (CMAKE_CURRENT_LIST_DIR)
endif ()
```

https://github.com/FFTW/fftw3/issues/130

I am not certain I can upgrade on macOS just for this. It is temperamental to update to make its universal CPU architecture build dependencies. That is a separate matter. msys2 I am using for Windows is the same version of FFTW, which is the latest. Perhaps they have patched it. I commented out the include in that file on macOS, as suggested in a comment, but then it fails in a different way. I prefer that this be changed to try `pkg-config` first and `find_package` as a fallback.

ddennedy:
This worked for me on Windows:
```cmake

# FFTW3 supports different build systems and the necessary
# bits for find_package are only installed if CMake was used

# First try to find FFTW3 using pkg-config
find_package(PkgConfig)

if(PkgConfig_FOUND)
    pkg_check_modules(FFTW3 IMPORTED_TARGET fftw3)

    if(FFTW3_FOUND)
        add_library(FFTW3::fftw3 INTERFACE IMPORTED)

        if(FFTW3_INCLUDE_DIRS)
            set_property(TARGET FFTW3::fftw3 PROPERTY
                    INTERFACE_INCLUDE_DIRECTORIES "${FFTW3_INCLUDE_DIRS}")
        endif()
        if(FFTW3_LINK_LIBRARIES)
            set_property(TARGET FFTW3::fftw3 PROPERTY
                    INTERFACE_LINK_LIBRARIES "${FFTW3_LINK_LIBRARIES}")
        endif()
        if(FFTW3_LDFLAGS_OTHER)
            set_property(TARGET FFTW3::fftw3 PROPERTY
                    INTERFACE_LINK_OPTIONS "${FFTW3_LDFLAGS_OTHER}")
        endif()
        if(FFTW3_CFLAGS_OTHER)
            set_property(TARGET FFTW3::fftw3 PROPERTY
                    INTERFACE_COMPILE_OPTIONS "${FFTW3_CFLAGS_OTHER}")
        endif()
    endif()
endif()

if(NOT TARGET FFTW3::fftw3)
    # Fallback to find_package
    find_package(FFTW3 CONFIG)
endif()
```

jlskuz:
Thanks for the details. Unfortunately switching priorities of cmake and pkg-config doesn't solve anything on MSVC. This is because pkg-config is available in MSVC, but it creates wrong dependency link flags etc.

I hope #1130 will satisfy all requirements, please take a look

PR #1116, opened 2025-08-10
Add libgen.c and libgen.h for MSVC

jlskuz:
libgen is POSIX and not part of standard C hence we need to maintain our own files similar to a few others we already have in src/win32

The libgen.c has been copied from MinGW.


bmatherly:
Will it get compiled three times? If so, is there a way to include it at a higher level so it is only compiled once?

jlskuz:
Yeah, that can be improved.

jlskuz:
Updated and tested again

ddennedy (approved):

PR #1117, opened 2025-08-10
Use CMake's find_package for SDL2

jlskuz:

See https://cmake.org/cmake/help/latest/module/FindSDL.html#examples

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,7 @@ if(WIN32)
   endif()
 endif()
 
-pkg_check_modules(sdl2 IMPORTED_TARGET sdl2)
+find_package(SDL2)
 
 if(MOD_QT OR MOD_QT6 OR MOD_PLUS)
   pkg_check_modules(FFTW IMPORTED_TARGET fftw3)
@@ -411,7 +411,7 @@ if(MOD_SDL1)
 endif()
 
 if(MOD_SDL2)
-  if(TARGET PkgConfig::sdl2)
+  if(SDL2_FOUND)
     list(APPEND MLT_SUPPORTED_COMPONENTS sdl2)
   else()
     set(MOD_SDL2 OFF)
--- a/src/melt/CMakeLists.txt
+++ b/src/melt/CMakeLists.txt
@@ -6,8 +6,8 @@ target_link_libraries(melt PRIVATE mlt Threads::Threads)
 
 target_compile_definitions(melt PRIVATE VERSION="${MLT_VERSION}")
 
-if(TARGET PkgConfig::sdl2 AND NOT ANDROID)
-    target_link_libraries(melt PRIVATE PkgConfig::sdl2)
+if(SDL2_FOUND AND NOT ANDROID)
+    target_link_libraries(melt PRIVATE SDL2::SDL2)
     target_compile_definitions(melt PRIVATE HAVE_SDL2)
     if(MINGW)
         target_link_libraries(melt PRIVATE ${MLT_PTHREAD_LIBS})
--- a/src/modules/sdl2/CMakeLists.txt
+++ b/src/modules/sdl2/CMakeLists.txt
@@ -12,7 +12,7 @@ add_custom_target(Other_sdl2_Files SOURCES
 
 target_compile_options(mltsdl2 PRIVATE ${MLT_COMPILE_OPTIONS})
 
-target_link_libraries(mltsdl2 PRIVATE mlt Threads::Threads PkgConfig::sdl2)
+target_link_libraries(mltsdl2 PRIVATE mlt Threads::Threads SDL2::SDL2)
 if(MSVC)
   target_link_libraries(mltsdl2 PRIVATE PThreads4W::PThreads4W)
 else()


ddennedy (changes requested):
Do not take away my pkg-config! I hate dealing with find_package.

jlskuz:
I see your point when it comes to maintaining custom `Find*.cmake`. The advantage of `find_package` is however that it is cross-platform compatible and in the case of SDL it is very convenient to use. As you can see in this patch it is the same amount of code.

I wonder what other options we have...

If you think it is better I could start to guard all the code by `if(MSVC)` but that means we will essential maintain the same code twice. Are there other options I miss?

ddennedy:
It seems like you are involved quite a bit with the building of kdenlive and maybe other KDE projects. So, it surprises me that you are not so concerned about breaking builds! You are being optimistic, but the pain I have experienced with maintaining builds makes me pessimistic. `find_package` is like magic to me. I frequently struggle to tell it how to find things because it often fails (mainly with things not installed to system?). How do I know this is going to work? I would have to make changes to my builds in branches and run all 4 of them to test this. Is supporting MSVC for KDE like a fun hobby for you and other KDE contributors? Because it seems masochistic to me. Is it because Qt WebEngine uses Chromium, which does not work on mingw?

I think the main option I have seen is that when `find_package` fails to find something it turns to `pkg-config`. Yes, that is as little more code.

jlskuz:
> It seems like you are involved quite a bit with the building of kdenlive and maybe other KDE projects. So, it surprises me that you are not so concerned about breaking builds!



> You are being optimistic, but the pain I have experienced with maintaining builds makes me pessimistic. find_package is like magic to me. I frequently struggle to tell it how to find things because it often fails (mainly with things not installed to system?).

I guess that's an important point. For me it is basically the contrary: since I started my involvment in software packaging I mainly work with CMake based projects and are familiar with it. Hence the magic things for me are pkg-config, autotools, etc. In my experience those are also the ones that cause trouble for cross-platform support. However I guess it really depends on the projects you work with in your daily life.

> How do I know this is going to work? I would have to make changes to my builds in branches and run all 4 of them to test this.

Testing changes for all upstream consumers of a downstream library is a challenge and almost impossible. It will happen that you break things. In fact https://github.com/mltframework/mlt/commit/64aa54375d2816d82210144bb5cebaf1ba911aa5 broke Kdenlive's MinGW builds and we are patching MLT to revert this change.

In general I am confident about my changes because

1. I frequently switch between platforms to test them. (Our setup with KDE Craft is quite convenient in that regard. I have a local setup that replicates the CI/CD pipelines accurately)
2. some changes are indeed hard to test until deployed, but we have a good coverage of CI and CD with stable and nightly builds which I observer and if something breaks, I will just fix it.

I don't monitor your pipelines, but if something breaks because of my changes we can a) revert it (I mean this is master and not immediately going into a stable release) and b) I am available to fix it if you tell me about it. Additionally if you provide me with specific requests about what I should test, I can certainly try to do that as part of my usual testing.

Also if it gives you a better feeling I can apply my patches to our nightly CD on all platforms before we merge it here.

> Is supporting MSVC for KDE like a fun hobby for you and other KDE contributors? Because it seems masochistic to me. Is it because Qt WebEngine uses Chromium, which does not work on mingw?

For many KDE people contributing is a hobby, yes. Fun aside, the entire KDE infrastructure does support MSVC quite well so the pain is not bigger than for most other non-Linux platforms. MinGW is also supported by the KDE infrastructure, but not that well (eg. there is only CD and no CI). I don't know the historic reasons for this, but there are several advantages of MSVC on Windows. The most important ones are:
- MSVC is from Microsoft and hence the native solution on Windows. This has advantages for the development and debugging experience 
- Deployment to the Microsoft Store is easier via appx files and it supports automated updates for user, which exe installer distribution via the store does not
- Qt WebEngine is surely another important aspect

> I think the main option I have seen is that when find_package fails to find something it turns to pkg-config. Yes, that is as little more code.

Okay so what you suggest is to keep using `find_package` but fallback to `pkg-config`?

That would work. There are two options however:
1. Do it in place in the CMakeLists.txt
2. Wrap it into a `Find*.cmake` like I did in https://github.com/mltframework/mlt/pull/1114/files#diff-0721b4d8350b09abf73114fb87d299694aaa6ca771e17adc848b1bc627b297de

ddennedy:
Thank you for the engaging discussion and perspective! I will merge this as-is, and we will see what happens. I hope it will work since Shotcut is not building its own SDL2. If something breaks I will see if I can fix it in the build script if it does not take too long. We can add in a pkg-config fallback as a last resort.

ddennedy:
Shotcut/Linux build failed on SDL2

https://github.com/mltframework/shotcut/actions/runs/16874329171/job/47795242849#step:4:54218

jlskuz:
The problem seems to be this: the `SDL2::SDL2` target is only available in SDL 2.0.12 while Ubuntu 20.04 has 2.0.10

There are several options to fix this:
1. Update your docker image to Ubuntu 22.04 (as you already considered in #1119)
2. Continue to use 20.04 but build SDL2 yourself, instead of using the apt package
3. Add fallback in MLT to `${SDL2_LIBRARIES}` like :
```
if(TARGET SDL2::SDL2)
    target_link_libraries(mytarget PRIVATE SDL2::SDL2)
else()
    target_include_directories(mytarget PRIVATE ${SDL2_INCLUDE_DIRS})
    target_link_libraries(mytarget PRIVATE ${SDL2_LIBRARIES})
endif()
```
4. Add a fallback in MLT to pkg-config as discussed yesterday.

Just let me know which option you prefer!

ddennedy:
I will upgrade the Linux target to 22.04.

PR #1118, opened 2025-08-10
Use CMake's find_package for ffmpeg/libav

jlskuz:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,13 +220,18 @@ endif()
 
 if(MOD_AVFORMAT)
   pkg_check_modules(libavformat IMPORTED_TARGET libavformat)
-  pkg_check_modules(libswscale IMPORTED_TARGET libswscale)
-  pkg_check_modules(libavutil IMPORTED_TARGET libavutil)
-  if(TARGET PkgConfig::libavformat AND TARGET PkgConfig::libswscale AND TARGET PkgConfig::libavutil)
-    pkg_check_modules(libavcodec IMPORTED_TARGET libavcodec)
-    pkg_check_modules(libavfilter IMPORTED_TARGET libavfilter)
-    pkg_check_modules(libavdevice IMPORTED_TARGET libavdevice)
-    pkg_check_modules(libswresample IMPORTED_TARGET libswresample)
+  find_package(FFmpeg COMPONENTS
+    AVFORMAT
+    SWSCALE
+    AVUTIL
+  )
+  if(AVFORMAT_FOUND AND SWSCALE_FOUND AND AVUTIL_FOUND)
+    find_package(FFmpeg COMPONENTS
+      AVCODEC
+      AVFILTER
+      AVDEVICE
+      SWRESAMPLE
+    )
     list(APPEND MLT_SUPPORTED_COMPONENTS avformat)
   else()
     set(MOD_AVFORMAT OFF)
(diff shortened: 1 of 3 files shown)


ddennedy:
Your description is funny to me because I frequently run into issues with `find_package` and I successfully use `pkg-config` on Windows and macOS. I am a little worried about maintaining somewhat complex cmake code as it is not a primary concern. This appears to break my preference for using `pkg-config` on macOS and Windows msys2. I require that.

jlskuz:
macOS is Unix and Msys2 is MinGW and has a lot in common with Unix, so yes it works well there.

It's indeed not ideal that we need to maintain the `FindFFmpeg.cmake` here. On the bright side I can say at least that I copied it from KDE where it is successfully used for several apps since many years.

For reference the problem with pkg-config on MSVC is that it fails at linking in several cases: It either does not find dependency libs (such as zlib) during linking or wrongly tries to pick up static versions where it should use shared libs.

ddennedy:
Well, this one I am very worried about because my Shotcut and melt builds build FFmpeg and install it a non-standard location. How I do tell this where to find it? With pkg-config it is as simple as setting an environment variable to the location where ffmpeg installed its .pc file.

jlskuz:
`-DFFmpeg_ROOT=/my/custom/location` as a cmake flag should work. See https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html

> Well, this one I am very worried about because my Shotcut and melt builds build FFmpeg and install it a non-standard location.

Is there a way that I can try these scripts myself to test changes?

ddennedy:
This repo has a Dockerfile that runs a build-melt script from the mlt-scripts repo. You do not need to use docker but telling that it exists as well as a GitHub Actions workflow to build it on release.

jlskuz:
`FFmpeg_ROOT` envvar should also work, see https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure for the full lookup order

jlskuz:
I tested this with the `build-melt.sh` script inside a `ubuntu:jammy` container now and it works without further tweaks. Since you set `CMAKE_INSTALL_PREFIX` to the location where FFmpeg is installed, it knows already where to look

ddennedy:
Now, for my local development on macOS where I build FFmpeg myself and install to `$HOME/opt` fails to locate FFmpeg dependencies even though I have defined `FFmpeg_ROOT` both as a CMake and env var, set to `$HOME/opt`. 😠 CMake `find_package` sucks! The cmake run does not fail; it is the compilation.

ddennedy:
```
FAILED: src/modules/avformat/CMakeFiles/mltavformat.dir/factory.c.o 
/Library/Developer/CommandLineTools/usr/bin/cc -Dmltavformat_EXPORTS -I/Users/ddennedy/src/mlt/src/framework/.. -std=gnu11 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.7 -fPIC -MD -MT src/modules/avformat/CMakeFiles/mltavformat.dir/factory.c.o -MF src/modules/avformat/CMakeFiles/mltavformat.dir/factory.c.o.d -o src/modules/avformat/CMakeFiles/mltavformat.dir/factory.c.o -c /Users/ddennedy/src/mlt/src/modules/avformat/factory.c
/Users/ddennedy/src/mlt/src/modules/avformat/factory.c:39:10: fatal error: 'libavcodec/avcodec.h' file not found
#include <libavcodec/avcodec.h>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
And `CMakeCache.txt` does contain `AVCODEC_INCLUDE_DIRS:PATH=/Users/ddennedy/opt/include`

ddennedy:
I resolved it for now by setting `CFLAGS` and `LDFLAGS`

PR #1119, opened 2025-08-10
Raise C++ standard to C++20

jlskuz:
Motivation:
In a few places, mostly code for the GPS filters, we use C++20 features already. Namely designated initializers aka. aggregate initialization (see https://en.cppreference.com/w/cpp/language/aggregate_initialization.html )

MSVC is very picky about this and hence fails with "error C7555: use of designated initializers requires at least "/std:c++20"."


ddennedy:
Shotcut portable and AppImage builds use gcc 10 on Ubuntu 20.04, and I am fairly certain this will break that. Portable builds are intended to provide a fair amount of backwards compatibility (even though many projects seem to be neglectful in that regard). So, how old should that extend? Ubuntu 20.04 recently passed end-of-support. I have been thinking about that because of a Wayland and a different bug that will be resolved by upgrading the Qt version. If 22.04 and its glibc version  is the new minimum, that is only a little over 3 years. That is not so forgiving IMHO. I think 5 years is more reasonable. I was thinking to wait until 26.04 arrives. 

I see my build script is compiling MLT with ` -std=c++11`. Let me try this branch with `c++20` and see what happens...

jlskuz:
The internet says C++20 is full supported since GCC 10, but testing is always better...

bmatherly:
GCC 11 and 13 are also available for Ubuntu 20.04. Maybe we could use those.

I would support switching to Ubuntu 22.04. By the time we get it tested and released, it would be 3.5 years old. That's a pretty good run, I think.

PR #1121, opened 2025-08-12
Make deps for modules that are ON required

jlskuz:
Related to #1115 

This uncovered that the modules QT6 and SPATIALAUDIO which, are ON by default, were not build by CI. This has been fixed by installing the deps.


ddennedy (approved):

jlskuz:
Done. Not sure if it just takes a while until a new pipeline is started or if you as a maintainer/reviewer need to re-trigger because I force pushed? (I think I saw such a requirement with another project on GitHub a while back)

ddennedy:
It is failing on "E: You must put some 'deb-src' URIs in your sources.list" Do you use a web browser to look at GitHub?

jlskuz:
Yeah, I do, but I am not very used to GitHub because I spend most of my time on GitLab. I checked right after I pushed, but it failed so fast that I didn't notice… 🤦‍♂️

ddennedy:
Still incompatible! Change the module default OFF

jlskuz:
Should I open another PR to update the Ubuntu version on CI? If so `ubuntu-latest` or `ubunt-24.04`?

PR #1122, opened 2025-08-12
Make all FFmpeg modules required

jlskuz:
As discussed in #1115


ddennedy (approved):

ddennedy:
For some reason I have not figured out, this breaks opening files--on any OS. Build and `melt -query` works fine, but not opening any files or listing codecs on the consumer `melt color: -consumer avformat vcodec=list`

PR #1124, opened 2025-08-14
Fix a few more ifdefs in MOD_AVFORMAT

jlskuz:
I overlooked them because the used a different syntax

Fixup for #1122


jlskuz:
BTW looking at the pipeline it seems all builds except `build-cmake` are skipped. Is that intended?

ddennedy:
I fixed the workflow issue. Thanks to fix this up!

PR #1126, opened 2025-08-14
Fix preprocessor directive in function-like macro args

jlskuz:
Fixes MSVC compiler error C5101: "use of preprocessor directive in function-like macro argument list is undefined behavior"

This behaviour according to C Standard 6.10.3 Macro replacement, ¶11


src/modules/avformat/link_avfilter.c
```diff
@@ -827,14 +827,17 @@ static int link_get_audio(mlt_frame frame,
...
             || *samples != pdata->avoutframe->nb_samples
             || *frequency != pdata->avoutframe->sample_rate) {
-            mlt_log_error(self,
-                          "Unexpected return format c %d->%d\tf %d->%d\tf %d->%d\n",
-                          *channels,
+
+            int expected_channels;
 #if HAVE_FFMPEG_CH_LAYOUT
-                          pdata->avoutframe->ch_layout.nb_channels,
+            expected_channels = pdata->avoutframe->ch_layout.nb_channels;
```
bmatherly: Move this up to line 822 so that you can remove the same #if up there.

bmatherly (changes requested):

src/modules/avformat/link_avfilter.c
jlskuz: That will only work if we put the `expected_channels` channels var one scope higher because the `if` condition is not finished within the `#if ... #endif` block. It think it matters much, but doesn't seem the cleanest solution either.

I'll just do whatever you prefer...

bmatherly: I prefer to move it up in scope. Otherwise, a future developer (me) will have to try to remember why we did it differently in two different places. Thanks for humoring me.

src/modules/avformat/link_avfilter.c
```diff
@@ -819,22 +819,25 @@ static int link_get_audio(mlt_frame frame,
             break;
         }
 
-        // Sanity check the output frame
+        int expected_channels;
```
bmatherly: Sorry. I wasn't clear. I'm looking for this:

```
#if HAVE_FFMPEG_CH_LAYOUT
        int expected_channels = pdata->avoutframe->ch_layout.nb_channels;
#else
        int expected_channels = pdata->avoutframe->channels;
#endif
        // Sanity check the output frame
        if (*channels != expected_channels
            || *samples != pdata->avoutframe->nb_samples
            || *frequency != pdata->avoutframe->sample_rate) {

            mlt_log_error(self,
                          "Unexpected return format c %d->%d\tf %d->%d\tf %d->%d\n",
                          *channels,
                          expected_channels,
                          *samples,
                          pdata->avoutframe->nb_samples,
                          *frequency,
                          pdata->avoutframe->sample_rate);
            error = 1;
            break;
        }
```

PR #1127, opened 2025-08-14
Use define instead of static const int

jlskuz:
Fixes MSVC Error C2057.

See https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2057


bmatherly:
LGTM

PR #1131, opened 2025-08-16
Fix void pointers arithmetic

jlskuz:
Pointer arithmetic with void pointers works in GNU C by considering the size of the void as 1 hence it worked.

However the C standard doesn't allow pointer arithmetic with void pointers. Hence MSVC fails with Error C2036.

Please that I am not too confident when it comes to "low level" pointer related C stuff. As far as I understand this is correct and should be low risk, but please review extra carefull anyways.


ddennedy:
We try to avoid it. Thanks

PR #1134, opened 2025-08-18
Add rgba and rgba formats to transition luma.

bmatherly:
I added rgba and rgba64 formats to transition luma. Everything seems to test find in Shotcut when i force it to use those image formats.

Posting for review here because there are a few things I am unsure of. I notice that the composite line function branches off to some code shared with transition composite. I did not recreate that code. So I am unsure if this will perfectly match those scenarios.

This is the specific case that I did not duplicate:
https://github.com/mltframework/mlt/blob/ff1f0d7215bf6ebde7291cd2f14f34f738eb88eb/src/modules/core/transition_luma.c#L170

I only implemented the case for `is_translucent && alpha_over`
Shotcut  sets "alpha_over". And i do not really know the significance of "is_translucent".

### DIFF
--- a/src/modules/core/transition_luma.c
+++ b/src/modules/core/transition_luma.c
@@ -1,6 +1,6 @@
 /*
  * transition_luma.c -- a generic dissolve/wipe processor
- * Copyright (C) 2003-2024 Meltytech, LLC
+ * Copyright (C) 2003-2025 Meltytech, LLC
  *
  * Adapted from Kino Plugin Timfx, which is
  * Copyright (C) 2002 Timothy M. Shead <tshead@k-3d.com>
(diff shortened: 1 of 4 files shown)


src/modules/core/transition_luma.c
```diff
@@ -137,8 +147,6 @@ static inline int dissolve_yuv(mlt_frame frame,
     uint8_t *alpha_dst;
     int mix = weight * (1 << 16);
 
-    if (mlt_properties_get(&frame->parent, "distort"))
-        mlt_properties_set(&that->parent, "distort", mlt_properties_get(&frame->parent, "distort"));
```
ddennedy, line 141: Why is this removed? I can see it was added in 4a39b72c and transition_composite can set a frame distort and that it can use the luma transition. Safer to not remove unrelated.

ddennedy:
I am fairly certain that `composite_line_yuv` is only used for legacy purposes such as by filter_luma, which should be deprecated (it is not yet). But there is a chance that this is exposed through Kdenlive for a dissolve not using alpha_over.

src/modules/core/transition_luma.c
ddennedy: Nevermind, I found it moved way below.

bmatherly: I moved it out to line 991 so that it would not have to be duplicated in every function for every format.

https://github.com/mltframework/mlt/pull/1134/files#diff-be1b84f01d456bb89ecbc9cca45a12eb5edeb698f6fab1c4b9fdc015ad512836R991

bmatherly: Besides this change, I tried to preserve all the code that operates on yuv422. There is quite a bit of consolidation/refactoring that could be done to make more code common and shared. But I want to balance cleanup with regression risk since we don't have unit tests for all these.

PR #1135, opened 2025-08-20
feat: Add comprehensive support for Windows/MSVC builds

sammiler:
This pull request introduces comprehensive support for building the MLT framework on Windows using the MSVC compiler. It significantly improves cross-platform compatibility and enables a native, modern development workflow for Windows developers.

This is achieved through a series of systematic improvements to the build system, the introduction of a robust compatibility layer for POSIX APIs, proper DLL symbol management, and numerous code-level adjustments for MSVC compatibility. Modern CMake features, particularly **CMake Presets**, are leveraged to provide a clean, maintainable, and non-intrusive implementation that does not affect existing build configurations for Linux or macOS.

#### 1. Build System & Dependency Management
*   **Git Submodules for Dependencies:** `vid.stab` and `frei0r` are now included as Git submodules. The CMake logic has been enhanced to automatically build them from the submodules if they are not found via `pkg-config`, streamlining dependency management on Windows.
*   **Modern CMake Presets:** A new set of `CMakePresets.json` configurations has been introduced to standardize the build process on Windows. This includes presets for Debug and Release builds, using **Ninja** as the default generator for improved performance.
*   **Clean Vcpkg Integration:** The presets are configured to integrate with `vcpkg` for dependency management. Hardcoded paths have been avoided; users can specify their `vcpkg` toolchain file locally via `CMakeUserPresets.json` or the `VCPKG_ROOT` environment variable.
*   **Platform-Specific Options via Presets:** Modules that are not yet compatible with MSVC (e.g., Movit) are now gracefully disabled *only* within the Windows presets, leaving the default configuration for other platforms untouched.

#### 2. Cross-Platform Compatibility Layer
*   **Centralized Compatibility Header:** A new shared header (`mlt_windows_compat.h`) has been created to centralize all POSIX-to-Windows API wrappers and definitions (e.g., `gettimeofday`, `unistd.h` functions, file descriptors, `alloca`). This replaces scattered `#ifdef` blocks, improving code readability and maintainability.
*   **Conditional Linking:** CMake now automatically links against necessary libraries for MSVC builds, such as `pthreads-win32` and MSVC-specific math libraries, resolving linker errors.

#### 3. API Visibility & Module Integration
*   **DLL Export/Import Macro:** A unified `MLT_PUBLIC_API` macro has been introduced to correctly handle `__declspec(dllexport)` and `__declspec(dllimport)` for shared library builds on Windows. All public APIs and shared symbols have been updated to use this macro.
*   **Module Fixes and Enablement:** The OpenCV module is now enabled by default for Windows builds, and compatibility issues related to its `unistd.h` usage have been resolved. Integration with `vid.stab` has also been fixed to handle MSVC-specific include paths.

#### 4. Code-Level Fixes and Refinements
*   **Data Alignment:** Cross-compiler macros are now used to ensure proper 16-byte alignment of key structs, preventing potential crashes and improving compatibility with external assembly libraries.
*   **Header Conflict Resolution:** A pre-emptive definition of `_WINSOCKAPI_` is added before including `windows.h` to prevent common redefinition errors with other libraries.
*   **Correct Pointer Arithmetic:** Pointer calculations for image buffers have been corrected to use byte-wise arithmetic, ensuring correct memory access and avoiding undefined behavior.


.gitmodules
```diff
@@ -2,3 +2,12 @@
...
 	ignore = dirty
+
+[submodule "external/vid.stab"]
+	path = external/vid.stab
+	url = https://github.com/georgmartius/vid.stab.git
+	branch = master
+[submodule "external/frei0r"]
+	path = external/frei0r
+	url = https://github.com/dyne/frei0r.git
+    branch = master
```
ddennedy, line 13: Remove these git submodules. I guess you added them because there is no vcpkg for them, but that needs to be solved a different way. Git submodules are not generally used for dependencies. Glaxnimate is an exception because it has no library.

ddennedy (changes requested):
Can you please also try to add a GitHub Actions workflow that builds this?

.gitmodules
ghost: Got it, thanks @ddennedy! Makes sense. I'll remove the submodules and will start looking into a proper vcpkg solution for the dependencies. I'll also work on the GitHub Actions workflow.
Will tackle it tomorrow. Thanks for the guidance!

eszlari: Have you considered https://cmake.org/cmake/help/latest/module/FetchContent.html ?

ddennedy: That is a good idea. I will accept that if the default to use it is on when vcpkg is on.

jlskuz: FetchContent sounds like a good stop gap solution, but the proper fix would be to submit packages to vcpkg

src/framework/mlt_api.h
jlskuz, line 1: https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html should be used here instead of a custom implementation.

The export/API improvement is is one change I'd prefer to have in a separate PR on its own.

jlskuz (changes requested):

jlskuz:
Thanks! Given the size of this change I can't review this all at once. To ensure good testing, convenient reviews, flexiblity in accepting patches and clean git history I suggest splitting a few things to separate pull requests. I suggest starting with the API export change as that touches a lot of files and hence will remove quite some noise from this PR.

src/framework/mlt_api.h
ghost: Thanks for the great feedback! I understand the need to split this up. I'll close this PR and resubmit the changes as smaller, focused PRs, starting with the GenerateExportHeader update.

ddennedy:
Thank you for your contribution.

.gitmodules
ghost: Thank you for the suggestion to use FetchContent. I've looked into it, and it seems both vid.stab and frei0r require some modifications to their CMake scripts to be integrated. I see two ways to handle this:
Using Patches: I can create .patch files and apply them with FetchContent's PATCH_COMMAND.
Using a Fork: I can fork the repositories, apply the changes on a branch, and point FetchContent to my fork.
The Fork approach seems cleaner and more maintainable to me. Would you have a preference between these two options? I want to make sure I'm heading in a direction you're comfortable with.

jlskuz: What kind of modification is needed? Ideally the modification will be submitted upstream. At least frei0r is under our control as @ddennedy has write permission on the repository.

sammiler: Hi @eszlari, thank you for looking into this!

That's a great question. The modifications needed for frei0r and vid.stab are not related to their source code at all, but rather to how their CMakeLists.txt files handle paths when they are consumed as a subproject (either via add_subdirectory or FetchContent).

The Core Issue: Path Resolution

The issue is a common one in older CMake projects: their scripts often use CMAKE_SOURCE_DIR to locate files. When MLT includes them as a subproject, CMAKE_SOURCE_DIR still points to MLT's top-level directory, not the dependency's own source directory. This causes their build scripts to fail when looking for their own files.

The fix is straightforward: patch their CMakeLists.txt to use CMAKE_CURRENT_SOURCE_DIR instead, which correctly points to their respective source trees. It's purely a build-system integration fix. While these patches could be submitted upstream, it might not be necessary given the overall strategy I'm proposing.

My Proposed Solution: A Two-Scenario Approach

My plan is to separate the build experience into two distinct user scenarios, which avoids complexity for most developers and provides a robust solution for those who need the optional modules.

For local/native builds (the main PR): The default CMakePresets will have optional modules like vid.stab and frei0r disabled by default. This provides the cleanest and fastest build experience for developers who just want to work on MLT's core. They won't need to deal with these dependencies at all.
For vcpkg users (the complete solution): The full power of MLT will be unlocked via vcpkg. I have developed a comprehensive vcpkg port for MLT that handles everything automatically. In this scenario:
The vcpkg port treats vid.stab and frei0r as proper dependencies.
It downloads their source, applies the necessary CMake path patches internally, and builds them correctly.
It then builds MLT with all the required module flags enabled.
Why this approach is better:

Clean & Simple for Everyone: Local developers get a minimal, fast build. vcpg users get a powerful, one-command installation (vcpkg install mlt[frei0r,vidstab]).
No Need for FetchContent or Submodules in MLT's core CMake: This keeps MLT's build system lean and free of dependency-fetching logic.
Encapsulates Complexity: All the "dirty work" (patching, dependency management) is perfectly encapsulated within the vcpkg port, where it belongs.
So, to answer your question directly: my current method is to provide these modules exclusively through a dedicated vcpkg port, where the necessary patches are handled cleanly. This seems like the most robust and user-friendly path forward.

What are your thoughts on this strategy？

ghost: Hi @jlskuz, thank you for looking into this!

That's a great question. The modifications needed for frei0r and vid.stab are not related to their source code at all, but rather to how their CMakeLists.txt files handle paths when they are consumed as a subproject (either via add_subdirectory or FetchContent).

The Core Issue: Path Resolution

The issue is a common one in older CMake projects: their scripts often use CMAKE_SOURCE_DIR to locate files. When MLT includes them as a subproject, CMAKE_SOURCE_DIR still points to MLT's top-level directory, not the dependency's own source directory. This causes their build scripts to fail when looking for their own files.

The fix is straightforward: patch their CMakeLists.txt to use CMAKE_CURRENT_SOURCE_DIR instead, which correctly points to their respective source trees. It's purely a build-system integration fix. While these patches could be submitted upstream, it might not be necessary given the overall strategy I'm proposing.

My Proposed Solution: A Two-Scenario Approach

My plan is to separate the build experience into two distinct user scenarios, which avoids complexity for most developers and provides a robust solution for those who need the optional modules.

For local/native builds (the main PR): The default CMakePresets will have optional modules like vid.stab and frei0r disabled by default. This provides the cleanest and fastest build experience for developers who just want to work on MLT's core. They won't need to deal with these dependencies at all.
For vcpkg users (the complete solution): The full power of MLT will be unlocked via vcpkg. I have developed a comprehensive vcpkg port for MLT that handles everything automatically. In this scenario:
The vcpkg port treats vid.stab and frei0r as proper dependencies.
It downloads their source, applies the necessary CMake path patches internally, and builds them correctly.
It then builds MLT with all the required module flags enabled.
Why this approach is better:

Clean & Simple for Everyone: Local developers get a minimal, fast build. vcpg users get a powerful, one-command installation (vcpkg install mlt[frei0r,vidstab]).
No Need for FetchContent or Submodules in MLT's core CMake: This keeps MLT's build system lean and free of dependency-fetching logic.
Encapsulates Complexity: All the "dirty work" (patching, dependency management) is perfectly encapsulated within the vcpkg port, where it belongs.
So, to answer your question directly: my current method is to provide these modules exclusively through a dedicated vcpkg port, where the necessary patches are handled cleanly. This seems like the most robust and user-friendly path forward.

What are your thoughts on this strategy?

eszlari: > The vcpkg port treats vid.stab and frei0r as proper dependencies.
> It downloads their source, applies the necessary CMake path patches internally, and builds them correctly.

You already submitted a PR for vid.stab: https://github.com/georgmartius/vid.stab/pull/135
Why not do it again and ask upstream to tag a new release?

Same for frei0r, just submit a PR: https://github.com/dyne/frei0r

jlskuz: > The fix is straightforward: patch their CMakeLists.txt to use CMAKE_CURRENT_SOURCE_DIR instead, which correctly points to their respective source trees. It's purely a build-system integration fix. While these patches could be submitted upstream, it might not be necessary given the overall strategy I'm proposing.

 And I agree with @eszlari it would be cleaner to have the patches upstream so that there are not need anywhere, neither in MLT nor in the vcpkg port. Also in the spirit of FOSS software it would be nice to get this contribute upstream so that other can benefit from it.

> No Need for FetchContent or Submodules in MLT's core CMake: This keeps MLT's build system lean and free of dependency-fetching logic.

I'd say this is generally a good idea.

> My Proposed Solution: A Two-Scenario Approach

I like to add that there is a third scenario: building MLT with MSVC but a different package manager than vcpkg (like KDE Craft which is used for Kdenlive). Your suggestion does not impact it in a bad way as far is I can see, I just want to make you aware if this scenario too.

ghost: Sure,i submit the pr first.

PR #1144, opened 2025-08-29
Fix crash using echo_pillar effect on image with alpha

j-b-m:
pillar_echo filter crashed if used with an rgba image after the recent rgba64 changes.
Reproducible with any image with alpha :
`melt image.png -attach pillar_echo`


bmatherly:
Thanks for your report and debugging. This is fixed here:
https://github.com/mltframework/mlt/commit/64037fb77ce73321e4e8da154400530c80a4c6a3

PR #1145, opened 2025-08-29
fix(build): Resolve build failures and add compatibility for MSVC

ghost:
This pull request introduces the foundational changes required to build the MLT framework using the Microsoft Visual C++ (MSVC) compiler on Windows. It adds a crucial POSIX compatibility layer and resolves several build-system issues that were blocking compilation on this platform.

### Key Changes:

*   **POSIX Compatibility Layer:**
    *   Introduces a new header (`msvc_posix_compat.h`) to provide implementations or wrappers for common POSIX functions and types not available natively on Windows (e.g., `gettimeofday`, parts of `unistd.h`).
    *   Numerous source files are updated with conditional includes (`#ifdef _MSC_VER`) to use this new compatibility layer, enabling cross-platform code to compile successfully.

*   **CMake Build System Fixes:**
    *   **mltladspa Module:** The `target_include_directories()` call for this target has been moved out of a conditional block. This resolves a critical build failure on MSVC where the conditions were never met, preventing the compiler from finding necessary headers.
    *   **Export Macro Generation:** The logic for creating export macro prefixes has been made more robust. It now uses a dedicated CMake command (`string(MAKE_C_IDENTIFIER)`) to ensure that all generated macro names are valid C identifiers, preventing potential errors with module names containing special characters.

### DIFF
--- a/src/framework/mlt.h
+++ b/src/framework/mlt.h
@@ -35,13 +35,13 @@
 extern "C" {
 #endif
 
-#include "mlt_export.h"
 #include "mlt_animation.h"
 #include "mlt_audio.h"
 #include "mlt_cache.h"
 #include "mlt_chain.h"
 #include "mlt_consumer.h"
 #include "mlt_deque.h"
+#include "mlt_export.h"
 #include "mlt_factory.h"
 #include "mlt_field.h"
 #include "mlt_filter.h"
--- a/src/framework/mlt_animation.h
+++ b/src/framework/mlt_animation.h
@@ -23,9 +23,9 @@
 #ifndef MLT_ANIMATION_H
 #define MLT_ANIMATION_H
 
+#include "mlt_export.h"
 #include "mlt_property.h"
 #include "mlt_types.h"
-#include "mlt_export.h"
 
 /** \brief Animation class
  *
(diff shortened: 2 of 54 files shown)


src/framework/msvc_posix_compat.h
```diff
@@ -0,0 +1,48 @@
...
+{
+    return _write(fd, buffer, count);
+}
+
+
+
+
+#endif // _MSC_VER
+
+#endif // MSVC_COMPAT_H
```
jlskuz: Reduce blank lines and remove duplicated `strcasecmp` and `strncasecmp` definitions

```suggestion
#ifndef MSVC_COMPAT_H
#define MSVC_COMPAT_H

#ifdef _MSC_VER 

#include <winsock2.h>
#include <ws2tcpip.h>

#include <windows.h>

#include <objbase.h>

#include <BaseTsd.h>
#include <io.h>
#include <string.h>
#include <malloc.h>

typedef SSIZE_T ssize_t;
#define STDIN_FILENO  _fileno(stdin)
#define STDOUT_FILENO _fileno(stdout)
#define STDERR_FILENO _fileno(stderr)
#define strcasecmp  _stricmp
#define strncasecmp _strnicmp
#define alloca _alloca 
#define tzname      _tzname
#define isatty       _isatty

static inline int close(int fd)
{
    return _close(fd);
}

static inline int write(int fd, const void *buffer, unsigned int count)
{
    return _write(fd, buffer, count);
}

#endif // _MSC_VER

#endif // MSVC_COMPAT_H
```

src/framework/CMakeLists.txt
```diff
@@ -88,6 +89,7 @@ target_link_libraries(mlt PRIVATE Threads::Threads ${CMAKE_DL_LIBS})
 
 if(MSVC)
   target_link_libraries(mlt PRIVATE msvccompat PThreads4W::PThreads4W)
+  target_link_libraries(mlt PUBLIC  ${GETTIMEOFDAY_LIBRARY})
```
jlskuz:
```suggestion
  target_link_libraries(mlt PUBLIC ${GETTIMEOFDAY_LIBRARY})
```

src/melt/melt.c
```diff
@@ -27,8 +27,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifndef _MSC_VER
 #include <unistd.h>
-
+#endif
+#ifdef _MSC_VER
+#include <framework/msvc_posix_compat.h>
+#endif
```
jlskuz:
```suggestion
#ifdef _MSC_VER
#include <framework/msvc_posix_compat.h>
#else
#include <unistd.h>
#endif
```

src/modules/gdk/producer_pixbuf.c
```diff
@@ -37,8 +37,12 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#ifndef _MSC_VER
 #include <unistd.h>
-
+#endif
+#ifdef _MSC_VER
+#include <framework/msvc_posix_compat.h>
+#endif
```
jlskuz:
```suggestion
#ifdef _MSC_VER
#include <framework/msvc_posix_compat.h>
#else
#include <unistd.h>
#endif
```

src/framework/msvc_posix_compat.h
jlskuz, line 1: I think this file should rather become part of `src/msvc` instead of `src/framework`

CMakeLists.txt
```diff
@@ -182,6 +182,13 @@ endif()
 
 if (MSVC)
   find_package(PThreads4W REQUIRED)
+  #find library from vcpkg gettimeofday
+    find_library(GETTIMEOFDAY_LIBRARY
+          NAMES gettimeofday
+  )
+    if (NOT GETTIMEOFDAY_LIBRARY)
+      message(FATAL_ERROR "Could not find gettimeofday.lib")
+    endif()
```
jlskuz, line 191: Given that this is vcpkg specific and we want to support other environments/package managers such as KDE Craft too, I'd prefer we include the files from https://github.com/microsoft/vcpkg/tree/7cfd63db7ff8855ab8e2baebeb80b37f5c0818db/ports/gettimeofday ourself. Since it consists of very few small files it seems okay to just  copy them `src/msvc` in my opinion.

jlskuz (changes requested):
Thanks for this follow up!

CMakeLists.txt
ghost: You are right, but I tried your method before and it always failed during the build, so I used this way instead.

jlskuz: How does it fail and why can't this be fixed?

ddennedy:
To remedy ongoing conflicts get `clang-format` version 14 and use it. There are build targets to use clang-format to apply and check (not enforced yet for pull requests, but I am considering it).

PR #1147, opened 2025-08-30
Fix recent void pointers arithmetic regression

jlskuz:
Pointer arithmetic with void pointers works in GNU C by considering the size of the void as 1 hence it worked.

However the C standard doesn't allow pointer arithmetic with void pointers.

Related to #1104 and #1131


bmatherly:
Fixed here: https://github.com/mltframework/mlt/commit/3882892dab0092ba5d8064050902e73fc597e486

Also added as an error to avoid this in the future: https://github.com/mltframework/mlt/commit/8af4342d9988ed6d7662ad543591990e8393c085

jlskuz:
Thanks!

PR #1151, opened 2025-09-03
Fix source documentation typos

luzpaz:
Fixed various typos in the source that would show up in documentation or user-facing context.


bmatherly:
Thanks!

PR #1155, opened 2025-09-18
Add mlt types for colorspace, trc, and primaries

bmatherly:
Formalize these values so their use can be more uniform in MLT as we add more color space conversion features.

### DIFF
--- a/src/framework/mlt_image.c
+++ b/src/framework/mlt_image.c
@@ -339,6 +339,72 @@ mlt_color_trc mlt_image_color_trc_id(const char *name)
     return mlt_color_trc_none;
 }
 
+const char *mlt_image_colorspace_name(mlt_colorspace colorspace)
+{
+    switch (colorspace) {
+    case mlt_colorspace_rgb:
+        return "rgb";
+    case mlt_colorspace_bt709:
+        return "bt709";
+    case mlt_colorspace_unspecified:
+        return "unspecified";
+    case mlt_colorspace_reserved:
+        return "reserved";
+    case mlt_colorspace_fcc:
+        return "fcc";
+    case mlt_colorspace_bt470bg:
+        return "bt470bg";
+    case mlt_colorspace_smpte170m:
+        return "smpte170m";
+    case mlt_colorspace_smpte240m:
+        return "smpte240m";
+    case mlt_colorspace_ycgco:
+        return "ycgco";
+    case mlt_colorspace_bt2020_ncl:
+        return "bt2020nc";
+    case mlt_colorspace_bt2020_cl:
+        return "bt2020c";
+    case mlt_colorspace_smpte2085:
+        return "smpte2085";
+    case mlt_colorspace_bt601:
+        return "bt601";
+    case mlt_colorspace_invalid:
+        return "invalid";
+    }
+    return "invalid";
+}
+
+mlt_colorspace mlt_image_colorspace_id(const char *name)
+{
+    const mlt_colorspace colorspaces[] = {mlt_colorspace_rgb,
+                                          mlt_colorspace_bt709,
+                                          mlt_colorspace_unspecified,
+                                          mlt_colorspace_reserved,
+                                          mlt_colorspace_fcc,
+                                          mlt_colorspace_bt470bg,
+                                          mlt_colorspace_smpte170m,
+                                          mlt_colorspace_smpte240m,
+                                          mlt_colorspace_ycgco,
+                                          mlt_colorspace_bt2020_ncl,
+                                          mlt_colorspace_bt2020_cl,
+                                          mlt_colorspace_smpte2085,
+                                          mlt_colorspace_bt601,
+                                          mlt_colorspace_invalid};
+
+    // Fall back to see if it was specified as a number
+    int value = strtol(name, NULL, 10);
+    if (!value && strcmp(name, "0"))
+        value = -1; // strtol returned an error;
+
+    for (int i = 0; name && i < sizeof(colorspaces); i++) {
+        const char *s = mlt_image_colorspace_name(colorspaces[i]);
+        if (value == colorspaces[i] || !strcmp(s, name))
+            return colorspaces[i];
+    }
+
+    return mlt_colorspace_invalid;
+}
+
 /** Fill an image with black.
   *
   * \bug This does not respect full range YUV if needed.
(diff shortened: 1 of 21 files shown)


src/framework/mlt_image.c
```diff
@@ -262,6 +262,206 @@ mlt_image_format mlt_image_format_id(const char *name)
...
+        return value;
+
+    return mlt_color_trc_none;
+}
+
+const char *mlt_image_colorspace_name(mlt_colorspace colorspace)
+{
+    switch (colorspace) {
+    case mlt_colorspace_rgb:
+        return "rgb";
```
bmatherly, line 346: FFMPEG uses "gbr" for this name. But I used "rgb" here.

src/framework/mlt_types.h
```diff
@@ -97,21 +97,63 @@ typedef enum {
 /** Colorspace definitions */
 
 typedef enum {
-    mlt_colorspace_rgb = 0,   ///< order of coefficients is actually GBR, also IEC 61966-2-1 (sRGB)
-    mlt_colorspace_bt709 = 1, ///< also ITU-R BT1361 / IEC 61966-2-4 xvYCC709 / SMPTE RP177 Annex B
+    mlt_colorspace_rgb = 0, ///< order of coefficients is actually GBR, also IEC 61966-2-1 (sRGB)
```
bmatherly, line 100: This enum was never used in the past. It only existed as a member in mlt_image. But we never used it. So I feel ok about changing the values. The values were previously set to match FFMpeg. I changed some values to match the hard-coded values that were previously in MLT. The values that were not previously used are unchanged from the FFMpeg values.

src/framework/mlt_types.h
```diff
@@ -97,21 +97,63 @@ typedef enum {
...
-    mlt_colorspace_bt2020_cl = 10, ///< ITU-R BT2020 constant luminance system
-    mlt_colorspace_smpte2085 = 11, ///< SMPTE 2085, Y'D'zD'x
+    = 470, ///< also ITU-R BT601-6 625 / ITU-R BT1358 625 / ITU-R BT1700 625 PAL & SECAM / IEC 61966-2-4 xvYCC601
+    mlt_colorspace_smpte170m = 170, ///< also ITU-R BT601-6 525 / ITU-R BT1358 525 / ITU-R BT1700 NTSC
+    mlt_colorspace_smpte240m = 240,   ///< functionally identical to above
+    mlt_colorspace_ycgco = 8,         ///< Used by Dirac / VC-2 and H.264 FRext, see ITU-T SG16
+    mlt_colorspace_bt2020_ncl = 2020, ///< ITU-R BT2020 non-constant luminance system
+    mlt_colorspace_bt2020_cl = 2021,  ///< ITU-R BT2020 constant luminance system
+    mlt_colorspace_smpte2085 = 11,    ///< SMPTE 2085, Y'D'zD'x
+    mlt_colorspace_bt601 = 601,       ///< BT.470 (625/PAL) or SMPTE170M (525/NTSC)
```
bmatherly, line 113: 601 is added as new to this structure - even though it is technically a superset of bt470gb and smpte170m. This allowed me to directly drop in the enum values in places were hard coded 601, 470 and 170 were used previously.

src/framework/mlt_types.h
```diff
@@ -97,21 +97,63 @@ typedef enum {
...
+    mlt_colorspace_invalid
 } mlt_colorspace;
 
+typedef enum {
```
bmatherly, line 117: These all match FFMpeg

src/framework/mlt_types.h
```diff
@@ -97,21 +97,63 @@ typedef enum {
...
+    mlt_color_trc_bt2020_10 = 14,    ///< ITU-R BT2020 for 10-bit system
+    mlt_color_trc_bt2020_12 = 15,    ///< ITU-R BT2020 for 12-bit system
+    mlt_color_trc_smpte2084 = 16,    ///< SMPTE ST 2084 for 10-, 12-, 14- and 16-bit systems
+    mlt_color_trc_smpte428 = 17,     ///< SMPTE ST 428-1
+    mlt_color_trc_arib_std_b67 = 18, ///< ARIB STD-B67, known as "Hybrid log-gamma"
+    mlt_color_trc_invalid
+} mlt_color_trc;
+
+typedef enum {
+    mlt_color_pri_none = 0,
```
bmatherly, line 142: This is a mix of using the previously hard coded values from MLT, and FFMpeg values for items that were not used in MLT.

src/modules/avformat/consumer_avformat.c
```diff
@@ -354,63 +356,12 @@ static void color_trc_from_colorspace(mlt_properties properties)
 static void color_primaries_from_colorspace(mlt_properties properties)
 {
     // Default color_primaries from MLT colorspace.
-    switch (mlt_properties_get_int(properties, "colorspace")) {
-    case 0: // sRGB
-    case 709:
-        mlt_properties_set_int(properties, "color_primaries", AVCOL_PRI_BT709);
```
bmatherly, line 360: This is one spot I am unsure of. Why were we setting the properties with FFMpeg values? Everywhere else we were setting the MLT values (709, 601525, 601625). Was this a bug? Or is there a reason we were setting "color_primaries" to mlt values in some places and FFMpeg values in this one place?

ddennedy: "color_primaries" gets put into an AVOption that encoders and muxers use to write important metadata. This cannot be changed to MLT internal values or must be converted from MLT's values to FFmpeg's.

FYI, the purpose of this function is to default them from the colorspace since there is a strong default correlation. However, one can override it by explicitly. In `consumer_start()`, there is:
```c
          if (!mlt_properties_get(properties, "color_primaries"))
            color_primaries_from_colorspace(properties);
```

src/framework/mlt_types.h
ddennedy: As far as I can recall, the original MLT values were put into avformat producer for the Movit integration long ago. Later, it looks like this frame property was integrated into the avfilter incorrectly: avfilters are going to expect FFmpeg values. I see you fixed that in this PR. This enum can be changed to use FFmpeg's values if you want to since they were internal only. "color_primaries" on the avformat consumer did not use the same values as mlt_frame but rather FFmpeg's. However, that should be changed to be consistent (as long as the consumer gets FFmpeg's values as it will be copied to an AVOption).

bmatherly:
Dan, Thanks for the history and context. Based on your comments. I added a commit to change the primaries values to match the FFMPeg values. However, I excluded the FFMpeg reserved values. I think the conversion function is still useful because it allows us to "sanitize" unexpected input values.

PR #1158, opened 2025-09-21
Fix crash on empty color primaries name

j-b-m:
The recent commit introducing color primaries (https://github.com/mltframework/mlt/commit/433f9b5c1c04a1684579e371b8041f3796e7a28e) causes crashes in Kdenlive when trying to open some projects.
Turns out in some cases, it tries to use an empty name for the primaries, so a check for NULL fixes the problem.


bmatherly:
Classic rookie mistake on my part. Thank for the fix, @j-b-m

PR #1159, opened 2025-09-29
Add color_transform normalize filter

bmatherly:
This filter can force the color TRC to whatever the consumer requests.

Posting this as a draft for conceptual review. The PR should not be pushed until it is tested with Shotcut.

This commit adds a new normalize filter that is added once by the loader on each clip, and then again by the consumer at the end of all the services. The first instance will set the TRC to  be whatever is set by consumer.mlt_color_trc. The second will change the TRC to consumer.color_trc.

Example usage:
`melt test.mp4 -consumer sdl2 mlt_color_trc=linear mlt_image_format=rgba64`

### DIFF
--- a/src/modules/core/filter_color_transform.yml
+++ b/src/modules/core/filter_color_transform.yml
@@ -1,6 +1,6 @@
 schema_version: 7.0
 type: filter
-identifier: colorspace
+identifier: color_transform
 title: Colorspace & Transfer Characteristics
 version: 1
 copyright: Meltytech, LLC
--- a/src/framework/mlt_types.c
+++ b/src/framework/mlt_types.c
@@ -67,7 +67,7 @@ mlt_deinterlacer mlt_deinterlacer_id(const char *name)
     return mlt_deinterlacer_invalid;
 }
 
-uint8_t srgb_to_linear(uint8_t value)
+static uint8_t srgb_to_linear(uint8_t value)
 {
     // Normalize the 8-bit value to the [0.0, 1.0] range
     double factor = (double) value / 255.0;
@@ -81,7 +81,7 @@ uint8_t srgb_to_linear(uint8_t value)
     return (uint8_t) CLAMP(linear_value, 0.0, 255.0);
 }
 
-/** Convert a standard sRGB color value to the specified colorspace.
+/** Convert a standard sRGB color value to the specified TRC (Transfer Characteristics).
  *
  * \param color a standard sRGB color value
  * \param trc_name a string representing the TRC to convert to


src/modules/core/filter_color_transform.c
```diff
@@ -0,0 +1,287 @@
...
+        // TRC matches. No conversion required
+        return 0;
+    }
+    if (out_trc == mlt_color_trc_none || frame_trc == mlt_color_trc_none) {
+        mlt_log_info(MLT_FILTER_SERVICE(filter),
+                     "Missing TRC - Frame: %s\tOut: %s\n",
+                     mlt_image_color_trc_name(frame_trc),
+                     mlt_image_color_trc_name(out_trc));
+        return 0;
+    }
```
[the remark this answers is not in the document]
bmatherly: This is not required because "ensure_color_properties" will make sure a valid value is returned.

src/modules/core/filter_brightness.c
```diff
@@ -40,22 +40,56 @@ static int sliced_proc(int id, int index, int jobs, void *cookie)
...
+            int32_t n = 128 * ((1 << 16) - m);
+            int min = ctx->full_range ? 0 : 16;
+            int max_luma = ctx->full_range ? 255 : 235;
+            int max_chroma = ctx->full_range ? 255 : 240;
+            for (int line = 0; line < slice_height; line++) {
+                uint8_t *p = ctx->image->planes[0]
+                             + ((slice_line_start + line) * ctx->image->strides[0]);
+                for (int pixel = 0; pixel < ctx->image->width; pixel++) {
+                    p[0] = CLAMP((*p * m) >> 16, min, max_luma);
+                    p[0] = CLAMP((*p * m + n) >> 16, min, max_chroma);
```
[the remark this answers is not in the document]
bmatherly: Fixed

src/framework/mlt_image.c
```diff
@@ -472,6 +472,123 @@ mlt_color_primaries mlt_image_color_pri_id(const char *name)
...
+    case mlt_image_yuv444p10:
+        if (height < 576) {
+            // Assume NTSC
+            colorspace = mlt_colorspace_smpte170m;
+        } else if (height < 720) {
+            // Assume PAL
+            colorspace = mlt_colorspace_bt470bg;
+        } else {
+            // Assume HDTV
+            colorspace = mlt_color_trc_bt709;
```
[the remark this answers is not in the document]
bmatherly: Fixed

src/framework/chain_normalizers.ini
```diff
@@ -14,6 +14,7 @@ rescaler=movit.resample,swscale,gtkrescale,rescale
 resizer=movit.resize,resize
 movitconvert=movit.convert
 imageconvert=avcolor_space,imageconvert
+colorspace=colorspace
```
ddennedy: Anything that involves pixel blending or interpolation should be done in linear color if that is requested. Is it possible to move to first in the list since `deinterlace` and `rescaler` does that? I have not looked into it yet, but this `colorspace` conversion should be skipped when movit is being used because it already does that and usually faster (GPU). Don't worry about the deinterlace case with respect to movit; it has that but I need to integrate it.

src/modules/core/filter_colorspace.yml
```diff
@@ -0,0 +1,47 @@
...
+      - auto # Choose best available
+      - avfilter.zscale
+      - avfilter.colorspace
+    description: |
+      The underlying conversion method to use. The method is based on the
+      encapsulated filter that is used. 'auto' will choose the best available
+      in this order:
+        * avfilter.zscale
+        * avfilter.colorspace
+    required: yes
```
ddennedy: "required" means the user must specify it, not simply that the filter needs some value.  For example, something that requires a file name to be specified would set the respective property to be required. But here, it is not since there is a default.

src/modules/core/filter_color_transform.yml
```diff
@@ -0,0 +1,47 @@
...
+      encapsulated filter that is used. 'auto' will choose the best available
+      in this order:
+        * avfilter.zscale
+        * avfilter.colorspace
+    required: yes
+    readonly: no
+    mutable: no
+    default: auto
+  - identifier: force_trc
+    title: Force TRC
```
ddennedy, line 39: Since I asked for a change in this file, might as well also at least once write "transfer characteristics" instead of only "TRC" everywhere. `title` is a good place since identifiers often use abbreviations.

src/modules/core/filter_colorspace.c
```diff
@@ -0,0 +1,360 @@
...
+}
+
+mlt_filter filter_colorspace_init(mlt_profile profile,
+                                  mlt_service_type type,
+                                  const char *id,
+                                  char *arg)
+{
+    const char *method = arg;
+    if (!method)
+        method = "auto";
```
ddennedy:
```suggestion
    const char *method = arg ? arg : "auto";
```

src/modules/core/filter_colorspace.c
```diff
@@ -0,0 +1,360 @@
...
+                            int *width,
+                            int *height,
+                            int writable)
+{
+    mlt_filter self = (mlt_filter) mlt_frame_pop_service(frame);
+    mlt_properties frame_properties = MLT_FRAME_PROPERTIES(frame);
+    mlt_properties filter_properties = MLT_FILTER_PROPERTIES(self);
+    mlt_image_format requested_format = *format;
+    int ret = mlt_frame_get_image(frame, image, format, width, height, writable);
+    if (ret || requested_format == mlt_image_movit)
```
ddennedy: I think this is where you are attempting to bypass this filter when movit is being used. `filter_movit_crop.cpp` takes care of interfacing with the producer since it a normalizer early in the chain. It requests image format `mlt_image_none`, which is supposed to tell the producer the MLT format closest to the source. If we move `colorspace` to the beginning of the normalizers, this will stop getting `mlt_image_movit` and `mlt_image_none` instead. You can test that format here as well.

src/framework/mlt_consumer.c
```diff
@@ -682,12 +683,43 @@ mlt_frame mlt_consumer_get_frame(mlt_consumer self)
...
         mlt_properties_set(frame_properties,
                            "consumer.channel_layout",
                            mlt_properties_get(properties, "channel_layout"));
         mlt_properties_set(frame_properties,
                            "consumer.color_range",
                            mlt_properties_get(properties, "color_range"));
+
+        if (mlt_properties_get(properties, "mlt_color_trc")) {
+            // Add a normalize filter to convert the mlt_color_trc to color_trc
+            mlt_cache_item cache_item = mlt_service_cache_get(service, "cs_filter");
```
ddennedy: What is the purpose of the cache here? mlt_cache is generally used when there are many of the same class of object, which is common with producers. However, it is uncommon to have more than one consumer and extremely uncommon to have more than 2.

ddennedy (changes requested):

src/framework/chain_normalizers.ini
bmatherly: Done. I am not sure the best place for it. It is easy to move later if we want. Also, I aspire to make a version of the filter that is based directly on zimg. If I do that, the filter could also do scaling and format conversion. So maybe that will change the order.

src/modules/core/filter_colorspace.yml
bmatherly: done

src/modules/core/filter_color_transform.yml
bmatherly: Done.

But this comment triggered me to call into question the name of the entire filter. The name of the filter is "Colorspace". But all it really does is change the TRC. Do we aspire for this filter to eventually also change the colorspace and maybe primaries? 

The colorspace name also conflicts with av_color_space and a possible future "z_color_space" filter. Maybe the name should be "filter_trc". Or maybe it should be "filter_work" to indicate that it only changes the image format as requested by the consumer. I am open minded about it.

src/modules/core/filter_colorspace.c
bmatherly: done

src/modules/core/filter_colorspace.c
bmatherly: done

src/framework/mlt_consumer.c
bmatherly: Good comment. I changed it to a data property with a private name (underscore).

src/modules/core/filter_color_transform.yml
ddennedy: That thought crossed my mind briefly as well. I suggest `color_transform`. That differentiates it and keeps it open to do more, but if you think it might also entail scaling maybe `conform_image`? I had considered `normalize_image`, but there are 2 existing video normalize filters from avfilter and frei0r that are about stretching the levels.

bmatherly: I like the "color_transform" idea. I changed it to that. Thanks for brainstorming with me.

Regarding the zimg library, I do not think that there is an advantage to create a filter that combines colorspace with scaling. It appears that the library does those in sequence anyway (not some optimized combined algorithm).

I think that the future zimg based filters will be:
* z_color_space - same functionality as av_color_space (but maybe faster)
* z_scale - same functionality as swsale
* z_color_transform - same functionality as the ffmpeg colorspace and zimg filters (but without the extra memcpy in/out)

These three filters could become options for the normalizing filters (any maybe the preferences for performance or for colorspace compliance).

ddennedy:
I've been doing a lot of testing of the more advanced scenarios involving movit, HLG and decklink plus a little bit of the others, which are working fine.

- [x] document `mlt_color_trc` in `consumer.h`
- [x] find more places where `get_color` and `anim_get_color` are called (C and C++) to handle frame color_trc
- [ ] ~~Update ten_bit presets~~

Update: strike changing the ten_bit presets

ddennedy:
I can do the other calls to `get_color()` if you have not started and want to work on `z_color_transform`

bmatherly:
> I can do the other calls to get_color() if you have not started and want to work on z_color_transform

That is fine with me. Although, I thought I had converted all the filters - it would be good to fix any that I missed. 

I am undecided about whether to convert producers (like the color producer) or just let the new color_transform filter make the conversion. There are multiple layers of logic to determine what TRC should be used. And if we copy/paste that all around the code, it might be frustrating to change later. I'd trust your judgement.

ddennedy:
I think this is ready now. Please review my changes and we can merge.

bmatherly:
Good to go. Thanks for your help.

PR #1161, opened 2025-10-19
Add static code analysis workflow with Cppcheck

jlskuz:
This adds a GitHub Workflow to run [Cppcheck](http://cppcheck.net/) static code analysis. Despite the name Cppcheck also supports not only C++ but also C.

The results are not only visible in the log but also exported as [GitHub Annotations](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message).

Yet the job fails because Cppcheck finds some issues. These have to be reviewed now and either fixed or set to be ignored (in case of false positives).

### DIFF
--- a/.github/workflows/static-code-analysis.yml
+++ b/.github/workflows/static-code-analysis.yml
@@ -13,10 +13,13 @@ jobs:
         -i src/modules/decklink/win
         -i src/modules/glaxnimate/glaxnimate/
         -i src/modules/plus/ebur128/
+        -i src/modules/xml/common.c
+        -i src/win32/strptime.c
         --include=src/framework/mlt_log.h
         --include=src/framework/mlt_types.h
         --library=cppcheck.cfg
         --suppress=ctuOneDefinitionRuleViolation
+        --suppress=syntaxError:src/modules/xml/common.c
     steps:
       - uses: actions/checkout@v4
       - name: Install Cppcheck
--- a/makefile
+++ b/makefile
@@ -24,7 +24,10 @@ cppcheck:
 		-i src/modules/decklink/win \
 		-i src/modules/glaxnimate/glaxnimate/ \
 		-i src/modules/plus/ebur128/ \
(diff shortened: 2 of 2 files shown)


bmatherly:
Cool! Thanks for working on this. I see you marked this as draft. Do you plan to resolve more of the findings? If so, let me know if you need any help.

jlskuz:
Sorry for the late reply, I caught a cold and hence was offline. Thanks for your help @ddennedy and the offer @bmatherly !

bmatherly:
image

I was hoping there would be some time for discussion before this was merged. But I guess there is still time.
I'm excited about the bugs that can be found with this tool. Here are some comments:

* Should we put the command line arguments in a .cppcheck-config file so they do not have to be duplicated in two places?
* Do we want a cmake target?
* Is there any way to avoid the in-code cppcheck-suppress comments? I really don't like those. If we want to keep them, should we add an explanation for why the supression is being allowed?

ddennedy:
You can move some things to the config file. See https://cppcheck.sourceforge.io/reference-cfg-format.pdf
You can make a PR with your ideas. The command lines are a little different between the two because `--template` used for the GitHub annotations breaks my IDEs' ability to click lines in the output. 
Why do you think cmake is more appropriate for this than `makefile`?
Not all suppressions need to be inline, but I prefer that for line-oriented since the separated method breaks too easily as the line numbers change.

ddennedy:
P.S. I am going to make a release in a couple of days, which is one reason I moved this along.

bmatherly:
> Why do you think cmake is more appropriate for this than makefile?

I'm not saying it is better. Just more "complete". I think we want cmake to  be a first-class-citizen. And I know many projects have moved to using it. If someone is using cmake, they will not "discover" the cppcheck analysis - which has now become a requirement for code submission.

> You can make a PR with your ideas. 

Will do. Thanks.

PR #1171, opened 2025-11-12
Fix multiple memory safety issues detected by ASan/UBSan/LeakSan

ShadowRider09:
- Fix null pointer dereference in filter_avcolour_space.c when profile is NULL
- Fix Pango/Fontconfig memory leak in producer_pango.c by adding instance counting
- Fix null pointer dereferences in producer_colour.c (two locations)
- Fix null pointer dereference in mlt_producer_set_in_and_out
- Fix null pointer dereference in filter_movit_convert.cpp
- Fix double-free in producer_xml.c by properly handling stack state All fixes have been verified with fuzzing tests. See FIXES_REPORT.md for details, and all of the crash is in the zip
[fixed_crashes.zip](https://github.com/user-attachments/files/23497459/fixed_crashes.zip)

### DIFF
--- a/src/modules/core/producer_colour.c
+++ b/src/modules/core/producer_colour.c
@@ -95,10 +95,21 @@ static int producer_get_image(mlt_frame frame,
     // Choose suitable out values if nothing specific requested
     if (*format == mlt_image_none || *format == mlt_image_movit)
         *format = mlt_image_rgba;
-    if (*width <= 0)
-        *width = mlt_service_profile(MLT_PRODUCER_SERVICE(producer))->width;
-    if (*height <= 0)
-        *height = mlt_service_profile(MLT_PRODUCER_SERVICE(producer))->height;
+    if (*width <= 0 || *height <= 0) {
+        mlt_profile profile = mlt_service_profile(MLT_PRODUCER_SERVICE(producer));
+        if (profile) {
+            if (*width <= 0)
+                *width = profile->width;
+            if (*height <= 0)
+                *height = profile->height;
+        } else {
+            if (*width <= 0)
+                *width = 16;
+            if (*height <= 0)
+                *height = 16;
+        }
+    }
+
 
     // Choose default image format if specific request is unsupported
     if (*format != mlt_image_yuv420p && *format != mlt_image_yuv422 && *format != mlt_image_rgb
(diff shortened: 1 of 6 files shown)


src/modules/core/producer_colour.c
```diff
@@ -95,10 +95,21 @@ static int producer_get_image(mlt_frame frame,
...
+            if (*width <= 0)
+                *width = profile->width;
+            if (*height <= 0)
+                *height = profile->height;
+        } else {
+            if (*width <= 0)
+                *width = 16;
+            if (*height <= 0)
+                *height = 16;
+        }
```
ddennedy, line 110:
```suggestion
        if (*width <= 0)
            *width = profile ? profile->width : 16;
        if (*height <= 0)
            *height = profile ? profile->height : 16;
```

src/modules/core/producer_colour.c
```diff
@@ -259,10 +270,16 @@ static int producer_get_frame(mlt_producer producer, mlt_frame_ptr frame, int in
         // Set producer-specific frame properties
         mlt_properties_set_int(properties, "progressive", 1);
         mlt_profile profile = mlt_service_profile(MLT_PRODUCER_SERVICE(producer));
-        mlt_properties_set_double(properties, "aspect_ratio", mlt_profile_sar(profile));
-        mlt_properties_set_int(properties, "meta.media.width", profile->width);
-        mlt_properties_set_int(properties, "meta.media.height", profile->height);
+	double sar = profile ? mlt_profile_sar(profile) : 1.0;
```
ddennedy, line 273:
```suggestion
	    double sar = profile ? mlt_profile_sar(profile) : 1.0;
```

src/modules/gdk/producer_pango.c
```diff
@@ -198,6 +199,7 @@ mlt_producer producer_pango_init(const char *filename)
         pthread_mutex_lock(&pango_mutex);
         if (fontmap == NULL)
             fontmap = (PangoFT2FontMap *) pango_ft2_font_map_new();
+	pango_instance_count++;
```
ddennedy, line 202:
```suggestion
	     pango_instance_count++;
```

Please use clang-format v14 to prevent formatting issues. We have a build target for that.

ddennedy (changes requested):
Thank you for your contribution! I only request some minor easy things to change.

bmatherly:
Some of these checks are good. But the profile checks add a lot of noise to the code. If a service was not initialized with a profile, then the developer is not using the API correctly and we should let them know right away - not mask it and make it more difficult to discover later. How would you feel about adding assert checks for profiles before using them? That would add less conditional checks to the code and give the developer the earliest warning that something is wrong.

jlskuz:
@ShadowRider09 could you provide the commands you used to discover these issues? We might consider to put them to CI to prevent regressions...

ShadowRider09:
> @ShadowRider09 could you provide the commands you used to discover these issues? We might consider to put them to CI to prevent regressions...
 I use the Asan+Fussing to find the safty issues, and there is the source code of fuzzer:
```c
#include <stdint.h>
#include <stddef.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>
#include <fcntl.h>

#include "mlt.h"

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    // Ignore inputs that are too small or too large
    if (size < 10 || size > 1024 * 1024) {
        return 0;
    }

    // Initialize MLT (only on the first call)
    static int initialized = 0;
    if (!initialized) {
        // Set the MLT repository path to our compiled version
        setenv("MLT_REPOSITORY", "/home/han/mlt/build-asan/out/lib/mlt", 1);
        mlt_factory_init(NULL);
        initialized = 1;
    }

    // Create a temporary file to store fuzzing input
    char tmpfile[] = "/tmp/fuzz_mlt_XXXXXX";
    int fd = mkstemp(tmpfile);
    if (fd == -1) {
        return 0;
    }

    // write data
    if (write(fd, data, size) != (ssize_t)size) {
        close(fd);
        unlink(tmpfile);
        return 0;
    }
    close(fd);

    // Attempt to load an XML file as a producer
    mlt_producer producer = mlt_factory_producer(NULL, "xml", tmpfile);

    if (producer != NULL) {
        // Attempt to obtain some frames to trigger more code
        mlt_frame frame = NULL;
        mlt_service_get_frame(MLT_PRODUCER_SERVICE(producer), &frame, 0);

        if (frame != NULL) {
            mlt_frame_close(frame);
        }

        // clean
        mlt_producer_close(producer);
    }

    // delete temp files
    unlink(tmpfile);

    return 0;
}
```

PR #1176, opened 2025-11-27
makefile: update codespell ignore word list

luzpaz:
Updated codespell dictionary to include additional terms (to avoid false positives) + alphabetized them


luzpaz:
soft bump

PR #1177, opened 2025-11-28
Support rgba64 in lift_gamma_gain

ddennedy:


src/modules/plus/filter_lift_gamma_gain.c
```diff
@@ -60,12 +66,49 @@ static void refresh_lut(mlt_filter filter, mlt_frame frame)
...
+        self->blut = malloc(256 * sizeof(uint8_t));
+
+        if (!self->rlut || !self->glut || !self->blut) {
+            mlt_log_error(MLT_FILTER_SERVICE(filter), "Failed to allocate 8-bit LUTs\n");
+            free(self->rlut);
+            free(self->glut);
+            free(self->blut);
+            self->rlut = NULL;
+            self->glut = NULL;
+            self->blut = NULL;
```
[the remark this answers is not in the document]
ddennedy: wrong

src/modules/plus/filter_lift_gamma_gain.c
```diff
@@ -194,23 +301,50 @@ static int filter_get_image(mlt_frame frame,
...
+        error = refresh_lut(filter, frame, *format);
+        if (!error) {
+            private_data *self = (private_data *) filter->child;
+            sliced_desc desc;
+            desc.filter = filter;
+            desc.image = *image;
+            desc.format = *format;
+            desc.width = *width;
+            desc.height = *height;
+            desc.rlut = self->rlut;
```
bmatherly, line 335: I do not think it is sufficient to protect the pointer copy with the mutex. A thread could free the pointers while another thread is still using the pointer copies. The allocated memory should be protected whenever it is in use.

I recommend an rwlock so that all threads reading the lut will be completed before it can be changed.

https://pubs.opengroup.org/onlinepubs/007904975/functions/pthread_rwlock_destroy.html

ddennedy: Another option is to extend the service lock to cover `mlt_slices_run_normal()`. This filter probably does not need frame threading, and that is simpler. What do you think?

bmatherly: Your change to move the lock resolves my concern about unprotected data. I do not have a preference between frame/slice threading. maybe the slice threading is better because it will reduce the latency to deliver a frame in real-time applications

bmatherly (approved):

PR #1179, opened 2025-12-03
Fix #1178 background image scaling in gpsgraphic filter

### LINKED ISSUE #1178: GPS Graphic filter: wrong scaling of background image
**Describe the bug**
GPS Graphic uses a wrong scaling for the background image. The scale is calculated by the distance of the GPS points. But the maps build from tiles usually don't use the gps coordinate system ESPG4326 / WG84, but the transformation EPSG3857. For the x-axis this doesn't matter, but y-axis is scaled differently.
In the current implementation this leads to the gps track being correctly positioned in x-axis, but incorrect in y-axis if the background image is a map from for example Open Street Map.

**To Reproduce**
Take a map from Open Street Map and try to match it to the gps track. It will be off on the upper and lower end if it matches in width.

**Expected behavior**
Perfect match of gps track and the map

**Desktop (please complete the following information):**
- MLT/melt version: 7.34.1 and still in git

**Additional context**
I implemented a transformation for the latitude, which compensates the different projection of EPSG3857. At least for me the map is now a perfect match. Projection should be done only once when loading / interpolating the file, so it shouldn't affect rendering speed. Hope I hit all relevant lines and didn't mess up anything.

[gpsgraphic.patch](https://github.com/user-attachments/files/23916115/gpsgraphic.patch)

gnampf1:
tranform latitude for gpsgraphic, as the tile server usually use a different coordinate represenation EPSG 3857, which leads to misscaling without transformation, especially on large areas. Transformation is done during loading of the file and interpolation, so hopefully not affecting speed of filter

### DIFF
--- a/src/modules/qt/filter_gpsgraphic.cpp
+++ b/src/modules/qt/filter_gpsgraphic.cpp
@@ -130,12 +130,12 @@ double get_by_src(mlt_filter filter,
     if (pdata->graph_data_source == gspg_location_src) {
         if (get_type == -1) {
             if (subtype == gpsg_latitude_id)
-                return pdata->minmax.min_lat;
+                return pdata->minmax.min_lat_projected;
             else if (subtype == gpsg_longitude_id)
                 return pdata->minmax.min_lon;
         } else if (get_type == 1) {
             if (subtype == gpsg_latitude_id)
-                return pdata->minmax.max_lat;
+                return pdata->minmax.max_lat_projected;
             else if (subtype == gpsg_longitude_id)
                 return pdata->minmax.max_lon;
         } else if (get_type == 0) {
@@ -472,8 +472,8 @@ static void find_minmax_of_data(mlt_filter filter)
         m = x;
     for (int i = 0; i < pdata->gps_points_size; i++) {
         gps_point_proc *crt = &pdata->gps_points_p[i];
-        assign_if_smaller(crt->lat_projected, pdata->minmax.min_lat);
-        assign_if_bigger(crt->lat_projected, pdata->minmax.max_lat);
+        assign_if_smaller(crt->lat, pdata->minmax.min_lat);
+        assign_if_bigger(crt->lat, pdata->minmax.max_lat);
         assign_if_smaller(crt->lon, pdata->minmax.min_lon);
         assign_if_bigger(crt->lon, pdata->minmax.max_lon);
         assign_if_smaller(crt->ele, pdata->minmax.min_ele);
@@ -485,13 +485,15 @@ static void find_minmax_of_data(mlt_filter filter)
         assign_if_smaller(crt->grade_p, pdata->minmax.min_grade_p);
         assign_if_bigger(crt->grade_p, pdata->minmax.max_grade_p);
     }
+    pdata->minmax.min_lat_projected = project_latitude(pdata->minmax.min_lat);
+    pdata->minmax.max_lat_projected = project_latitude(pdata->minmax.max_lat);
 #undef assign_if_smaller
 #undef assign_if_bigger
 
-    // //compute the gps track aspect ratio (from coords) // the other one seems better almost every time
+    //compute the gps track aspect ratio (from coords)
     double map_aspect_ratio = 1;
     double map_width = pdata->minmax.max_lon - pdata->minmax.min_lon;
-    double map_height = pdata->minmax.max_lat - pdata->minmax.min_lat;
+    double map_height = pdata->minmax.max_lat_projected - pdata->minmax.min_lat_projected;
     if (map_width && map_height) {
         map_aspect_ratio = map_width / map_height;
         pdata->map_aspect_ratio_from_distance = map_aspect_ratio;
--- a/src/modules/qt/gps_parser.cpp
+++ b/src/modules/qt/gps_parser.cpp
@@ -34,6 +34,11 @@ double project_latitude(double lat)
     return log(tan(((90 + (double)lat) * MATH_PI) / 360)) / (MATH_PI / 180);
 }
 
+double unproject_latitude(double lat_projected)
+{
+    return (atan(pow(exp(1), lat_projected * (MATH_PI / 180)))*360) / MATH_PI - 90;
+}
+
 //shifts all (longitude) values from near 180 to 0
 double get_180_swapped(double lon)
 {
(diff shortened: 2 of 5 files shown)


dany123:
@ddennedy is there a .zip created with the built app from this PR to do some hands-on tests? I'm curious how big the difference is between the 2 systems as I didn't notice any major shifts in my personal tracks.

I had a first look at the changes and they seem to cover well most of the cases I can think of. I'll have another look the next few days as I'm short on time today.

There's 3 places where I think we need to print the original unprojected lat point:
- the "now text" option which prints the lat,lon coordinates on the bottom right corner of the map
- the grid ("Draw legend" option)
- the middle_point I think it's also being computed from the projected latitude now (correct me if I'm wrong)

The get_src change makes getting the original lat point rather messy, so maybe it's faster to just make a function to compute the reverse of project_latitude? 

For the find_minmax_of_data() change, I think it would be better to just do the conversion at the very end, not inside the for, to avoid unnecessary computations. This would also quickly fix the 3rd bullet point by moving the middle_point computation before the conversion. (I should have added an exact line comment for this one but it's 1AM and need to go).

gnampf1:
The error depends on the distance I'ld say, as it's not a linear scaling. Thats why it doesn't help to simply scale the image itself.
In my case its more than 100km. I've attached some samples, old is without my patch, new is with my patch, and then also the base image and the track (Day1.xml, rename to gpx) I used
new
old
Test1
[Day1.xml](https://github.com/user-attachments/files/23938815/Day1.xml)

I guess you might be right with min/max, maybe the projection should only be done there when the aspect ratio is calculated, so at other places the unprojected value can be used. In case of speed it should not matter, as I just replaced lat with projected lat there, so no additional comparison / calculation.

src/modules/qt/gps_parser.cpp
```diff
@@ -1013,9 +1024,11 @@ void process_gps_smoothing(gps_private_data gdata, char do_processing)
             }
             if (nr_div != 0) {
                 gps_points_p[i].lat = lat_sum / nr_div;
+                gps_points_p[i].lat_projected = project_latitude(gps_points_p[i].lat);
                 gps_points_p[i].lon = lon_sum / nr_div;
             } else {
                 gps_points_p[i].lat = gps_points_r[i].lat;
+                gps_points_p[i].lat_projected = gps_points_r[i].lat;
```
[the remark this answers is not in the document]
ddennedy: I am willing to merge this if you address some of these review comments. Why haven’t you?

src/modules/qt/filter_gpsgraphic.cpp
```diff
@@ -488,30 +488,15 @@ static void find_minmax_of_data(mlt_filter filter)
 #undef assign_if_smaller
 #undef assign_if_bigger
 
-    //compute the map aspect ratio using real distances (used to correctly overlay background image over gps track)
+    // //compute the gps track aspect ratio (from coords) // the other one seems better almost every time
```
dany123: the 2nd comment got merged from below, you can delete that

src/modules/qt/filter_gpsgraphic.cpp
```diff
@@ -472,8 +472,8 @@ static void find_minmax_of_data(mlt_filter filter)
         m = x;
     for (int i = 0; i < pdata->gps_points_size; i++) {
         gps_point_proc *crt = &pdata->gps_points_p[i];
-        assign_if_smaller(crt->lat, pdata->minmax.min_lat);
-        assign_if_bigger(crt->lat, pdata->minmax.max_lat);
+        assign_if_smaller(crt->lat_projected, pdata->minmax.min_lat);
+        assign_if_bigger(crt->lat_projected, pdata->minmax.max_lat);
```
dany123: As I mentioned in the longer comment the middle point computation (lines 501-504) needs to be fixed otherwise the user can't get the proper background map coords.

This is the only important one, we can treat the other 2 lat_projected -> [real] lat later.

dany123 (changes requested):

dany123:
> I've attached some samples, old is without my patch, new is with my patch

Thank you for the example, I haven't used such long tracks on detailed maps before so I didn't catch the problem in my usual exports. 

This is a good fix. Just address the small comments (including the ones from copilot) and it's ok from me.

src/modules/qt/filter_gpsgraphic.cpp
dany123: I just want to clarify here just in case the math works out different from what I understood it: the original map image boundaries and center stay the same as from before projection changes, correct? It is the drawn GPS track that gets re-projected to properly match the map so we still need the original latitude to get the center of the needed map image.

src/modules/qt/gps_parser.cpp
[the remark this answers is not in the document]
gnampf1: Should be fixed with the now commit now

src/modules/qt/filter_gpsgraphic.cpp
gnampf1: Min/Max is now available both unprojected and projected, for use at the different places. 
Not-Dot-Text and Grid-Text are now calculated back from the projected coordinates, as we can't use the original gps data, and due to the non linear type of math we cannot simply use percent of total. Downside is that it will affect the performance if grid or now dot text is enabled, as we have to do the calculation then for every frame I guess.

There are 2 problems in using the gps coordinates:
1) you get the wrong aspect ratio, which leads to the background image having to little height. But even if you change this by scaling the image, the is 2)
2) its not really a linear projection due to Tangens and Logarithm. Thats why its more obvious if the distance is larger, than with a small distance.

Center will always be the same. Output of center coordinates is now also gps-coordinate-system again, and needs to be used with the online maps, as they take GPS coordinates, but project them for display. Same if you work with the boundaries, like I do with the program "tile-stitch" to generate the background image.

src/modules/qt/gps_parser.cpp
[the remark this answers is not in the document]
dany123: I think you missed this one on the else block, the assignment on line 1036 should either be:
gps_points_p[i].lat_projected = gps_points_r[i].lat_projected;     //because it was computed at file read and is the original raw point
or
gps_points_p[i].lat_projected = project_latitude(gps_points_r[i].lat);   //to just compute it again

src/modules/qt/filter_gpsgraphic.cpp
dany123: Thanks for the details. This looks good to me now.

PR #1185, opened 2025-12-14
Replace strptime with portable code

jlskuz:
This change removes the usage of `strptime`, because it is not portable (not supported by MinGW and MSVC). We only use it for the GPS filters which use Qt anyways hence we can use QDateTime instead.

For quality assurance during the port I wrote tests for `DatetimeXMLstring_to_mseconds`. This is the first test for code that is not part of the framework. Hence some refactoring of CMake code was required.


bmatherly:
The strptime.c file is there to provide the function for toolchains that do not have it. We are currently compiling it with Mingw. Can you explain more why it is not working for you?

It is convenient that you can replace it with QT functions. But that won't help future uses that might need the same functionality in a different module that does not have access to QT.

I do not think the unit test is necessary since the implementation is such a simple wrapper around QT functions.

Did you know that we have a different implementation in producer.c which uses sscanf?
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_producer.c#L1222
Maybe we could unity all the date parsing functions so that we are not maintaining multiple.

jlskuz:
> The strptime.c file is there to provide the function for toolchains that do not have it. We are currently compiling it with Mingw. Can you explain more why it is not working for you?

Yes with MinGW it works so far, however with MSVC (which I am aiming for) the code in `strptime.c` does not compile due to an error related to `tzname`. It probably isn't too hard to fix, but my idea was to replace it with something that works cross-plattform and avoids custom implementations instead of creating further hacks.

> But that won't help future uses that might need the same functionality in a different module that does not have access to QT.

That is right, but my focus was on the present and what I saw was only one user in the Qt module. If you both prefer to keep it, I can find a different solution, but keeping something for the uncertain future just in case does in my opinion not make sense.

> Did you know that we have a different implementation in producer.c which uses sscanf?
https://github.com/mltframework/mlt/blob/master/src/framework/mlt_producer.c#L1222
Maybe we could unity all the date parsing functions so that we are not maintaining multiple.

I did not know that and will check it out. Unifying similar implementations does certainly sound like a good idea.

> I do not think the unit test is necessary since the implementation is such a simple wrapper around QT functions.

Let's discuss this after we have a decision on the general direction and know whether we keep the Qt implementation.

_nb. my general goal is to have as little platform specific code as possible. Unfortunately the usage of POSIX features makes that hard, but in some cases it is hardly avoidable given that MLT is for most parts is a C project and not a C++ project._

ddennedy:
> with MSVC (which I am aiming for)

When I was writing the last release notes, I was uncertain about the current status of this. I wrote the vague "Various changes to work with MSVC compiler." Can you comment about this please? Does it compile and basically run yet using most modules that default to ON?

jlskuz:
> When I was writing the last release notes, I was uncertain about the current status of this. I wrote the vague "Various changes to work with MSVC compiler." Can you comment about this please? Does it compile and basically run yet using most modules that default to ON?

No, it doesn't yet. With my recent two patches and this patch it builds with all modules disabled. My roadmap is:
- get this patch merged (or a different one solving the issue)
- Add CI for MSVC and all modules disabled
- Enable modules one by one and fix issues as they appear
- Start to test this at runtime and fix runtime issues

ddennedy:
I appreciate both perspectives, and I like the introduction of tests for the modules..

> does not compile due to an error related to tzname

You just added
```c
#ifdef _MSC_VER
#define strncasecmp _strnicmp
#define strcasecmp _stricmp
#endif
```

It looks to me like you could have added
```c
#define tzset _tzset
#define tzname _tzname
```

> keeping something for the uncertain future just in case does in my opinion not make sense.

I am pessimistic there is much need to parse date strings in a multimedia processing framework, which explains why it is confined to this recently added GPS in the Qt module. So, I lean towards replacing it with Qt; it makes much sense to use that in a Qt module. Please fix the merge conflict.

PR #1186, opened 2025-12-16
OpenFX Module

joinlaw:
continuation on the close PR #1051 on this version unsupported plugins such as draw suite and opengl plugins are filter out so they don't appear on the plugins (filters) list.

Some plugins such as `com.genarts:OpenGLSamplePlugin` and `net.sf.openfx.Shadertoy` doesn't return `kOfxStatErrMissingHostFeature` status in `kOfxActionDescribe` and `kOfxImageEffectActionDescribeInContext` OpenFX mainEntry action but they are harmless when they run because they have no effect when mentioned suits are not available.

Some plugins such as `com.genarts:OpenGLSamplePlugin` contain colons `:` in the pluginIdentifier which I replace with `^` (`openfx.com.genarts^OpenGLSamplePlugin`) to avoid any issue with MLT filters naming constraints.

### DIFF
--- a/src/modules/openfx/factory.c
+++ b/src/modules/openfx/factory.c
@@ -168,16 +168,8 @@ MLT_REPOSITORY
         setenv("OCIO", "ocio://ocio://default", 1);
     }
 
-    char *dir,
-        *openfx_path = getenv("OFX_PLUGIN_PATH"),
-        *load_unsupported_plugins = getenv(
-            "MLT_OFX_LOAD_UNSUPPORTED_PLUGINS"); /* Load unsupported plugins for debugging purposes */
-
-    bool is_load_unsupported_plugins = false;
-    if (load_unsupported_plugins) {
-        is_load_unsupported_plugins = strcmp(load_unsupported_plugins, "true") == 0;
-    }
-
+    char *dir;
+    char *openfx_path = getenv("OFX_PLUGIN_PATH");
     size_t archstr_len = strlen(OFX_ARCHSTR);
 
     mltofx_context = mlt_properties_new();
@@ -216,6 +208,8 @@ MLT_REPOSITORY
 
                     ofx_get_plugin = dlsym(dlhandle, "OfxGetPlugin");
                     ofx_get_number_of_plugins = dlsym(dlhandle, "OfxGetNumberOfPlugins");
+                    if (!ofx_get_plugin || !ofx_get_number_of_plugins)
+                        goto parse_error;
                     NumberOfPlugins = ofx_get_number_of_plugins();
 
                     char dl_n[16] = {
@@ -257,16 +251,23 @@ MLT_REPOSITORY
 
                     for (int i = 0; i < NumberOfPlugins; ++i) {
                         OfxPlugin *plugin_ptr = ofx_get_plugin(i);
+                        if (!plugin_ptr)
+                            goto parse_error;
+
+                        int detected = mltofx_detect_plugin(plugin_ptr);
+
+                        if (!detected)
+                            continue;
 
                         char *s = NULL;
                         size_t pluginIdentifier_len = strlen(plugin_ptr->pluginIdentifier);
                         s = malloc(pluginIdentifier_len + 8);
                         sprintf(s, "openfx.%s", plugin_ptr->pluginIdentifier);
 
-                        /* if colon `:` exists in plugin identifier
-			   change it to accent sign `^` because `:`
-			   can cause issues with mlt if put in filter
-			   name */
+                        // if colon `:` exists in plugin identifier
+                        // change it to accent sign `^` because `:`
+                        // can cause issues with mlt if put in filter
+                        // name
                         char *str_ptr = strchr(s, ':');
                         while (str_ptr != NULL) {
                             *str_ptr++ = '^';
@@ -285,20 +286,12 @@ MLT_REPOSITORY
                         mlt_properties_set(p, "dli", dl_n);
                         mlt_properties_set_int(p, "index", i);
 
-                        /* Sometimes error codes other than kOfxStatErrMissingHostFeature
-			   returned from kOfxActionDescribe like kOfxStatErrMemory, kOfxStatFailed, kOfxStatErrFatal */
-                        bool plugin_supported = mltofx_is_plugin_supported(plugin_ptr)
-                                                == kOfxStatOK;
-                        /* WIP: this is only creating them as filter I should find a way to see howto detect producers
-			   if they exists in OpenFX plugins
-			*/
-                        if (plugin_supported || is_load_unsupported_plugins) {
-                            MLT_REGISTER(mlt_service_filter_type, s, filter_openfx_init);
-                            MLT_REGISTER_METADATA(mlt_service_filter_type,
-                                                  s,
-                                                  metadata,
-                                                  "filter_openfx.yml");
-                        }
+                        MLT_REGISTER(mlt_service_filter_type, s, filter_openfx_init);
+                        MLT_REGISTER_METADATA(mlt_service_filter_type,
+                                              s,
+                                              metadata,
+                                              "filter_openfx.yml");
+                        free(s);
                     }
 
                 parse_error:
--- a/src/modules/openfx/mlt_openfx.c
+++ b/src/modules/openfx/mlt_openfx.c
@@ -1510,30 +1510,16 @@ OfxHost MltOfxHost = {NULL, MltOfxfetchSuite};
 void mltofx_init_host_properties(OfxPropertySetHandle host_properties)
 {
     propSetString(host_properties, kOfxPropName, 0, "MLT");
-    propSetString(host_properties, kOfxImageEffectPropContext, 0, kOfxImageEffectContextGeneral);
-    propSetString(host_properties,
-                  kOfxImageEffectPropSupportedPixelDepths,
-                  0,
-                  kOfxBitDepthByte); /* kOfxBitDepthByte is hardcoded */
+    propSetString(host_properties, kOfxImageEffectPropContext, 0, kOfxImageEffectContextFilter);
+    propSetString(host_properties, kOfxImageEffectPropSupportedPixelDepths, 0, kOfxBitDepthByte);
     propSetString(host_properties, kOfxImageEffectPropSupportedPixelDepths, 1, kOfxBitDepthShort);
-    propSetString(host_properties, kOfxImageEffectPropSupportedPixelDepths, 2, kOfxBitDepthHalf);
-    propSetString(host_properties, kOfxImageEffectPropSupportedPixelDepths, 3, kOfxBitDepthFloat);
 
     propSetString(host_properties,
                   kOfxImageEffectPropSupportedComponents,
                   0,
-                  kOfxImageComponentNone);
-    propSetString(host_properties,
-                  kOfxImageEffectPropSupportedComponents,
-                  1,
                   kOfxImageComponentRGBA);
-    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 2, kOfxImageComponentRGB);
-    propSetString(host_properties,
-                  kOfxImageEffectPropSupportedComponents,
-                  3,
-                  kOfxImageComponentAlpha);
-
-    propSetInt(host_properties, kOfxImageEffectPropSupportsMultipleClipDepths, 0, 1);
+    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 1, kOfxImageComponentRGB);
+    propSetInt(host_properties, kOfxImageEffectPropSupportsMultipleClipDepths, 0, 0);
 
     /* TODO: add support for OpenGL plugins */
     propSetString(host_properties, kOfxImageEffectPropOpenGLRenderSupported, 0, "false");
@@ -1783,7 +1769,7 @@ void mltofx_set_output_clip_data(OfxPlugin *plugin,
                   (double) width / (double) height);
 }
 
-OfxStatus mltofx_is_plugin_supported(OfxPlugin *plugin)
+int mltofx_detect_plugin(OfxPlugin *plugin)
 {
     mlt_properties image_effect = mlt_properties_new();
     mlt_properties clips = mlt_properties_new();
@@ -1825,20 +1811,25 @@ OfxStatus mltofx_is_plugin_supported(OfxPlugin *plugin)
     propSetString((OfxPropertySetHandle) props,
                   kOfxImageEffectPropContext,
                   0,
-                  kOfxImageEffectContextGeneral);
+                  kOfxImageEffectContextFilter);
 
     plugin->setHost(&MltOfxHost);
 
     OfxStatus status_code = kOfxStatErrUnknown;
     status_code = plugin->mainEntry(kOfxActionLoad, NULL, NULL, NULL);
     mltofx_log_status_code(status_code, "kOfxActionLoad");
+
+    if (status_code != kOfxStatOK) {
+        return 0;
+    }
+
     status_code
         = plugin->mainEntry(kOfxActionDescribe, (OfxImageEffectHandle) image_effect, NULL, NULL);
     mltofx_log_status_code(status_code, "kOfxActionDescribe");
 
     if (status_code != kOfxStatOK) {
         plugin->mainEntry(kOfxActionUnload, NULL, NULL, NULL);
-        return status_code;
+        return 0;
     }
 
     /* test describe in context action because some plugins say they are unsupported at this stage */
@@ -1847,14 +1838,57 @@ OfxStatus mltofx_is_plugin_supported(OfxPlugin *plugin)
                                     (OfxPropertySetHandle) MltOfxHost.host,
                                     NULL);
 
+    int describe_in_context_valid = 0;
     mltofx_log_status_code(status_code, "kOfxImageEffectActionDescribeInContext");
     if (status_code != kOfxStatErrMissingHostFeature) {
         plugin->mainEntry(kOfxActionUnload, NULL, NULL, NULL);
-        return kOfxStatOK;
+        describe_in_context_valid = 1;
     }
 
+    int i = 0;
+    int count = 0;
+
+    propGetDimension((OfxPropertySetHandle) props, kOfxImageEffectPropSupportedContexts, &count);
+    for (i = 0; i < count; i++) {
+        char *context;
+        propGetString((OfxPropertySetHandle) props,
+                      kOfxImageEffectPropSupportedContexts,
+                      i,
+                      &context);
+        if (!strcmp(context, kOfxImageEffectContextFilter)) {
+            break;
+        }
+    }
+    if (i == count) {
+        mlt_log_debug(NULL, "[ofx] Plugin not a filter: %s\n", plugin->pluginIdentifier);
+        /* since plugin is not filter then load fail so we must not unload it */
+        return 0;
+    }
+
+    count = 0;
+
+    propGetDimension((OfxPropertySetHandle) props, kOfxImageEffectPropSupportedPixelDepths, &count);
+    for (i = 0; i < count; i++) {
+        char *depth;
+        propGetString((OfxPropertySetHandle) props,
+                      kOfxImageEffectPropSupportedPixelDepths,
+                      i,
+                      &depth);
+        if (!strcmp(depth, kOfxBitDepthByte) || !strcmp(depth, kOfxBitDepthShort)) {
+            break;
+        }
+    }
+    if (i == count) {
+        mlt_log_debug(NULL, "[ofx] Plugin not byte/short: %s\n", plugin->pluginIdentifier);
+        /* since no pixel depth is supported by us then plugin is load fail so we must not unload it */
+        return 0;
+    }
+
+    if (describe_in_context_valid)
+        return 1;
+
     plugin->mainEntry(kOfxActionUnload, NULL, NULL, NULL);
-    return status_code;
+    return 0;
 }
 
 void *mltofx_fetch_params(OfxPlugin *plugin, mlt_properties params, mlt_properties mlt_metadata)
@@ -1898,7 +1932,7 @@ void *mltofx_fetch_params(OfxPlugin *plugin, mlt_properties params, mlt_properti
     propSetString((OfxPropertySetHandle) props,
                   kOfxImageEffectPropContext,
                   0,
-                  kOfxImageEffectContextGeneral);
+                  kOfxImageEffectContextFilter);
 
     plugin->setHost(&MltOfxHost);
     OfxStatus status_code = kOfxStatErrUnknown;
--- a/src/modules/openfx/mlt_openfx.h
+++ b/src/modules/openfx/mlt_openfx.h
@@ -75,7 +75,7 @@ void mltofx_set_output_clip_data(OfxPlugin *plugin,
                                  int height,
                                  mlt_image_format format);
 
-OfxStatus mltofx_is_plugin_supported(OfxPlugin *plugin);
+int mltofx_detect_plugin(OfxPlugin *plugin);
 
 void *mltofx_fetch_params(OfxPlugin *plugin, mlt_properties params, mlt_properties mlt_metadata);
 


joinlaw:
@luzpaz

luzpaz:
@joinlaw any chance you can show the feature in action ? (like a screencast?)

joinlaw:
@luzpaz 

demos

gmic-ball-plugin

gmic-circle-art-plugin

[basic-gain-plugin.webm](https://github.com/user-attachments/assets/3948f5f1-088a-4dc5-b7f4-6ca6aec02f32)

ddennedy:
Thank you, I can start taking a look around middle of January. I am taking a break over the holiday before some release activities at the end of the year, which is often followed by increased support.

src/modules/openfx/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+  filter_openfx.c
+)
+
+add_custom_target(Other_openfx_Files SOURCES
+  filter_openfx.yml
+)
+
+target_compile_options(mltopenfx PRIVATE ${MLT_COMPILE_OPTIONS})
+
+target_link_libraries(mltopenfx PRIVATE mlt m PkgConfig::glib ${CMAKE_DL_LIBS})
```
ddennedy: All modules now need some things to handle the Microsoft Visual C++ compiler. You can see recent changes (git log) of other modules' CMakeLists.txt.

```suggestion
include(GenerateExportHeader)
generate_export_header(mltopenfx)
target_compile_options(mltopenfx PRIVATE ${MLT_COMPILE_OPTIONS})
target_include_directories(mltopenfx PRIVATE ${CMAKE_CURRENT_BINARY_DIR})

target_link_libraries(mltopenfx PRIVATE mlt PkgConfig::glib ${CMAKE_DL_LIBS})
if(NOT MSVC)
  target_link_libraries(mltopenfx PRIVATE m)
endif()
```

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,313 @@
+/*
+ * factory.c -- the factory method interfaces
+ * Copyright (C) 2024 Meltytech, LLC
```
ddennedy: Change every copyright to the year 2025.

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,313 @@
...
+                                     mlt_service_type type,
+                                     const char *id,
+                                     char *arg);
+
+static void plugin_mgr_destroy(mlt_properties p)
+{
+    int cN = mlt_properties_count(mltofx_context);
+
+    int j;
+    for (j = 0; j < cN; ++j) {
```
ddennedy: Feel free to use modern C from 1999 (C99) that allows declaring variables in the for loop initializer:
```suggestion
    for (int j = 0; j < cN; ++j) {
```

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,313 @@
...
+    mltofx_init_host_properties(MltOfxHost.host);
+
+    char *dir, *openfx_path = getenv("OFX_PLUGIN_PATH"),
+      *load_unsupported_plugins = getenv("MLT_OFX_LOAD_UNSUPPORTED_PLUGINS"); /* Load unsupported plugins for debugging purposes */
+
+    bool is_load_unsupported_plugins = false;
+    if (load_unsupported_plugins)
+      {
+	is_load_unsupported_plugins = strcmp(load_unsupported_plugins, "true") == 0 ? true : false;
+      }
```
ddennedy: Code formatting is messy here. We want code to be formatted with clang-format-14. We include a `.clang-format` file for it as well as build targets `clang-format` and `clang-format-check`. You can also see frequent commit messages "clang-format" where one of the maintainers had to cleanup, frequently from an outsider's pull request. If you have trouble locating a static build of clang-format-14, use the next closest version that is convenient for you to at least get close.

Also, I see you are using `<stdbool.h>`, but it took me a while to find it and caused some confusion. It is better to include it directly instead of through "mlt_openfx.h". Also, it is bad style in any language to do things like `if (x == true)` or `expression ? true : false`. The boolean literals are unnecessary clutter. This should be like:

`is_load_unsupported_plugins = strcmp(load_unsupported_plugins, "true") == 0;`

src/modules/openfx/filter_openfx.c
```diff
@@ -0,0 +1,274 @@
...
+{
+    mlt_filter filter = (mlt_filter) mlt_frame_pop_service(frame);
+
+    mlt_properties properties = MLT_FILTER_PROPERTIES(filter);
+    OfxPlugin *plugin = mlt_properties_get_data(properties, "ofx_plugin", NULL);
+    mlt_properties image_effect = mlt_properties_get_data(properties, "ofx_image_effect", NULL);
+    mlt_properties params = mlt_properties_get_data(image_effect, "mltofx_params", NULL);
+    mlt_properties image_effect_params = mlt_properties_get_data(image_effect, "params", NULL);
+
+    *format = mlt_image_rgba;
```
ddennedy: I am not going to accept any new code that is 8-bit only. I see OpenFX has `kOfxBitDepthShort` that appears to correspond to the `mlt_image_rgba64` that we added in 2025. Please try to include it.

src/modules/openfx/filter_openfx.yml
```diff
@@ -0,0 +1,12 @@
...
+identifier: openfx
+title: OpenFX
+version: 1
+license: GPLv2
+language: en
+url: https://openeffects.org/
+creator: mr.fantastic <mrfantastic@firemail.cc>
+tags:
+  - Video
+description: Process videos using OpenFX plugins.
```
ddennedy: Please specify which versions of OpenFX this supports.

src/modules/openfx/ofxDialog.h
```diff
@@ -0,0 +1,72 @@
+
+#ifndef _ofxDialog_h_
+#define _ofxDialog_h_
+
+// Copyright OpenFX and contributors to the OpenFX project.
```
ddennedy, line 5: Please add somewhere (probably not here since it might be overwritten when these files get updated) the URL of your source for these OpenFX files. Maybe in `CMakeLists.txt` or `mlt_openfx.c`

src/modules/openfx/mlt_openfx.c
```diff
@@ -0,0 +1,2299 @@
...
+
+    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 0, kOfxImageComponentNone);
+    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 1, kOfxImageComponentRGBA);
+    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 2, kOfxImageComponentRGB);
+    propSetString(host_properties, kOfxImageEffectPropSupportedComponents, 3, kOfxImageComponentAlpha);
+
+    propSetInt(host_properties, kOfxImageEffectPropSupportsMultipleClipDepths, 0, 1);
+
+    /* TODO: add support for OpenGL plugins */
+    propSetString(host_properties, kOfxImageEffectPropOpenGLRenderSupported, 0, "false");
```
ddennedy, line 1525: We might not ever be able to support for OpenGL (or CUDA or Metal) without a good way to prevent multiple heavy data transfers between CPU and GPU RAM. For example, the Movit module and proper management by the app such as Shotcut addresses that.

No code change requested here. Just FYI to inform any future plans. I guess the key point is that OpenGL integration should not be standalone and needs to coordinate with the movit module. Also, the `qglsl` consumer is a way for `melt` to provide an OpenGL context on the render thread. Also, the future of OpenGL is also rather uncertain. The GPU tech landscape is messy.

src/modules/openfx/filter_openfx.c
```diff
@@ -0,0 +1,274 @@
...
+        uint8_t *src_copy = malloc(*width * *height * 4);
+        if (src_copy == NULL)
+            goto out;
+        memcpy(src_copy, *image, *width * *height * 4);
+        mltofx_set_source_clip_data(plugin, image_effect, src_copy, *width, *height);
+        mltofx_set_output_clip_data(plugin, image_effect, *image, *width, *height);
+
+        mltofx_action_render(plugin, image_effect, *width, *height);
+
+        free(src_copy);
```
ddennedy: You need to use `mlt_pool_alloc` for media memory allocations in MLT. Also, you should use the framework to calculate the size. Basically, update a lot of this code to use `mlt_image` and its functions:
`mlt_image_set_values()`, `mlt_image_alloc_data()` (uses mlt_pool), `mlt_image_calculate_size()` (for memcpy), and `mlt_image_close()`.

src/modules/openfx/filter_openfx.c
```diff
@@ -0,0 +1,274 @@
...
+        mltofx_get_clip_preferences(plugin, image_effect);
+
+        uint8_t *src_copy = malloc(*width * *height * 4);
+        if (src_copy == NULL)
+            goto out;
+        memcpy(src_copy, *image, *width * *height * 4);
+        mltofx_set_source_clip_data(plugin, image_effect, src_copy, *width, *height);
+        mltofx_set_output_clip_data(plugin, image_effect, *image, *width, *height);
+
+        mltofx_action_render(plugin, image_effect, *width, *height);
```
ddennedy: I am worried that calling the plugin from multiple frame threads is not safe. I do not see where `kOfxImageEffectPluginRenderThreadSafety` is checked. MLT supports multiple render threads also known as frame-threading when the absolute value of the consumer `real_time` property is greater than 1 (value specifies number of threads). Thus, you likely need to use `mlt_service_lock()` and `mlt_service_unlock()` around at least the render function and maybe more. Think: is it safe to change these plugin properties from another thread while one thread is in `mltofx_action_render()`?

izakikoparadox:
I was literally just searching the progress for OFX support and I'm pleasantly surprised to see effort being made within hours ago. I guess the big question is, Is there a supposed date when I'd finally be able to use OFX plugins in kdenlive?

bmatherly:
@joinlaw I am interested in helping to sponsor this change into the code base. I intend to fetch your branch and start testing. Do you have any work-in-progress that I should have? Also, do you want any help resolving the comments above? Let me know how i can help.

joinlaw:
> @joinlaw I am interested in helping to sponsor this change into the code base. I intend to fetch your branch and start testing. Do you have any work-in-progress that I should have? Also, do you want any help resolving the comments above? Let me know how i can help.

Well after this first merge request get merged the main priority is to implement OpenFX OpenGL plugins suite and DrawSuite (openfx extension something like postscript or cairo vector graphics).

Once I have free time I will resolve all comments by @ddennedy and hopefully we will have basic OpenFX support so we can implement all other openfx extensions later.

src/modules/openfx/ofxDialog.h
```diff
@@ -0,0 +1,72 @@
+
```
bmatherly, line 1: We do not want to duplicate the OFX host header files in our source tree. Today, when installing the OFX host environment for myself, I was surprised to find that they do not offer an install target for the host files. HOWEVER, they do offer pre-packaged development files with each release:
https://github.com/AcademySoftwareFoundation/openfx/actions/runs/19546967037
The directory structure of that package is documented in their github workflow:
https://github.com/AcademySoftwareFoundation/openfx/blob/main/.github/workflows/build.yml#L364
We should remove all these header files, and require the packager to have the official package from OFX installed - or to provide their own if they build it themselves. This will allow us to support compilation with different versions of the OFX framework over time.

As a convenience, we could offer a script for the packager to download and extract the official packages. That is not a requirement from me - just a suggestion.

joinlaw: and possibly using a git submodule

bmatherly: MLT would still need libOfxHost.a and libOfxSupport.a. Where would they get those from? And how would we guarantee that the tag/release that we chose to get with git submodule matches the library versions the packager has? 

I am open minded, but i do not want to build OFX as part of building this OFX MLT module because then our build would depend on conan - which is completely unnecessary if the packager just uses the official package release from OFX.

ddennedy:
Please turn off this new module in `.github/workflows/build-windows-msvc.yml` for now. There is still a lot work to do to get modules, in general, to build with MSVC.

src/modules/openfx/ofxDialog.h
joinlaw: `libOfxSupport.a` is utility library to be used by plugin developers and as you see it is statically linked by the developers.

`libOfxHost.a` is sample/example reference implementation C++ library classes and my code is free from it because openfx module is pure C not C++.

The only thing needed is the header files at this folder only https://github.com/AcademySoftwareFoundation/openfx/tree/main/include so we don't need conan.

bmatherly:
@joinlaw, can you provide more commentary about how we can help potential users get access to openfx plugins?

For example, I see you are using G'MIC plugins in your working copy. I have seen previous requests for that. So, I think in order to provide that, I will need:
* G'MIC library
* One or possibly more OFX plugins that adapt G'MIC to openfx. I see a couple projects out there
  * https://github.com/NatronGitHub/openfx-gmic
  * https://github.com/MrKepzie/openfx-gmic

I haven't found binary releases on the GitHub pages. And my Unbuntu package manger does not seem to provide the OFX files (it does provide the G'MIC library).

Where will users get the OFX files from to use with this module? Will KDENLIVE help users download them?

joinlaw:
> @joinlaw, can you provide more commentary about how we can help potential users get access to openfx plugins?
> 
> For example, I see you are using G'MIC plugins in your working copy. I have seen previous requests for that. So, I think in order to provide that, I will need:
> 
>     * G'MIC library
> 
>     * One or possibly more OFX plugins that adapt G'MIC to openfx. I see a couple projects out there
>       
>       * https://github.com/NatronGitHub/openfx-gmic
>       * https://github.com/MrKepzie/openfx-gmic
> 
> 
> I haven't found binary releases on the GitHub pages. And my Unbuntu package manger does not seem to provide the OFX files (it does provide the G'MIC library).
> 
> Where will users get the OFX files from to use with this module? Will KDENLIVE help users download them?

Actually I downloaded the Natron portable package and used them also I built the example plugins nothing more than that but I have seen before few other free software plugins over the internet.

I added them to the environment variable

```
export OFX_PLUGIN_PATH=/path/to/openfx/examples/destdir:/path/to/Natron-2.5.0-Linux-x86_64-no-installer/Plugins/OFX/Natron
```

ddennedy:
There are a lot of OpenFX plugins, many of them commercial. Besides web search there are a lot of other so-called organic means to learn about them such as special interest web sites and forums. It is a lot like the VST plugin ecosystem. Expect to potentially deal with a lot of support and troubleshoot with the diversity. Properly supporting and debugging them in host may also cost money if you want to actually provide assistance.

src/modules/openfx/filter_openfx.c
joinlaw: In 3f8f847 I add code that will use rgba64 or rgba or rgb when plugin support it.

But after converting frame to `rgba64` and getting it back to `yuv422` the image become black this happen to me in kdenlive and melt how can I fix this?.

src/modules/openfx/ofxDialog.h
bmatherly: I am still not in favor of including all these header files. But if this is the best way, I could accept if they are moved to a "thirdparty" folder as is a convention in other projects.

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,312 @@
...
+                    for (int i = 0; i < NumberOfPlugins; ++i) {
+                        OfxPlugin *plugin_ptr = ofx_get_plugin(i);
+
+                        char *s = NULL;
+                        size_t pluginIdentifier_len = strlen(plugin_ptr->pluginIdentifier);
+                        s = malloc(pluginIdentifier_len + 8);
+                        sprintf(s, "openfx.%s", plugin_ptr->pluginIdentifier);
+
+                        /* if colon `:` exists in plugin identifier
+			   change it to accent sign `^` because `:`
```
bmatherly: Thanks for running clang-format. But there are still tabs sprinkled around in the block comments. I suppose clang-format does not fix those  because they are in block comments. Here and some other places.

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,312 @@
...
+                            params,
+                            0,
+                            (mlt_destructor) mlt_properties_close,
+                            NULL);
+
+    mltofx_fetch_params(pt, params);
+    return result;
+}
+
+MLT_REPOSITORY
```
bmatherly, line 159: When I run melt, i always see a new information message:

> [OpenColorIO Info]: Color management disabled. (Specify the $OCIO environment variable to enable.)

This message is printed even if I do not try to use any openfx filters. Is there a way to avoid the message?

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,312 @@
...
+#include "shlobj.h"
+#include "tchar.h"
+#endif
+
+extern mlt_filter filter_openfx_init(mlt_profile profile,
+                                     mlt_service_type type,
+                                     const char *id,
+                                     char *arg);
+
+static void plugin_mgr_destroy(mlt_properties p)
```
bmatherly, line 90: When I run melt and use an openfx filter, melt always ends with a crash. Here is the gdb backtrace:

```
0x00007fffe6f3d385 in ?? () from /home/brian/Downloads/Natron-2.5.0-Linux-x86_64-no-installer/Plugins/OFX/Natron/Misc.ofx.bundle/Contents/Linux-x86-64/Misc.ofx
(gdb) 
#0  0x00007fffe6f3d385 in ??? () at /home/brian/Downloads/Natron-2.5.0-Linux-x86_64-no-installer/Plugins/OFX/Natron/Misc.ofx.bundle/Contents/Linux-x86-64/Misc.ofx
#1  0x00007fffe7809b89 in ??? () at /home/brian/Downloads/Natron-2.5.0-Linux-x86_64-no-installer/Plugins/OFX/Natron/Misc.ofx.bundle/Contents/Linux-x86-64/Misc.ofx
#2  0x00007ffff0967b70 in plugin_mgr_destroy (p=0x555555705150) at /home/brian/shotcut/src/mlt/src/modules/openfx/factory.c:108
#3  0x00007ffff7fa2abe in clear_property (self=0x55555586e450) at /home/brian/shotcut/src/mlt/src/framework/mlt_property.c:124
#4  0x00007ffff7fa433a in mlt_property_close (self=0x55555586e450) at /home/brian/shotcut/src/mlt/src/framework/mlt_property.c:920
#5  0x00007ffff7f9fe6c in mlt_properties_close (self=0x555555561360) at /home/brian/shotcut/src/mlt/src/framework/mlt_properties.c:1508
#6  0x00007ffff7f8ab3a in mlt_factory_close () at /home/brian/shotcut/src/mlt/src/framework/mlt_factory.c:520
#7  0x000055555555b5a5 in main (argc=21, argv=0x7fffffffd8d8) at /home/brian/shotcut/src/mlt/src/melt/melt.c:1155
```

Do you also experience this?

src/modules/openfx/filter_openfx.yml
```diff
@@ -0,0 +1,12 @@
...
+identifier: openfx
+title: OpenFX
+version: 1
+license: GPLv2
+language: en
+url: https://openeffects.org/
+creator: mr.fantastic <mrfantastic@firemail.cc>
+tags:
+  - Video
+description: Process videos using OpenFX 1.5 plugins.
```
bmatherly, line 12: When I print the metadata for the filter, this description is generic.

For example:
./shotcut/Shotcut/Shotcut.app/melt -query filter=openfx.net.sf.openfx.AddPlugin
The metadata shows:
description: Process videos using OpenFX 1.5 plugins.

However, when I look at the plugin description here:
[https://natron.readthedocs.io/en/rb-2.5/plugins/net.sf.openfx.AddPlugin.html](https://natron.readthedocs.io/en/rb-2.5/plugins/net.sf.openfx.AddPlugin.html)
I see:

> Description:
> Add a constant to the selected channels.

Can we update the metadata description with the description from the plugin?

src/modules/openfx/factory.c
```diff
@@ -0,0 +1,312 @@
...
+
+    OfxGetPluginFn get_plugin = mlt_properties_get_data(mltofx_dl, dli, NULL);
+    OfxPlugin *pt = get_plugin(index);
+
+    dli[0] = 's';
+    OfxSetHostFn set_host = mlt_properties_get_data(mltofx_dl, dli, NULL);
+    if (set_host != NULL)
+        set_host(&MltOfxHost);
+
+    mlt_properties_set(result, "identifier", id);
```
bmatherly, line 143: When I run this command:
`melt -query filter=openfx.net.sf.openfx.Mirror`
Here is the result:

```
[OpenColorIO Info]: Color management disabled. (Specify the $OCIO environment variable to enable.)
---
schema_version: 7.0
type: filter
identifier: openfx.net.sf.openfx.Mirror
title: openfx.net.sf.openfx.Mirror
version: 1
license: GPLv2
language: en
url: "https://openeffects.org/"
creator: mr.fantastic <mrfantastic@firemail.cc>
tags:
  - Video
description: Process videos using OpenFX 1.5 plugins.
parameters:
  Controls:
    animation: yes
  flip:
    type: boolean
    animation: yes
    title: Vertical (flip)
    identifier: flip
  flop:
    type: boolean
    animation: yes
    title: Horizontal (flop)
    identifier: flop
...
Bus error (core dumped)
```

Besides the noisy info message and crash, the metadata is ill-formed. It should be:

```
parameters:
  - identifier: flip
    type: boolean
    animation: yes
    title: Vertical (flip)

  - identifier: flop
    type: boolean
    animation: yes
    title: Horizontal (flop)
```

The "Controls" item should not be there at all. And the identifier should be first and have a hyphen. See the schema in the framework.

src/modules/openfx/filter_openfx.c
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    int error = mlt_frame_get_image(frame, image, format, width, height, 1);
+
+    if (error == 0) {
+        mlt_position position = mlt_filter_get_position(filter, frame);
+        mlt_position length = mlt_filter_get_length2(filter, frame);
+        int params_count = mlt_properties_count(params);
+
+        for (int i = 0; i < params_count; ++i) {
```
bmatherly, line 66: I've been trying to test using melt. I can get the filters to work. But I can't get any parameters to work. For example, here is a command I am using:

`melt test.mpr -filter openfx.net.sf.openfx.Mirror flip=1`

But the image is not flipped. Any idea what might be wrong?

src/modules/openfx/factory.c
joinlaw: Plugins print their logs but I think this goes to the stderr so in bash we can forward it to `2> /dev/null` this is what I know but **maybe** there is OpenFX host property that turns off logging if plugins is implemented according to the standard.

src/modules/openfx/factory.c
joinlaw: Sometimes I face similar issues with the git version of melt and not only `openfx` branch but using kdenlive does not cause this issue. I need to do more testing.

src/modules/openfx/filter_openfx.yml
joinlaw: I didn't implement it yet I will look at it.

src/modules/openfx/factory.c
bmatherly: I don't think we want to turn off logging. Here is a patch that quiets the OCIO message:

```
diff --git a/src/modules/openfx/factory.c b/src/modules/openfx/factory.c
index 3007ef68..ca006ef5 100644
--- a/src/modules/openfx/factory.c
+++ b/src/modules/openfx/factory.c
@@ -161,6 +161,13 @@ MLT_REPOSITORY
     MltOfxHost.host = (OfxPropertySetHandle) mlt_properties_new();
     mltofx_init_host_properties(MltOfxHost.host);
 
+    // Quiet warnings from plugins that use OCIO
+    char *ocio = getenv("OCIO");
+    if (!ocio || !strcmp(ocio, "")) {
+        // If the OCIO environment variable is not set, set it to a built-in default config
+        setenv("OCIO", "ocio://ocio://default", 1);
+    }
+
     char *dir,
         *openfx_path = getenv("OFX_PLUGIN_PATH"),
         *load_unsupported_plugins = getenv(

```

src/modules/openfx/factory.c
bmatherly: There seem to be a handful of plugins that crash when we call 

            pt->mainEntry(kOfxActionUnload, NULL, NULL, NULL);

From inspection, I notice that we load and then unload all plugins while we are discovering them. Then, we load them all again when we are generating metadata - and we leave them all loaded until the program closes and we call plugin_mgr_destroy. Do you know if it is good to leave them all loaded? Does that consume extra resources? Should we try to only keep open plugins that we are using?

ddennedy:
As a git submodule, now whenever I make a release source tarball I need to copy those files into it manually. That made sense for glaxnimate because it does not have a lib and was changing regularly for a while. It does not make sense for this. Leave all OpenFX dependencies external like we do for everything else today.

joinlaw:
> As a git submodule, now whenever I make a release source tarball I need to copy those files into it manually. That made sense for glaxnimate because it does not have a lib and was changing regularly for a while. It does not make sense for this. Leave all OpenFX dependencies external like we do for everything else today.

I could do that but no distribution currently is distributing this headers as a package not redhat flavours, debian or arch of course there is few free/libre OpenFX hosts currently after more libre plugins and hosts developed we will see it packaged to all major distros.

ddennedy:
I see. I wonder why I vaguely recall Brian talking about a `thirdparty` directory to put things into, but now I cannot find that comment to follow the past discussion around this. I think some other distros have OpenFX, but I wonder why not Debian and Fedora? What does Natron do... git submodule.
We do have other examples in our repo of including third party things: `decklink`, `plus/ebur128`, `plusgpl/cJSON.h`, `rtaudio`. Some reasons have been convenience for lack of packaging (`decklink`) or the very slim nature of them.

Brian, what do you want to do? Regarding git submodule, what I am saying that it is virtually including the headers in the repo since I have to copy them into the release tarball anyways. Since these intentionally do not change frequently, I am willing to accept checking them in and update manually as needed.

bmatherly:
> Brian, what do you want to do? Regarding git submodule, what I am saying that it is virtually including the headers in the repo since I have to copy them into the release tarball anyways. Since these intentionally do not change frequently, I am willing to accept checking them in and update manually as needed.

I agree with your comments about other examples of 3rd party code inside our code. I think this situation qualifies for that. I prefer a thirdparty folder and put the files in our git repository. I am also OK with manually updating them from time-to-time. I think it will be infrequent - and I volunteer to help test it.

Similar to  ebur128, I would like our build scripts to download the latest version to compile against and only fall back to internal thirdparty files as a convenience for other developers/packagers. I wonder if there is an easy way for a developer to specify an include path for a different plugin version so that they can override the files that we choose to include. Maybe a cmake parameter? I guess that could be "extra credit". After we merge this PR, I might try to work on that.

src/modules/openfx/filter_openfx.c
bmatherly: Can you provide a melt command line that demonstrates this? If so, I volunteer to look into it.

src/modules/openfx/factory.c
bmatherly: I put a commit here which you are welcome to cherry-pick
https://github.com/mltframework/mlt/commit/4cc5e4f586dffadd6eccf07c696f9795c7d4cd44

bmatherly:
Here is a proposed commit to improve the plugin detection. It limits loaded plugins to ones that support "Filter" context and support byte or short formats.

https://github.com/mltframework/mlt/commit/21f51fd4181c04518b61ce69e9d488a6411b1bdb

I notice that there are only a couple of plugins that support short, and about a dozen that support float. Most support byte only.

Tylemagne:
Is this PR proposing full OpenFX support such that something like NTSC-RS could work in KDenLive?

Ex: https://ntsc.rs/

joinlaw:
> Is this PR proposing full OpenFX support such that something like NTSC-RS could work in KDenLive?
> 
> Ex: https://ntsc.rs/

Yes it is working and ofcourse it is planned to support all plugins possible but with this pull request openfx and drawing suite plugins will not be supported yet.

the good thing that plugin does not use OpenGL


https://github.com/user-attachments/assets/b9d750d1-dc1d-4d6d-ba12-fea921a5260b

joinlaw:
> Here is a proposed commit to improve the plugin detection. It limits loaded plugins to ones that support "Filter" context and support byte or short formats.
> 
> [21f51fd](https://github.com/mltframework/mlt/commit/21f51fd4181c04518b61ce69e9d488a6411b1bdb)
> 
> I notice that there are only a couple of plugins that support short, and about a dozen that support float. Most support byte only.

Applied this and now I think the code is ready to merge :+1: .

bmatherly:
Thanks. I will take a look this weekend. Just so you know, I might wait to merge until after the pending Shotcut release.

ddennedy:
Feel free to merge it when ready, and I will disable it in the Shotcut builds until after this next stabilization release.

bmatherly:
> Applied this and now I think the code is ready to merge 👍 .

Thanks for your continued work, @joinlaw. Unfortunately, there are still issues the be worked out. I have moved the working branch into the mltframework project and I am actively working on it under this new PR:
https://github.com/mltframework/mlt/pull/1204

You are welcome to continue to make contributions to that new branch.

I am going to close this PR. Hopefully 3rd time is a charm!

PR #1189, opened 2025-12-20
Add vcpkg related files and MSVC CI job

jlskuz:
Regarding CI this is an initial draft that rebuilds all dependencies all the time. For use in production caching should be added. There are two options for caching (see https://github.com/microsoft/vcpkg-tool/pull/1662):

> 1. Use [GitHub Packages with NuGet](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=linux-runner), this method uses vcpkg's caching mechanism and preserves caching granularity per port. This is the method recommended by the vcpkg team.
> 2. Use [actions/cache](https://github.com/actions/cache) in your workflow to cache the installed directory, this is not integrated with vcpkg but is the method endorsed by the GitHub Actions team. The downside of this method is that the whole installed tree gets cached as a single artifact, so single port updates invalidate all packages.

I tested both approaches in my fork successfully. Given that the NuGet approach is recommended, I tend to use it. However I as soon as I add code for it, it will flood the registry of the mltframework GitHub organization with packages, hence I want to hear your feedback beforehand. Any objections?

### DIFF
--- a/.github/workflows/build-windows-msvc.yml
+++ b/.github/workflows/build-windows-msvc.yml
@@ -1,0 +1,29 @@
+name: build-test-windows-msvc-on-push
+
+on: [push, pull_request]
+
+jobs:
+  build-cmake-msvc:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Restore from cache and setup vcpkg executable and data files.
+        uses: lukka/run-vcpkg@v11
+        with:
+          vcpkgJsonGlob: 'vcpkg.json'
+
+      - name: Setup MSVC
+        uses: ilammy/msvc-dev-cmd@v1
+
+      - name: Configure C/C++ with CMake
+        run: |
+          cmake --preset vcpkg-ninja
+
+      - name: Build with Ninja
+        run: |
+          cmake --build build
+
+      - name: Install MLT
+        run: |
+          cmake --install build


ddennedy:
Take it out of draft if you want me to review it.

jlskuz:
I don't need a full review yet, but like to get a "go" on using the approach that involves the package registry as it will be painful to clear/revert ones I started to use it.

jlskuz:
I took it out of draft anyways now. I think 8 minutes build time without caching is acceptable as a stop gap solution and can be merged. Caching will become more important as soon as we enable heavier deps like Qt and FFmpeg.

vcpkg-configuration.json
```diff
@@ -0,0 +1,14 @@
+{
+  "default-registry": {
+    "kind": "git",
+    "baseline": "cfcbdb245f1179a5a493890a0b69531d66969e62",
```
ddennedy, line 4: How does this commit ID get updated? Can it be a tag instead?

jlskuz: One has to update that manually once in a while.

As far as I can tell from [the documentation](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-configuration-json#registry-baseline) and open feature requests (like [this](https://github.com/microsoft/vcpkg/discussions/35770) or [this](https://github.com/microsoft/vcpkg/discussions/25087)) only a long commit hash, but no tags are supported.

jlskuz:
Thanks for merging! Do you have any preference regarding the caching mechanism? If not I will open a PR with the NuGet approach soon.

ddennedy:
I do not have a preference for the caching.

PR #1193, opened 2026-01-09
Fix outline encroaches on text color

bmatherly:
As reported here: https://forum.shotcut.org/t/outline-outlining-text-in-filter-subtilte-burn-in-and-typewriter-with-unwanted-inline/50726

Posting this as a PR so we can consider if we need to do anything for backwards compatibility. I was thinking we could just push it out. But I want to make sure users would not be surprised.

I think this change significantly improves the outline.

Before:
before

After:
after

Even the maximum outline size still preserves the text fill:
image

### DIFF
--- a/src/modules/qt/filter_qtext.cpp
+++ b/src/modules/qt/filter_qtext.cpp
@@ -262,35 +262,6 @@ static QColor get_qcolor(mlt_properties filter_properties,
     return QColor(color.r, color.g, color.b, color.a);
 }
 
-static QPen get_qpen(mlt_properties filter_properties,
-                     mlt_properties frame_properties,
-                     int position,
-                     int length)
-{
-    QColor color;
-    int outline = mlt_properties_get_int(filter_properties, "outline");
-    QPen pen;
-
-    pen.setWidth(outline);
-    if (outline) {
-        color = get_qcolor(filter_properties, "olcolour", frame_properties, position, length);
-    } else {
-        color = get_qcolor(filter_properties, "bgcolour", frame_properties, position, length);
-    }
-    pen.setColor(color);
-
-    return pen;
-}
-
-static QBrush get_qbrush(mlt_properties filter_properties,
-                         mlt_properties frame_properties,
-                         int position,
-                         int length)
-{
-    QColor color = get_qcolor(filter_properties, "fgcolour", frame_properties, position, length);
-    return QBrush(color);
-}
-
 static void transform_painter(QPainter *painter,
                               mlt_rect frame_rect,
                               QRectF path_rect,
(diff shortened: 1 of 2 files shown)


ddennedy:
I think it needs some versioning since it changes behavior significantly.

bmatherly:
OK. I can bump the yaml version. Do you think we should do anything differently in Shotcut for the new version?

ddennedy:
I mean a project that was made before this change should look very similar to before. People will be surprised when they upgrade and the text outline is much thinner. You can make it a user option (property). Would you characterize this as stroke (before) and now outline? Unfortunately, we already have an `outline` property. Otherwise, that would have been perfect as a boolean for this. Maybe `outer_line`. 😆

bmatherly:
Your comment reminded me that the same change applies to producer_qtext as well. I've updated the commit to change both.

> I mean a project that was made before this change should look very similar to before. People will be surprised when they upgrade and the text outline is much thinner.

I understand what you mean. On the one-hand, I agree that this is going to change the behavior quite a bit. On the other hand, I think the old behavior is really a bug and is basically unusable beyond a tiny bit of outline.

Here is an example comparison vs the pango implementation:

**Pango**
`melt -profile atsc_1080p_60 pango text=hello size=200 outline=10 olcolour=0xff0000ff`
pango

Current qtext
`melt -profile atsc_1080p_60 qtext text=hello size=200 outline=10 olcolour=0xff0000ff`
qtext_old

New qtext
`melt -profile atsc_1080p_60 qtext text=hello size=200 outline=10 olcolour=0xff0000ff`
qtext_new

As you can see, nobody would actually want the current implementation - which is why people have complained.
So I was hoping that we might just consider this a bug fix and deal with the small amount of surprise that might come from a few people. But I am on the fence, open minded, and happy to follow your preference.

If we decide to make it backwards compatible, here is my suggestion:
* MLT: Pick a new property name for the new behavior. For the sake of this discussion, lets call it "stroke" (although, i would argue the current mode is really stroke).
* MLT: Add the stroke property to any services that use text and outline: pango producer, qtext producer, qtext filter, dynamictext filter, gpstext filter, subtitle filter, text filter, timer filter, count producer, subtitle producer
* MLT: For the qt based text services, the outline property will use current behavior while the stroke property will use new behavior. For the pango based service, the outline and stroke properties will have the same behavior
* MLT: the stroke property will always override the outline property when stroke is set. If stroke is not set, outline will be used
* MLT: the outline property for all these services can be deprecated for future removal.
* Shotcut: We will change all the text based services that have outline to use stroke instead. But we will keep the label as "Outline". 
* Shotcut: If a user opens an existing project  that used  the "outline" property, it will look exactly the same. However, if they open the filter UI, the "Outline" item will have a value of zero because it is really reading the stroke property. If the user changes the outline, the underlying stroke value will override the currently set "outline". There will be no way for them to remove the outline. So if they want to remove all outline/stroke, they will have to delete the filter and re-add it to clear the legacy "outline" property. Or maybe we can detect when "stroke" has been set and clear "outline".

ddennedy:
Ok I can be convinced to treat it as a bug. We have some goals to insulate behavior changes, but it is not 100% rigorous. But let’s get @j-b-m  approval as well since kdenlive might be affected (outside of the titler).

ddennedy:
We will make a change if JB later objects, but I am going to merge now for the Shotcut 26.1 beta.

PR #1214, opened 2026-03-27
Switch to QImage scaling when downscaling as QPainter is awful

j-b-m:
Seems like a Qt6 regression, but QPainter scaling gives terrible results when downscaling. So in such cases, scale using QImage before painting. Kdenlive issue [2123](https://invent.kde.org/multimedia/kdenlive/-/issues/2123)

Comparison of results (updated QImage scaling on top)
scaling-comparison


src/modules/qt/filter_qtblend.cpp
```diff
@@ -206,9 +206,30 @@ static int filter_get_image(mlt_frame frame,
 
     int image_size = mlt_image_format_size(*format, *width, *height, NULL);
 
+    char *interps = mlt_properties_get(frame_properties, "consumer.rescale");
+    if (interps) {
+        interps = strdup(interps);
```
bmatherly: Can you explain why you copy the string? I think you can use it directly.

bmatherly (commented):
The artifacts kind of look like the operation is not working on pixels that have some transparency (around the border of the logos). I did not recreate the problem myself. Does it only happen on images with transparency? Or does it still occur if the image background is fully opaque? I'm just wondering if there is some parameter we need to pass to the painter to make sure it handles transparency (alpha) correctly.

ddennedy:
> does it still occur if the image background is fully opaque?

Looking at the images attached here, even the lines inside the opaque areas have very bad aliasing suggesting no interpolation.

ddennedy:
JB, are the images you provided scaled-up for emphasis?

In the docs, both QPainter and QImage say the hints provided both do bilinear interpolation:
https://doc.qt.io/qt-6/qpainter.html#RenderHint-enum
https://doc.qt.io/qt-6/qt.html#TransformationMode-enum

However, Claude suggested this difference:

The Qt docs description is misleading/simplified. Despite the "bilinear filtering" wording, Qt's internal implementation of `QImage::scaled()` with `SmoothTransformation` uses an area-averaging algorithm (`qSmoothScaleImage` → `qImgScaleAARGBA` in Qt's source), not just a bilinear 4-sample lookup.

The meaningful difference is:

| Method | Downscaling algorithm |
|--------|--------|
| `QPainter` with `SmoothPixmapTransform` | Bilinear: maps each output pixel back to the source and samples only the 4 nearest source pixels |
| `QImage::scaled(SmoothTransformation)` | Area-weighted: averages all source pixels that contribute to each output pixel, weighted by their coverage |

For a 4× downscale, bilinear reads 4 out of ~16 contributing source pixels; the area-average reads all ~16. The Qt docs use "bilinear" as a loose synonym for "smooth/filtered" vs "nearest-neighbor" — not as a precise description of the algorithm.

So the fix is correct: pre-scaling via `QImage::scaled(Qt::SmoothTransformation)` produces proper area-averaged results, while relying on QPainter's transform for downscaling only ever bicubic/bilinearly-samples from the original high-res source, causing aliasing.

src/modules/qt/filter_qtblend.cpp
ddennedy: Using `strdup` on this string property is a common thing in our source code that dates way back. The property getter returns a pointer, and after the call the string it points to can be changed including NULLed, leading to comparison over-reads or crashes. I think this is more of an issue for a string property on a service getting changed by an application, but these are checking a frame property set by `mlt_consumer`  that is not going to be mutated by another thread. 

This technique goes back to a bug fix in 9a19d7b3c. The bug report is not available, and there are other fixes mixed in with that commit. So, hard to say what was the actual fix. Maybe this was the ultimate fix, and something later actually did.

bmatherly:
>> does it still occur if the image background is fully opaque?

> Looking at the images attached here, even the lines inside the opaque areas have very bad aliasing suggesting no interpolation.

Oh yeah. I see what you mean.

src/modules/qt/filter_qtblend.cpp
bmatherly: > This technique goes back to a bug fix in https://github.com/mltframework/mlt/commit/9a19d7b3ccc4af7cb37bb958eefd4f421c51a04d

Thanks for the reference. I suppose the author (or the AI they used) just followed other examples for interp.

> The property getter returns a pointer, and after the call the string it points to can be changed including NULLed, leading to comparison over-reads or crashes.

While that is a true statement, this strdup() trick doesn't fix that because the pointer could still be changed after the get() call and before the strdup call. So this trick only reduces the probability. If data being changed is a concern, then the service should be locked.

I would prefer that we not propagate this unnecessary strdup and start setting a good example for future authors. But I can appreciate that it is following an established convention.

j-b-m: Yes, I just copied code used in the qtblend transition that was initially based on the affine transition where this is [still used](https://github.com/mltframework/mlt/blob/master/src/modules/plus/transition_affine.c#L691). But sure, I will clean that up

j-b-m:
As Dan said, this is unrelated to alpha.. but yes the images attached above were scaled to better show the difference. 
Here is an unscaled comparison, reducing an image (red `S` on a white background) with an original height of 4000 to a height of 270. Using QPainter on the left, QImage on the right.
qpainter-vs-qimage-4000px-to-270px

The quality loss is proportional to the scaling applied in the effect. We probably need to also apply the same trick to the qtblend transition.
I will try to cleanup a bit this MR in the next days.

ddennedy:
FYI I plan to make a release by April 24 if you want to get this in.

ddennedy:
You need to address the comments including the ones from Copilot and provide an overall status (possibly by requesting me to review it). "address" means to give a brief reply and resolve it if resolved. If I see some resolved ones but others unresolved that means not everything is resolved, and this should not be merged.

src/modules/qt/transition_qtblend.cpp
```diff
@@ -266,15 +282,32 @@ static int get_image(mlt_frame a_frame,
                 scale *= transformScale;
             }
         }
-        transform.translate((rect.w - (b_width * scale)) / 2.0, (rect.h - (b_height * scale)) / 2.0);
-        transform.scale(scale, scale);
+        if (scale < 1.) {
+            // Use higher quality QImage scaling
+            scaled = true;
+            const QSize finalSize(qRound(topImg.width() * scale), qRound(topImg.height() * scale));
+            // Center image in rect
```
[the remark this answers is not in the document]
j-b-m: Fixed

src/modules/qt/filter_qtblend.cpp
```diff
@@ -206,9 +206,25 @@ static int filter_get_image(mlt_frame frame,
 
     int image_size = mlt_image_format_size(*format, *width, *height, NULL);
 
+    char *interps = mlt_properties_get(frame_properties, "consumer.rescale");
+    bool hqPainting = strcmp(interps, "nearest") != 0;
```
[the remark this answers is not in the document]
j-b-m: Fixed

j-b-m:
Yes, sorry I was hoping to give it a review just after pushing  my changes, but had to deal with other stuff, should have marked it as draft. Should be ok now, sorry.

src/modules/qt/filter_qtblend.cpp
```diff
@@ -216,9 +216,26 @@ static int filter_get_image(mlt_frame frame,
 
     int image_size = mlt_image_format_size(*format, *width, *height, NULL);
 
+    char *interps = mlt_properties_get(frame_properties, "consumer.rescale");
+    bool hqPainting = interps
+                      && (strcmp(interps, "nearest") == 0 || !strcmp(interps, "neighbor") == 0);
```
ddennedy, line 221: This expression is wrong and inverted logic

```suggestion
    bool hqPainting = interps
                      && strcmp(interps, "nearest") && strcmp(interps, "neighbor");
```
That reads: not nearest **and** not neighbor. IOW, anything but nearest neighbor causes HQ (interpolation).

src/modules/qt/transition_qtblend.cpp
```diff
@@ -248,9 +242,32 @@ static int get_image(mlt_frame a_frame,
...
         *format = mlt_image_rgba;
+
+    bool scaled = false;
+    QImage scaledSource;
+
+    // We don't do subpixel smoothing for nearest neighbour interpolation
+    // so people can use that to upscale pixel art and keep the hard edges.
+    char *interps = mlt_properties_get(properties, "consumer.rescale");
+    bool hqPainting = interps
+                      && (strcmp(interps, "nearest") == 0 || !strcmp(interps, "neighbor") == 0);
```
ddennedy, line 253: Wrong expression

```suggestion
    bool hqPainting = interps
                      && strcmp(interps, "nearest") && strcmp(interps, "neighbor");
```

ddennedy (changes requested):

src/modules/qt/filter_qtblend.cpp
ddennedy: But you removed protection for the string making it worse. If you are going to remove `strdup()` you need to add `mlt_service_lock()` before getting it and `mlt_service_unlock()` after you are done with using it. This must be fixesd.

j-b-m:
Thanks for noticing the interpolation error!

PR #1222, opened 2026-04-19
avformat: set stream.type to "subtitle" for subtitle streams

xevrion:
The switch statement in `producer_avformat.c` that sets `meta.media.X.stream.type` already handles `AVMEDIA_TYPE_VIDEO` and `AVMEDIA_TYPE_AUDIO`, but has no case for `AVMEDIA_TYPE_SUBTITLE`. This means subtitle streams in MKV files (and other containers) leave the `stream.type` property completely unset.

This patch adds the missing case:

```c
case AVMEDIA_TYPE_SUBTITLE:
    mlt_properties_set(meta_media, key, "subtitle");
    break;
```

This makes subtitle stream detection consistent with how video and audio streams are already reported, and allows applications using MLT to reliably detect subtitle streams by checking `meta.media.X.stream.type == "subtitle"` instead of having to fall back to inspecting codec names.

Tested with an MKV file containing an embedded SRT subtitle stream, `meta.media.2.stream.type` is now correctly set to `"subtitle"`.

### DIFF
--- a/src/modules/avformat/producer_avformat.c
+++ b/src/modules/avformat/producer_avformat.c
@@ -477,6 +477,9 @@ static AVRational guess_frame_rate(producer_avformat self, AVStream *stream)
             frame_rate.num = 48000;
             frame_rate.den = 1001;
             break;
+        case AVMEDIA_TYPE_SUBTITLE:
+            mlt_properties_set(meta_media, key, "subtitle");
+            break;
         default:
             break;
         }


src/modules/avformat/producer_avformat.c
```diff
@@ -477,6 +477,9 @@ static AVRational guess_frame_rate(producer_avformat self, AVStream *stream)
             frame_rate.num = 48000;
             frame_rate.den = 1001;
             break;
+        case AVMEDIA_TYPE_SUBTITLE:
+            mlt_properties_set(meta_media, key, "subtitle");
```
ddennedy, line 481: This does not compile. It looks like you (or your AI) meant to change `find_default_streams()` instead of this function. I will apply your change to it and close this. With it `melt -consumer xml video-with-substitles` now shows, for example (good):
```xml
    <property name="meta.media.2.stream.type">subtitle</property>
    <property name="meta.media.2.codec.name">mov_text</property>
    <property name="meta.media.2.codec.long_name">3GPP Timed Text subtitle</property>
    <property name="meta.media.2.codec.bit_rate">41</property>
    <property name="meta.attr.2.stream.language.markup">eng</property>
    <property name="meta.attr.2.stream.handler_name.markup">SubtitleHandler</property>
```
Before it was missing `<property name="meta.media.2.stream.type">subtitle</property>`

Issue #1221, opened 2026-04-16
crash when use timeremap filter


guchangx:
**Describe the bug**
crash when use timeremap filter in ubuntu use latest master code

<img width="2685" height="1334" alt="Image" src="https://github.com/user-attachments/assets/9acf1dc9-cdd8-42d4-8585-a1aa377c425c" />

<img width="2670" height="2018" alt="Image" src="https://github.com/user-attachments/assets/0149d5a5-eaa5-4eee-bf28-2b4aebf61c0b" />

**To Reproduce**
1. edit a mlt file and use mlt execute
2. vscode debug param is "args": [
                "xml:${workspaceFolder}/demo/speed.mlt",
                "-consumer",
                "avformat:${workspaceFolder}/build/speed.mp4"
            ],
3. speed.mlt file will be add in the end

**Expected behavior**
call null pointer 

**Desktop (please complete the following information):**
 - OS: use .devcontainer
 - OS use .devcontainer
- MLT/melt version: build from master, the latest commit is "Fix avformat metadata" 

**Additional context**
i also test same speed.mlt in my win 11, use mlt.exe in shotcut release version 26.2.26, same run param, that is good. can work.
just replace mp4 resource path.

```
<?xml version="1.0" standalone="no"?>
<mlt LC_NUMERIC="C" version="7.37.0" title="Shotcut version 26.2.26">
  <profile description="automatic" width="1920" height="1080" progressive="1" sample_aspect_num="1" sample_aspect_den="1" display_aspect_num="16" display_aspect_den="9" frame_rate_num="30" frame_rate_den="1" colorspace="709"/>
  <chain id="chain0" title="Shotcut version 26.2.26" out="00:00:07.900">
    <property name="length">00:00:07.900</property>
    <property name="eof">pause</property>
    <property name="resource">/workspaces/mlt/media/test1.MP4</property>
    <property name="mlt_service">avformat-novalidate</property>
    <property name="meta.media.nb_streams">1</property>
    <property name="meta.media.0.stream.type">video</property>
    <property name="meta.media.0.stream.frame_rate">30</property>
    <property name="meta.media.0.stream.sample_aspect_ratio">0</property>
    <property name="meta.media.0.codec.width">1920</property>
    <property name="meta.media.0.codec.height">1080</property>
    <property name="meta.media.0.codec.rotate">0</property>
    <property name="meta.media.0.codec.pix_fmt">yuv420p</property>
[...]

PR #1223, opened 2026-04-20
add log for miss chain_normalizers.ini file result in convert_audio is null

guchangx:
for https://github.com/mltframework/mlt/issues/1221#issue-4274955518

The new point log is convenient for beginners.

### DIFF
--- a/src/framework/mlt_chain.c
+++ b/src/framework/mlt_chain.c
@@ -454,7 +454,10 @@ extern void mlt_chain_attach_normalizers(mlt_chain self)
         char temp[PATH_MAX];
         snprintf(temp, PATH_MAX, "%s/chain_normalizers.ini", mlt_environment("MLT_DATA"));
         normalisers = mlt_properties_load(temp);
-        mlt_factory_register_for_clean_up(normalisers, (mlt_destructor) mlt_properties_close);
+        if(normalisers)
+            mlt_factory_register_for_clean_up(normalisers, (mlt_destructor) mlt_properties_close);
+        else
+            mlt_log_error("Failed to load normalizers from %s\n", temp);
     }
 
     // Apply normalisers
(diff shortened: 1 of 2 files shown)


src/framework/mlt_chain.c
```diff
@@ -454,7 +454,10 @@ extern void mlt_chain_attach_normalizers(mlt_chain self)
         char temp[PATH_MAX];
         snprintf(temp, PATH_MAX, "%s/chain_normalizers.ini", mlt_environment("MLT_DATA"));
         normalisers = mlt_properties_load(temp);
-        mlt_factory_register_for_clean_up(normalisers, (mlt_destructor) mlt_properties_close);
+        if(normalisers)
```
ddennedy: bad formatting

```suggestion
        if (normalisers)
```

src/framework/mlt_chain.c
```diff
@@ -454,7 +454,10 @@ extern void mlt_chain_attach_normalizers(mlt_chain self)
...
+            mlt_factory_register_for_clean_up(normalisers, (mlt_destructor) mlt_properties_close);
+        else
+            mlt_log_error("Failed to load normalizers from %s\n", temp);
```
ddennedy: This fails to compile. Why do you not even compile before making a pull request? That is rude.

src/framework/mlt_properties.c
```diff
@@ -243,6 +243,8 @@ static int load_properties(mlt_properties self, const char *filename)
 
         // Close the file
         fclose(file);
+    } else {
+        mlt_log(NULL, MLT_LOG_WARNING, "Failed to open properties file: %s\n", filename);
```
ddennedy: Use `mlt_log_warning()` instead of this form.

ddennedy (changes requested):

src/framework/mlt_chain.c
guchangx: After my observation, i found that PR did not automatically trigger the build. so how to start build manually?

src/framework/mlt_properties.c
```diff
@@ -243,6 +243,8 @@ static int load_properties(mlt_properties self, const char *filename)
 
         // Close the file
         fclose(file);
+    } else {
+        mlt_log_warning(NULL, "Failed to open file %s for yaml parsing!\n", filename);
```
ddennedy: This is not a YAML file. It is a simpler Java style properties format.

```suggestion
        mlt_log_warning(NULL, "Failed to open properties file %s for parsing!\n", filename);
```

ddennedy (changes requested):

PR #1226, opened 2026-04-24
Avoid a potential overflow in sprintf() call

dirkmueller:
1. `default_value1` and `default_value2` are of type `double`.
2. Values are clamped between `FLT_MIN` and `FLT_MAX`.
3. `FLT_MAX` is approximately `3.402823e+38`.
4. Formatting `FLT_MAX` using `%.4f` requires 44 bytes per value.
5. Formatting two such values requires up to 89 bytes.
6. Target buffer `default_value` is explicitly 20 bytes.
7. Writing 89 bytes to a 20-byte buffer causes a stack-based buffer overflow.

Also change a few more sprintf() to snprintfs(). In general compilers will optimize this away if they can determine it is safe to do so, so it is in most cases a noop. except for where it matters :-)


ddennedy (approved):

ddennedy:
thank yu

PR #1232, opened 2026-05-01
xml: resolve relative root against document

Komzpa:
## Summary

Resolve a relative `<mlt root="...">` against the directory of the XML document being parsed, instead of leaving it relative to the process current working directory.

This matters for portable MLT/Kdenlive projects opened from a GUI or another caller whose cwd is unrelated to the project file. For example, if `/tmp/project/output/repro.kdenlive` contains `root="."` and a media resource `../media.mp4`, the resource should resolve under `/tmp/project/output`, not under the caller's cwd.

The patch also reuses the same absolute-path helper for normal resource qualification and adds an avformat XML regression test using `avformat-novalidate`.

## Reproducer

I reproduced this with a Kdenlive/MLT project shaped like this:

- XML file: `/tmp/mlt-root-repro/output/repro.kdenlive`
- XML root: `<mlt root=".">`
- media resources: `../DJI_0884.MP4`, `3.400000:../DJI_0884.MP4`, and `../assets/audio/...`
- command launched from `/tmp`, deliberately not from the project directory

Current packaged MLT 7.38.0 resolves the paths against cwd and logs failures like:

```text
[producer_xml] failed to load producer "./../DJI_0884.MP4"
[producer_xml] failed to load producer "3.400000:./../DJI_0884.MP4"
```

In that repro I got 28 `failed to load producer` lines.

With this patch and the same system modules:

```sh
MLT_REPOSITORY=/tmp/mlt-root-repro/modules \
MLT_DATA=/usr/share/mlt-7 \
build/out/bin/melt /tmp/mlt-root-repro/output/repro.kdenlive \
  -consumer xml:/tmp/mlt-root-repro/new.xml
```

The command exits 0 and the same grep finds 0 `failed to load producer` lines. The resolved resources go through paths such as `/tmp/mlt-root-repro/output/./../DJI_0884.MP4`, which is the document-relative location.

## Validation

- `clang-format -i src/modules/xml/producer_xml.c src/tests/test_mod_avformat/test_mod_avformat.cpp`
- `cmake --build build -j$(nproc)`
- `git diff --check`
- Manual repro above: old packaged MLT fails to load the document-relative media; patched `libmltxml.so` resolves it correctly.

I could not run the new `test_mod_avformat` test locally because this machine lacks the FFmpeg development packages needed to configure `MOD_AVFORMAT=ON` in a fresh test build. The runtime repro used the patched XML module together with the system avformat module.

### DIFF
--- a/src/modules/xml/producer_xml.c
+++ b/src/modules/xml/producer_xml.c
@@ -246,6 +246,27 @@ static inline int is_known_prefix(const char *resource)
     return 0;
 }
 
+static inline int is_absolute_path(const char *path)
+{
+    int drive_letter = strlen(path) > 3 && path[1] == ':' && (path[2] == '/' || path[2] == '\\');
+    return path[0] == '/' || path[0] == '\\' || drive_letter;
+}
+
+static void set_root_property(deserialise_context context, const char *root)
+{
+    const char *document_root = mlt_properties_get(context->producer_map, "root");
+
+    if (root && root[0] && document_root && document_root[0] && !is_absolute_path(root)
+        && !strchr(root, ':')) {
+        char *qualified = malloc(strlen(document_root) + strlen(root) + 2);
+        sprintf(qualified, "%s/%s", document_root, root);
+        mlt_properties_set_string(context->producer_map, "root", qualified);
+        free(qualified);
+    } else {
+        mlt_properties_set_string(context->producer_map, "root", root);
+    }
+}
+
 // Prepend the property value with the document root
 static inline void qualify_property(deserialise_context context,
                                     mlt_properties properties,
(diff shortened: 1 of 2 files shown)


ddennedy:
`root` is not supposed to be a relative path. The docs state that property is used to resolve relative paths within the document if not relative to the document location. Now, you are making it relative to relative, and that is not useful. Maybe you should fix the document generation.

ddennedy:
In other words, it already works relative to the document without a `root`. Does kdenlive require root? Because melt and Shotcut do not.

j-b-m:
Kdenlive will only write absolute path to the xml root property, or omit it when archiving a project, but I don't think there is a way to create a relative path.

ddennedy:
Rejected

PR #1236, opened 2026-05-06
Add libplacebo GPU module with render and shader filters

ddennedy:
supersedes #1201


ddennedy:
@copilot resume review

src/modules/placebo/gpu_context.h
```diff
@@ -0,0 +1,73 @@
...
+pl_renderer placebo_renderer_get(void);
+
+/* Lock/unlock the render mutex. pl_renderer and pl_dispatch are NOT
+   thread-safe, so callers must hold this lock during pl_render_image
+   and pl_dispatch calls. */
+void placebo_render_lock(void);
+void placebo_render_unlock(void);
+
+/* Tear down all GPU resources. Registered via atexit on first init. */
+void placebo_gpu_release(void);
```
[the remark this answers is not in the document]
ddennedy: done

CMakeLists.txt
```diff
@@ -309,6 +311,15 @@ if(MOD_OPENFX)
...
 endif()
 
+if(MOD_PLACEBO)
+  pkg_check_modules(libplacebo IMPORTED_TARGET libplacebo)
+  if(libplacebo_FOUND)
+    list(APPEND MLT_SUPPORTED_COMPONENTS placebo)
+  else()
+    set(MOD_PLACEBO OFF)
+  endif()
+endif()
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/placebo/filter_placebo_shader.c
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /* Lazily allocate private data; destruction is handled by mlt_properties */
+    shader_private *priv = mlt_properties_get_data(filter_props, "_shader_priv", NULL);
+    if (!priv) {
+        priv = calloc(1, sizeof(shader_private));
+        if (!priv)
+            return mlt_frame_get_image(frame, image, format, width, height, writable);
+        mlt_properties_set_data(filter_props, "_shader_priv", priv, 0, shader_private_destroy, NULL);
+    }
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/placebo/filter_placebo_shader.c
```diff
@@ -0,0 +1,471 @@
...
+
+    /* Ensure shader is loaded */
+    if (!ensure_shader(filter, priv, gpu)) {
+        /* No shader available -- pass through */
+        return mlt_frame_get_image(frame, image, format, width, height, writable);
+    }
+
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,508 @@
...
+
+    mlt_log_error(NULL, "[placebo] No GPU backend available\n");
+    return 0;
+
+done:
+    if (s_cache)
+        pl_gpu_set_cache(s_gpu, s_cache);
+
+    load_cache();
+
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,508 @@
...
+{
+    pl_tex tex = (pl_tex) ptr;
+    if (!tex)
+        return;
+    pl_gpu gpu = placebo_gpu_get();
+    if (!gpu)
+        return;
+    placebo_render_lock();
+    pl_tex_destroy(gpu, &tex);
+    placebo_render_unlock();
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,508 @@
...
+pl_gpu placebo_gpu_get(void)
+{
+    LOCK();
+    if (!s_initialized) {
+        s_initialized = 1;
+        if (!init_gpu()) {
+            /* Stay initialized=1 to prevent retry spam on every frame */
+            UNLOCK();
+            return NULL;
+        }
```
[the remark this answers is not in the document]
ddennedy: done

src/modules/core/loader.ini
```diff
@@ -7,7 +7,7 @@
 # the second and third are applied as applicable).
 
 # image filters
-image_convert=movit.convert,avcolor_space,imageconvert
+image_convert=placebo.convert,movit.convert,avcolor_space,imageconvert
```
bmatherly: Using this pattern, will we be able to mix placebo, movit, avfilter, and other modules that use the new private type? What if we change the placebo shader filter_process() to call push_convert()? If we do that, frames with placebo images will have the converter they need without the need for a normalizer. I wonder if there is any need for a placebo normalizer at all (or a normalizer for other future modules that use private type).

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,522 @@
...
+
+int placebo_frame_wants_tex(mlt_frame frame, mlt_image_format format)
+{
+    return format == mlt_image_private
+           && mlt_properties_get_int(MLT_FRAME_PROPERTIES(frame), "_placebo_requested") > 0;
+}
+
+int placebo_frame_is_tex(mlt_frame frame, mlt_image_format format)
+{
+    const char *image_format = mlt_properties_get(MLT_FRAME_PROPERTIES(frame), "mlt_image_format");
```
bmatherly: I would suggest a different key for  this property because "placebo" is, by definition, NOT a valid mlt_image_format. How about "_priv_format_id"? This could be a standard convention that we use for other modules that use the private type.

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,522 @@
...
+    return format == mlt_image_private && image_format && !strcmp(image_format, "placebo");
+}
+
+int placebo_frame_set_tex(mlt_frame frame, pl_tex tex)
+{
+    mlt_properties props = MLT_FRAME_PROPERTIES(frame);
+    mlt_properties_set_int(props, "format", mlt_image_private);
+    mlt_properties_set(props, "mlt_image_format", "placebo");
+    mlt_properties_set_data(props, "alpha", NULL, 0, NULL, NULL);
```
bmatherly: Nit: mlt_properties_clear() might be slightly better since it removes "alpha" from the encapsulated properties map.

src/modules/placebo/gpu_context.c
```diff
@@ -0,0 +1,522 @@
...
+ *   upstream on the same frame;
+ * - results describe the frame's current image payload and must only be set
+ *   when that payload really is a placebo texture.
+ * If these are conflated, one filter can clear another filter's in-flight
+ * request or a non-placebo result can be mistaken for a valid private texture. */
+
+void placebo_frame_set_requested_tex(mlt_frame frame, int requested)
+{
+    mlt_properties props = MLT_FRAME_PROPERTIES(frame);
+    int count = mlt_properties_get_int(props, "_placebo_requested");
```
bmatherly, line 416: Maybe we don't want to support or worry about this hypothetical future idea...

But, in a future world where avfilter also uses mlt_image_private, would this still work if we had a filter stack: placebo->avfilter->placebo?

I think the second placebo might try to interpret the private avframe type as a placebo. Maybe we could use a stack to push/pop private type requests?

bmatherly (commented):
Thanks for this great PR. Here are some comments for your consideration. I did not run the code, so I apologize in advance if my comments are just lacking understanding.

src/modules/core/loader.ini
ddennedy: yes, because the private format is tagged in a property. I used the wrong key. See below

> What if we change the placebo shader filter_process() to call push_convert()?

Push appends, and the functions handling private should come earlier because they often do not handle all the image formats. For example, when placebo converts away from GPU, it can only do rgba, but the request might be different.

src/modules/placebo/gpu_context.c
ddennedy: That was supposed to be "mlt_image_private" as described in https://github.com/mltframework/mlt/pull/1235#issuecomment-4384757760
It is not really necessary to preface frame methods with underscore to hide them from serialization because they don't serialize, and here the value is possibly meaningful across modules.

src/modules/placebo/gpu_context.c
ddennedy: ok

src/modules/placebo/gpu_context.c
ddennedy: It is not a hypothetical. It was needed now to make placebo->brightness(rgba)->placebo, placebo->sepia(yuv422)->placebo, and placebo->movit->placebo work, which is all I have tested so far (in addition to placebo alone and placebo->placebo). If avfilter follows this pattern, it can intermingle.

>  I think the second placebo might try to interpret the private avframe type as a placebo. 

No, that is why it tags the requested mlt_format_private with property "_placebo_requested" and placebo.convert requires for that (`placebo_frame_wants_tex()`). It cannot trust `mlt_image_format==mlt_image_private` alone per my comment in the other PR linked above. The placebo render and shader filters also check property "mlt_image_private"=="placebo" after requesting it through `mlt_frame_get_image()`  to be sure it can continue.

src/modules/placebo/gpu_context.c
ddennedy: changed to "mlt_image_private"

src/modules/core/loader.ini
ddennedy: I added a `mlt_frame_prepend_convert_image()`, called it in placebo.shader's process callback, and removed it from `loader.ini`, and that is working. I will clean that up and push it.

ddennedy:
I am still learning about libplacebo, and the new shader filter is obvious even though [available shaders](https://github.com/mpv-player/mpv/wiki/User-Scripts#user-shaders) are rather limited to a few topics. However, I found the render filter rather odd. What exactly does it do for us? For example, it exposes scaling options but does no scaling. After some brief reading and a discussion with GPT:

---

- The current `placebo.render` filter is best described as a GPU finishing pass, not a general-purpose scaling filter. Its clearest practical value today is debanding, dithering, and improving gradient quality before output, plus staying on the GPU when chained with `placebo.shader`.
- In the current MLT integration, scaling is not meaningfully exercised. Both `placebo.render` and `placebo.shader` render to the incoming frame width and height, so the `upscaler` and `downscaler` settings are effectively dormant unless the filter is placed in a path where the requested frame size actually differs.
- `placebo.shader` does not change the frame resolution in a way that would cause `placebo.render` to perform consumer-resolution scaling later. It also preserves the current frame dimensions.
- Debanding and dithering are real, working features in this wrapper today. Those are the strongest reasons to use the filter in practice.
- The current tone-mapping / color-mapping surface is much weaker than the deband/dither surface. Although `pl_color_map_params` and tone-mapping selection are wired, both source and destination frames are currently declared as sRGB, so the filter is not expressing a robust source-to-target color transform. In other words, the tone-mapping control exists, but the surrounding color-state information needed to make it fully meaningful is not really there yet.
- Because of that, the metadata and user-facing description should emphasize “GPU finishing pass”, “debanding”, “dithering”, and “cleaner gradients before output”, rather than leading with scaling.
- On future scope: if the goal is one simple finishing filter, debanding and dithering should stay in `placebo.render`. The most valuable next steps would be richer deband/dither controls and, separately, a proper color/HDR filter once source/target color metadata is propagated correctly.

---

### A couple of my thoughts

Maybe this module should provide new normalizing filters. It can do color (de)linearization, image (pixel format and colorspace) conversion, deinterlace, and scaling. It is a little disconcerting that these new filters are hard-wired around 8-bit rgba.

---

**Pixel formats**
Libplacebo is not really built around one fixed FFmpeg-style “supported format list.” It is built around frames, planes, and GPU texture formats. In practice that means it can work with:

- RGB and YUV/YCbCr sources
- packed and planar layouts
- single-plane and multi-plane images
- chroma-subsampled formats like 4:4:4, 4:2:2, and 4:2:0
- alpha
- higher bit depths, not just 8-bit
- integer and float textures

So for planning, `8-bit RGBA`, `10-bit YUV420`, `16-bit RGB`, and HDR-oriented formats are all in scope conceptually.

The caveat is important: exact ingest/output depends on the upload path and the GPU/backend texture support. Libplacebo’s model is capability-based, not “these exact 27 formats always work everywhere.”

**Color spaces / color metadata**
Libplacebo is also designed to understand and transform color metadata, including:

- RGB vs YCbCr representations
- color matrices
- transfer characteristics
- color primaries
- full vs limited range
- SDR and HDR workflows
- gamut mapping
- tone mapping
- linear-light processing where appropriate

So yes, converting to and from linear light is within libplacebo’s capability model. It also supports HDR-related workflows like PQ and HLG, and source/target color transforms rather than just “treat everything as sRGB.”

A practical mental model is:

- Input can be something like `10-bit 4:2:0 BT.2020 PQ`
- Libplacebo can interpret that correctly
- It can linearize internally when needed for scaling/color operations
- It can tone-map / gamut-map to another target
- Output can be something like `BT.709 SDR`, higher-bit-depth RGB, or another supported target frame layout

**What libplacebo is good at**
If you are planning a future integration, libplacebo is well-suited for:

- high-bit-depth RGB workflows
- YUV input without flattening everything early
- HDR to SDR and HDR to HDR conversions
- linear-light aware scaling / filtering
- preserving color metadata instead of hardcoding sRGB
- GPU-native processing with debanding, dithering, gamut mapping, tone mapping, and scaling in one pipeline

**What not to assume**
Do not assume:

- every FFmpeg pixel format maps 1:1 with no adaptation
- every backend supports every texture format equally
- an integration automatically preserves YUV/high-bit-depth/linear support just because libplacebo can

That last point is where wrappers often narrow the capability.

The shortest accurate summary is:

- Yes, libplacebo can handle linear-light aware color conversion.
- Yes, it can work with higher-bit-depth RGB.
- Yes, it can work with YUV pixel formats, including planar/chroma-subsampled video formats.
- No, there is not one tiny fixed “supported pixel format” list independent of backend and upload path.

---

OK, clearly this PR is just a start, and shader is clearly general purpose and extensible. However, now I think this render filter should be refactored into 2 new filters with no scaling or tonemapping options: deband and dither.

PR #1239, opened 2026-05-08
Remove dependency on libavdevice

dimula73:
It seems like the actual usage of avdevice has been removed long ago and replaced with SDL.


dimula73:
I'm not very sure my understanding is right, but I haven't found any usage of avdevice by MLT internally. I have a feeling that this dependency is a legacy artifact from ~2011.

In Krita we have been building `ffmpeg` without `avdevice` and MLT worked fine until the recent dependency policy change in https://github.com/mltframework/mlt/commit/604972b255c53927082de3989237f857b7827b74

ddennedy:
MLT can use device input through the avformat producer.

ddennedy (changes requested):
If you really want something like this, change the build to make it optional instead of removing it. For example, Shotcut supports these input devices through MLT: alsa, pulseaudio, v4l2, dshow, avfoundation.

dimula73:
> Shotcut supports these input devices through MLT: alsa, pulseaudio, v4l2, dshow, avfoundation.

Isn't it a responsibility of the caller to create (and, hence, initialize) avdevice then? I haven't found any other avdevice-related code in MLT.

> If you really want something like this, change the build to make it optional instead

I could do that. Is it okay that it would technically revert a part of https://github.com/mltframework/mlt/commit/604972b255c53927082de3989237f857b7827b74 ?

ddennedy:
> Isn't it a responsibility of the caller to create (and, hence, initialize) avdevice then?

no

> Is it okay that it would technically revert a part 

Yes, that is basically what I said or maintain this patch.

dimula73:
Okay, I'll modify the patch, thanks!

dimula73:
This MR can now be closed in favor of the new one, that makes avdevice optional: https://github.com/mltframework/mlt/pull/1246

PR #1240, opened 2026-05-09
Add audio and image format fields to metadata

bmatherly:
Add image_formats and audio_formats to the filter metadata so that applications can look up what formats are supported by the service. This can be used in Shotcut to map image formats to keywords rgba, yuv, and 10bit.

### DIFF
--- a/src/modules/avformat/filter_info.yml
+++ b/src/modules/avformat/filter_info.yml
@@ -1,0 +1,2180 @@
+# FFmpeg filter supported MLT formats
+# Filters without known formats are placeholders encoded as "" (not empty maps).
+# Filters marked as 'metadata only' process metadata only, passing data unchanged.
+# Filters marked as 'format agnostic' accept all formats dynamically with no static restrictions.
+# Filters marked as 'hardware only' accept only hardware/GPU formats not directly supported by MLT.
+
+aap:
+  audio_formats:
+    - f32le
+abench: ""  # metadata only
+acompressor:
+  audio_formats:
+    - f32le
+acontrast:
+  audio_formats:
+    - f32le
+acopy: ""  # metadata only
+acrossfade:
+  audio_formats:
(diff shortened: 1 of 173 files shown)


src/modules/avformat/filter_info.yml
```diff
@@ -0,0 +1,2180 @@
+# FFmpeg filter supported MLT formats
```
bmatherly, line 1: Even though FFMpeg has a rich framework for a filter to describe the formats it supports, the public API does not provide a way for the application to query those formats. They are only used internally for negotiating the connections in a filtergraph.

I used AI to help me write a script to inspect the FFMpeg source code and extract these format lists. I suppose we will maintain it manually over time. Or maybe we should keep the script around so we can re-run it from time to time.

This is the one thing that gives me pause about this whole PR. Maybe it is not worth the long-term maintenance.

src/modules/avformat/filter_info.yml
```diff
@@ -0,0 +1,2180 @@
...
+# Filters without known formats are placeholders encoded as "" (not empty maps).
```
bmatherly, line 2: Maybe this file could be combined with blacklist.txt, resolution_scale.yml, and yuv_only.txt. I am willing to merge those in a separate PR if we are interested in that.

ddennedy: I do not that is very useful.

src/modules/avformat/filter_info.yml
ddennedy: Please add the script to src/modules/avformat

bmatherly: Done

src/modules/avformat/filter_info.yml
bmatherly: 👍

PR #1246, opened 2026-05-11
Make avdevice library optional again

dimula73:
It is not used by avformat when performing playback via SDL2, the only thing it does is initializing avdevice and not using it later.

This is a follow-up of the previous rejected MR: https://github.com/mltframework/mlt/pull/1239


ddennedy (approved):
working for me with avdevice included 👍

jlskuz:
Making the dependency optional is in principle fine for me, but I would prefer to have it explicitly managed with a `WITH_AVDEVICE` switch. I know we still have this implicit behavior in other places (like for ladaspa.h) still, but my goal would be to have them all controlled explicitly in the future. See #1115 for more context.

src/modules/avformat/CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ if(WIN32)
   endif()
 endif()
 
+if(AVDEVICE_FOUND)
+  target_compile_definitions(mltavformat PRIVATE AVDEVICE)
```
jlskuz:
```suggestion
  target_compile_definitions(mltavformat PRIVATE USE_AVDEVICE)
```

src/modules/avformat/factory.c
```diff
@@ -45,7 +45,9 @@ extern mlt_link link_swresample_init(mlt_profile profile, mlt_service_type, cons
 // ffmpeg Header files
 #include "mltavformat_export.h"
 #include <libavcodec/avcodec.h>
+#ifdef AVDEVICE
```
jlskuz:
```suggestion
#ifdef USE_AVDEVICE
```

src/modules/avformat/factory.c
```diff
@@ -58,7 +60,9 @@ static void avformat_init()
     // Initialise avformat if necessary
     if (avformat_initialised == 0) {
         avformat_initialised = 1;
+#ifdef AVDEVICE
```
jlskuz:
```suggestion
#ifdef USE_AVDEVICE
```

jlskuz (changes requested):

dimula73:
Hi, @jlskuz!

I have applied your `USE_AVDEVICE` suggestions. Does this patch need any other changes before it could be merged?

ddennedy:
@jlskuz prefers that options be explicit instead of discovered by what is available. For example, for the `jackrack` module that originally provided only ladspa we added in the top level `CMakeLists.txt`
```
option(USE_LV2 "Enable LV2 features" ON)  
option(USE_VST2 "Enable LV2 features" ON)
```
I think the same is desired for avformat:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 36c8d82f..ea7745f3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ option(CLANG_FORMAT "Enable Clang Format" ON)
 option(BUILD_TESTS_WITH_QT6 "Build test against Qt 6" ON)

 option(MOD_AVFORMAT "Enable avformat module" ON)
+option(USE_AVDEVICE "Integrate FFmpeg libavdevice for capture" ON)
 option(MOD_DECKLINK "Enable DeckLink module" ON)
 option(MOD_FREI0R "Enable Frei0r module" ON)
 option(MOD_GDK "Enable GDK module" ON)
diff --git a/src/modules/avformat/CMakeLists.txt b/src/modules/avformat/CMakeLists.txt
index aebb4637..3b197daa 100644
--- a/src/modules/avformat/CMakeLists.txt
+++ b/src/modules/avformat/CMakeLists.txt
@@ -51,7 +51,7 @@ if(WIN32)
   endif()
 endif()

-if(AVDEVICE_FOUND)
+if(USE_AVDEVICE AND AVDEVICE_FOUND)
   target_compile_definitions(mltavformat PRIVATE USE_AVDEVICE)
 endif()
````
Beyond that, I tested this branch again, and what you have still works for me.

dimula73:
Hi, @ddennedy!

I have a question about your proposed patch. When `USE_AVDEVICE` is `ON`, should the build fail when avdevice is not found?

dimula73:
Okay, I've implemented an option. The build will fail if the option is ON, but avdevice is missing.

PR #1251, opened 2026-05-27
openfx: honor OFX bottom-left row order with negative row bytes

bmatherly:
Set kOfxImagePropRowBytes negative and shift kOfxImagePropData to the last scanline for Source and Output clips so CPU images are presented to OpenFX in lower-left origin order without repacking.

An easy way to test this is to apply the AboutGMIC plugin and see if the image is upside down

All the openfx plugins were operating on the image upside down because openfx wants the first pixel to be the bottom left. The API allows negative row-bytes, so this patch allows us to provide the correct orientation without repacking the frame. 

I was feeling confident about this patch with my testing and then I ran into a crash. Specifically openfx.net.fxarena.openfx.Charcoal crashes with this patch. When I  test that same plugin without this patch, it does not crash, but it also doesn't work (black frame). Maybe something else is wrong with that plugin that this change is exposing. So I'm posting this patch here to allow for more testing.

### DIFF
--- a/src/modules/openfx/mlt_openfx.c
+++ b/src/modules/openfx/mlt_openfx.c
@@ -247,9 +247,16 @@ static OfxStatus clipGetRegionOfDefinition(OfxImageClipHandle clip, OfxTime time
     propGetInt((OfxPropertySetHandle) clip_prop, kOfxImagePropRegionOfDefinition, 1, &y1);
     propGetInt((OfxPropertySetHandle) clip_prop, kOfxImagePropRegionOfDefinition, 2, &x2);
     propGetInt((OfxPropertySetHandle) clip_prop, kOfxImagePropRegionOfDefinition, 3, &y2);
-    bounds->x1 = (double) x1;
+    // OFX spec requires canonical coordinates: x scaled by PAR, y in pixels.
+    double par = 1.0;
+    mlt_properties par_props = mlt_properties_get_properties(clip_prop, "0");
+    if (par_props && mlt_properties_exists(par_props, kOfxImagePropPixelAspectRatio))
+        par = mlt_properties_get_double(par_props, kOfxImagePropPixelAspectRatio);
+    if (par <= 0.0)
+        par = 1.0;
+    bounds->x1 = (double) x1 * par;
     bounds->y1 = (double) y1;
-    bounds->x2 = (double) x2;
+    bounds->x2 = (double) x2 * par;
     bounds->y2 = (double) y2;
 
     if (bounds->x2 < bounds->x1 || bounds->y2 < bounds->y1) {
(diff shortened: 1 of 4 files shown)


ddennedy:
That is a good find. I have tended to avoid gmic because they are so slow. I used  `openfx.net.sf.openfx.CropPlugin` quite a bit because it supports slices threading and I found other problems with it. FYI, this is one that is using pairs of numbers as widget: point and size. It was never upside-down, but now it crops from bottom-left corner. However, I just noticed it has a `reformat` param with a description "Translates the bottom left corner of the crop rectangle to be in (0,0)." It does not do anything, neither before or after this branch. But it seems that is related to this. With NTSC-rs, I see this correctly places the bottom rows of distortion (Head Switching & Tracking Noise) along the bottom instead of the top better matching examples I have seen. 

BTW, if you try out Crop I suggest to turn on "Black Outside"; otherwise, it only affects the alpha channel.

ddennedy:
I found another related problem not fixed by this. The GMIC Charcoal is working for me both on master and this branch, but both appear upside-down! On this branch, About GMIC renders only what it draws/affects correctly, and usually it draws over the entire frame but not always. Where it does not affect the entire frame is upside-down:
 
Screenshot_2026-05-26_22-33-07

I only see this with GMIC plugins, and many plugins are OK (e.g. CropOFX and MirrorOFX) but others make everything black (*CImg). It seems there still might be some fundamental things incorrect.

src/modules/openfx/mlt_openfx.c
```diff
@@ -2311,7 +2311,16 @@ void mltofx_set_source_clip_data(OfxPlugin *plugin,
...
                   depth_format);
 
-    propSetInt((OfxPropertySetHandle) clip_prop, kOfxImagePropRowBytes, 0, width * depth_byte_size);
+    int row_bytes = width * depth_byte_size;
+    uint8_t *image_origin = image;
+    if (row_bytes > 0 && height > 0) {
+        // OFX CPU images are addressed from lower-left. Point data at the
+        // first byte of the last scanline and use negative row bytes.
+        image_origin = image + ((size_t) (height - 1) * (size_t) row_bytes);
+        row_bytes = -row_bytes;
```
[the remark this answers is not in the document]
ddennedy: I have a local commit that adds a `mlt_origin` property for this purpose and fixes a few other little things with respect to non-square pixels that I found along the way of troubleshooting. Nothing has solved the core issues of flipped image with GMIC and black with CImg. I even tried a change that removes the negative row bytes, and flips the image between MLT and OFX plugin--input and output.

I think this change does help, in general, and should be merged. Then, one can add a vertical flip filter manually for now as needed.

I even checked out the Natron source and asked Claude about it:

> What special hacks or workarounds exists for certain openfx plugins?

In summary, there is A LOT, especially for various commercial plugins, but nothing it raised about image flipping.

src/modules/openfx/mlt_openfx.c
```diff
@@ -2311,7 +2311,16 @@ void mltofx_set_source_clip_data(OfxPlugin *plugin,
                                                   : (format == mlt_image_rgb ? "rgb" : "Unknown")),
                   depth_format);
 
-    propSetInt((OfxPropertySetHandle) clip_prop, kOfxImagePropRowBytes, 0, width * depth_byte_size);
+    int row_bytes = width * depth_byte_size;
+    uint8_t *image_origin = image;
+    if (row_bytes > 0 && height > 0) {
```
ddennedy: Here is a workaround for G'MIC that is working for me

```suggestion
    int gmic = !strncmp(plugin->pluginIdentifier, "eu.gmic.", 8);
    if (!gmic && !row_bytes > 0 && height > 0) {
```

bmatherly: I tried your workaround and I did not see any problem with it (other than it is unfortunate to have a workaround). Do you want to push what you have? Otherwise, I can update this branch and merge it.

ddennedy (approved):

PR #1253, opened 2026-06-05
OpenFX fixes for CImg

ddennedy:
Changes I made to get the CImg-based OpenFX plugins working.

### DIFF
--- a/src/modules/openfx/mlt_openfx.c
+++ b/src/modules/openfx/mlt_openfx.c
@@ -790,7 +790,7 @@ static OfxStatus paramGetValueImpl(OfxParamHandle paramHandle, va_list ap)
         if (status != kOfxStatOK) {
             status = propGetInt((OfxPropertySetHandle) param_props, "OfxParamPropDefault", 0, value);
             if (status != kOfxStatOK)
-                return kOfxStatErrUnknown;
+                *value = 0;
         }
     } else if (strcmp(param_type, kOfxParamTypeChoice) == 0) {
         // MltOfxParamValue stores the string label; convert to integer index for the OFX plugin.
@@ -820,12 +820,12 @@ static OfxStatus paramGetValueImpl(OfxParamHandle paramHandle, va_list ap)
                                     0,
                                     value);
                 if (status != kOfxStatOK)
-                    return kOfxStatErrUnknown;
+                    *value = 0;
             }
         } else {
             status = propGetInt((OfxPropertySetHandle) param_props, "OfxParamPropDefault", 0, value);
             if (status != kOfxStatOK)
-                return kOfxStatErrUnknown;
+                *value = 0;
         }
     } else if (strcmp(param_type, kOfxParamTypeDouble) == 0) {
         double *value = va_arg(ap, double *);
@@ -837,7 +837,7 @@ static OfxStatus paramGetValueImpl(OfxParamHandle paramHandle, va_list ap)
                                    0,
                                    value);
             if (status != kOfxStatOK)
-                return kOfxStatErrUnknown;
+                *value = 0.0;
         }
     } else if (strcmp(param_type, kOfxParamTypeString) == 0
                || strcmp(param_type, kOfxParamTypeStrChoice) == 0) {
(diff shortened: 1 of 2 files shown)


src/modules/openfx/mlt_openfx.c
```diff
@@ -780,7 +876,7 @@ static OfxStatus paramGetValueImpl(OfxParamHandle paramHandle, va_list ap)
         if (status != kOfxStatOK) {
             status = propGetInt((OfxPropertySetHandle) param_props, "OfxParamPropDefault", 0, value);
             if (status != kOfxStatOK)
-                return kOfxStatErrUnknown;
+                *value = 0;
```
bmatherly, line 879: This one and the similar ones have me a little worried. It seems wrong to confidently return a value when clearly the parameter is not defined. But the documentation for paramGetValue() and similar functions only list Bad Handle the an error response. Error Unknown seems right intuitively. But the documentation is not clear when it should be used and I do not know in practice what plugins are expecting.

I think the most important thing would be that we not return a NULL pointer type when we should be returning an error. But there is not a pointer param  (only Props). So that should be OK for this. Let's try it an see what happens.

bmatherly:
I don't know if this comment applies to this review. But I was thinking that if we submit bug fixes for specific plugins, maybe we should list the plugin in the commit message so that a future git blame will give us a clue about how to re-test the  change.

src/modules/openfx/mlt_openfx.c
ddennedy: Is there a reason you resolved this thread?

I did not like the change proposed by AI at first either, but this is one of the things it suggested when checking the debug log. I double-checked and sure enough. When questioned further, it suggested that an error returned from here that a plugin does not properly handle is more of a general compatibility concern than a 0 value, which makes sense. We currently still have a lot of compatibility issues; I do not know if you are aware of that. So, at this point we need to be as proactive (e.g. provide more project and clip info already done for the most part) and least negative as possible.  

I do not know what you mean by "not return a NULL pointer." `OfxStatus` is an int, and the output value pointer is provided as an argument. 

What I can tell you is that this is required to get the CImg plugins to do anything but change the video to black--all of them. And of the open source OpenFX plugins these are some of the actually most useful ones! You have not tested them? I have another separate plugin, GoPro 360 Reframe, that also makes the video black, I was hoping this would fix as well,  but it did not.

> I was thinking that if we submit bug fixes for specific plugins, maybe we should list the plugin in the commit message so that a future git blame.

I agree, which is why I put "CImg" in the commit message. This set of changes applies to all of them (net.sf.cimg.*)

bmatherly: > Is there a reason you resolved this thread?

Just to indicate that I am not requesting any changes.

> I do not know what you mean by "not return a NULL pointer."

In propGetPointer() we return an error instead of returning 0 (which the plugin would interpret as a NULL pointer). I think that change resolved a crash and should be kept.

> I agree, which is why I put "CImg" in the commit message. This set of changes applies to all of them (net.sf.cimg.*)

Good progress. I propose that we merge this change before the release.

ddennedy: > returning 0 (which the plugin would interpret as a NULL pointer)

0 is a numeric or color value, not a pointer address, and would not be interpreted as NULL. The only output param currently that could be NULL is string.

> I think that change resolved a crash and should be kept.

Which change? I have not experienced any crashing on OpenFX plugins for a few weeks, and not on this branch either.

bmatherly: Sorry. My comment is confusing. I was referring to a change we made in propGetPointer() long ago which is not in any way related to this PR. I have not experienced crashes recently, either. I was just expressing concern that returning a zero could cause a crash if it were interpreted as a pointer  - which I do not  believe they are in this change. So it should be good.

ddennedy: Well, I have one thing to resolve related to this particular commit you called out. It made the openfx.net.sf.openfx.Clamp and another others in the same namespace invert the color! What the heck?! See what I mean by the "twilight zone?" I only found the link when testing this today with the commit reverted to triple check my claims. Previously, I spent a lot of time all over the place looking for this color inversion.

bmatherly: > Well, I have one thing to resolve ...

"one more thing"... HA! It really is a rabbit hole. I have been going one plugin at a time looking for things to improve. I have been accumulating some changes which I will offer up later.

ddennedy: > invert the color

This was due to a bug in the plugin, and I was using an old build provided by Ubuntu package. building openfx-misc from source fixed it after spending a lot of tokens, which are billed for usage now, sigh

PR #1254, opened 2026-06-08
Gmic improvements

bmatherly:


src/modules/openfx/openfx/include/nuke/fnPublicOfxExtensions.h
```diff
@@ -0,0 +1,143 @@
+//------------------------------------------------------------------------------
```
bmatherly, line 1: GMIC and Natron are using this extension - as well as a few other plugins I came across when searching for these symbols on GitHub.

src/framework/mlt_property.c
```diff
@@ -2106,18 +2107,94 @@ mlt_properties mlt_property_get_properties(mlt_property self)
...
+ *
  * \public \memberof mlt_property_s
  * \param self a property
  * \return true if the property is animated
  */
 
 int mlt_property_is_anim(mlt_property self)
 {
-    return self->animation || (self->prop_string && strchr(self->prop_string, '='));
+    if (self->animation)
```
bmatherly: One of the AboutGMIC notes has a url with an equals sign that was tripping over this and being corrupted. This check is more robust.

ddennedy: This animation change might block merging this before the next Shotcut release. While we are not doing string animation anywhere yet, I think this also affects any property currently with a string representation included. Is this change intended to be covered by existing tests or should another be added?

ddennedy:
Besides a UI improvement is there some difference in image effect with this?

src/framework/mlt_property.c
bmatherly: Good point. I will peel that commit into a separate PR and add a dedicated test.

bmatherly:
I removed the animation related changes.

ddennedy (approved):

bmatherly:
> Besides a UI improvement is there some difference in image effect with this?

No. It just improves the UI controls/widgets.

PR #1256, opened 2026-06-10
[win] Don't build a custom nanosleep() when not needed

dimula73:
Instead of relying of the includes order we should properly
check if the function is actually provided by our version
of pthread library.

The patch basically reverts an original hack to resolve the
issue in msys2 and implements a proper detection mechanism
for the presence of nanosleep() in the pthread library


dimula73:
There is still an unorthodox usage of `WIN_PTHREADS_TIME_H` in `mlt_types.h`, which dates back from 2015. Theoretically, it could also be replaced with `HAVE_PTHREAD_NANOSLEEP`, though we would then have to add a public-interface definition for libmlt, like that:

```
target_compile_definitions(mlt PUBLIC "HAVE_PTHREAD_NANOSLEEP=${HAVE_PTHREAD_NANOSLEEP}")
```

Though I'm not sure if such a change is acceptable or not

ddennedy:
> unorthodox usage of WIN_PTHREADS_TIME_H in mlt_types.h

It used to exist even if not in a standard. Google search on it  first hit leads me to;
https://github.com/tonytheodore/winpthreads/blob/master/include/pthread_time.h

Since it is unorthodox it is OK to change it IMO.

ddennedy (approved):

ddennedy:
I approve the change but wait to see what you do about `mlt_types.h`

dimula73:
Hi, @ddennedy!

I did a small test here and it seems like it is not easy to add `HAVE_PTHREAD_NANOSLEEP` into a global definitions interface of the entire library (.pc and Mlt7Config.cmake files will have to be adjusted for that manually). 

So, we have three options here:

1) Merge this patch as it is and ignore the fact that the implementation and prototype for MLT's `nanosleep()` are guarded by different definitions.

2) Revert `nanosleep()`'s implementation to be guarded by the undocumented `WIN_PTHREADS_TIME_H` and just include `pthread_time.h` into the translation unit as well (this version will be the nearest to the original implementation)

3) Add a global mlt_config.h file, which would be generated by CMake and provide this macro for mlt_types.h.

ddennedy:
`nanosleep()` for win32 is only supposed to be used internally but was added to `mlt_types.h` for convenience. We can add a `src/win32/mlt_win32.h`, but then it needs to be conditionally included wherever `nanosleep` is called. Or, your option 2 if `__MINGW32__`

src/framework/mlt_types.h
```diff
@@ -22,6 +22,10 @@
 #ifndef MLT_TYPES_H
 #define MLT_TYPES_H
 
+#ifndef GCC_VERSION
+#define GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+#endif
+
```
ddennedy: This PR has too many unnecessary changes:
- reverting a copyright year
- adding this GCC_VERSION macro that is not used

src/win32/win32.c
```diff
@@ -46,16 +46,16 @@ int usleep(unsigned int useconds)
...
 {
 	if (rqtp->tv_nsec > 999999999) {
 		/* The time interval specified 1,000,000 or more microseconds. */
 		errno = EINVAL;
 		return -1;
 	}
-    return usleep(rqtp->tv_sec * 1000000 + rqtp->tv_nsec / 1000);
-}
+	return usleep( rqtp->tv_sec * 1000000 + rqtp->tv_nsec / 1000 );
+}
```
ddennedy, line 58: Do not change white space: spaces to tabs, and extra spaces added inside parantheses

CMakeLists.txt
```diff
@@ -207,9 +207,6 @@ if(WIN32)
     find_package(dlfcn-win32 REQUIRED)
     set(CMAKE_DL_LIBS dlfcn-win32::dl)
   endif()
-  if(MINGW)
-    string(REPLACE "iconv" "pthread" MLT_PTHREAD_LIBS "${Iconv_LIBRARY}")
```
ddennedy: This was added for a reason not too long ago. See git blame. Explain why it is no longer needed.

ddennedy (changes requested):

ddennedy:
Please address the comments.

dimula73:
Hi, @ddennedy!

I plan to return to this patch some time soon, I was just busy with other things (and then the vacation came :) )

PR #1257, opened 2026-06-11
Cropofx

bmatherly:
Map all 2D and 3D types to MLT rect

### DIFF
--- a/src/modules/openfx/mlt_openfx.c
+++ b/src/modules/openfx/mlt_openfx.c
@@ -3230,15 +3230,17 @@ void *mltofx_fetch_params(OfxPlugin *plugin, mlt_properties params, mlt_properti
                     if (!isnormal(y) || y <= FLT_MIN || y >= FLT_MAX)
                         y = 0.0;
 
-                    mlt_rect r = {x, y, 0, 0, 1};
-                    mlt_properties_set_rect(p, "default", r);
+                    char rect_str[100];
+                    snprintf(rect_str, sizeof(rect_str), "%g %g", x, y);
+                    mlt_properties_set(p, "default", rect_str);
 
                 } else if (strcmp(param_type, kOfxParamTypeInteger2D) == 0) {
                     int ix = 0, iy = 0;
                     propGetInt((OfxPropertySetHandle) ppp, p_name, 0, &ix);
                     propGetInt((OfxPropertySetHandle) ppp, p_name, 1, &iy);
-                    mlt_rect r = {ix, iy, 0, 0, 1};
-                    mlt_properties_set_rect(p, "default", r);
+                    char rect_str[100];
+                    snprintf(rect_str, sizeof(rect_str), "%d %d", ix, iy);
+                    mlt_properties_set(p, "default", rect_str);
 
                 } else if (strcmp(param_type, kOfxParamTypeDouble3D) == 0) {
                     double x3 = 0.0, y3 = 0.0, z3 = 0.0;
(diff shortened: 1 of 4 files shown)


src/framework/mlt_property.c
```diff
@@ -1889,7 +1897,7 @@ int mlt_property_set_rect(mlt_property self, mlt_rect value)
 {
     pthread_mutex_lock(&self->mutex);
     clear_property(self);
-    self->types = mlt_prop_rect | mlt_prop_data;
+    self->types = mlt_prop_rect;
     self->length = sizeof(value);
     self->data = calloc(1, self->length);
```
ddennedy, line 1894: I think it was like this because mlt_rect is using the data member. I am a little nervous about the changes to `mlt_property.c`, but I want to release a new MLT with the next Shotcut, and I want to include this change to reach some stabilization for the OpenFX parameters interface. I guess it is OK to separate data and rect. and it is strictly necessary for the OpenFX change?

bmatherly: The crash occurs when the user clicks on the help button (?). The help panel gets the service metadata and then traverses it to build an HTML page. When traversing metadata, if a metadata is a data field, it assumes the data is a child properties item and interprets it as a properties object (this was all implemented before we had the formal nested properties functions). So the pointer is populated with a rect, but interpreted as a properties and then it crashes. Maybe we can find a different way to work around that?

src/framework/mlt_property.c
```diff
@@ -904,7 +910,9 @@ void *mlt_property_get_data(mlt_property self, int *length)
 
     // Return the data (note: there is no conversion here)
```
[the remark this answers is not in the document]
bmatherly: This is a good comment. Also, the access of self->length should be inside the mutex.

src/modules/openfx/mlt_openfx.c
```diff
@@ -2370,6 +2419,14 @@ void mltofx_init_host_properties(OfxPropertySetHandle host_properties)
...
     propSetInt(host_properties, kOfxParamHostPropPageRowColumnCount, 1, 0);
     propSetInt(host_properties, kOfxParamHostPropSupportsParametricAnimation, 0, 0);
+    // Declare support for normalised default coordinate systems on Double2D/Double params.
+    // When present, plugins call setDefaultCoordinateSystem(eCoordinatesNormalised) on param
+    // descriptors whose defaults are expressed in normalized [0,1] coordinates, which our
+    // metadata loop then reads as normalized_coordinates="yes".
+    propSetString(host_properties,
+                  kOfxParamPropDefaultCoordinateSystem,
+                  0,
+                  kOfxParamCoordinatesNormalised);
```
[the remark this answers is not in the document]
bmatherly: This probably needs more investigation and testing. But the suggestion is probably good. I was using AI to try to find workarounds for the default normalization problem and this is one suggestion it came up with after inspecting Natron and plugin source code. But it is probably unnecessary.

Note this Natron documentation which explicitly lists the Crop plugin as not working with Davinci.
https://github.com/NatronGitHub/openfx-misc?utm_source=chatgpt.com#the-default-parameters-are-too-small-on-davinci-resolve

src/framework/mlt_property.c
ddennedy: Oh, I did not realize it was fixing a crash. That explanation makes much sense. Well, I’m thinking to merge these latest and make another beta this weekend.

bmatherly: I worked around this. I set the rect values as strings.

I created a separate PR for the "rect is not data" change which we can debate after the release.
https://github.com/mltframework/mlt/pull/1258

src/modules/openfx/mlt_openfx.c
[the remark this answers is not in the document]
bmatherly: I applied this suggestion. I understand the default coordinate system parameter better, implemented it, and added a workaround for the crop plugin which does not report it correctly.

bmatherly:
I think it would be safe to merge this and the Shotcut PR before the test release.

PR #1269, opened 2026-07-09
Ladspa deduplicate

bmatherly:
I stumbled across this while investigating something else. The ladspa.* filters are registered twice. The second registration overwrites the first. In the current framework, if a duplicate service is registered, it overwrites the first. This change adds a warning for that situation and avoids the duplicate registration for ladspa. 

There are also duplicate registrations in openfx that will be handled in a different PR.


src/framework/mlt_repository.c
```diff
@@ -230,52 +230,51 @@ void mlt_repository_register(mlt_repository self,
...
+    }
+
+    if (mlt_properties_get_data(service_list, service, NULL)) {
+        mlt_log_error(NULL,
+                      "%s: Duplicate %s service registration for \"%s\" (overwriting previous "
+                      "registration)\n",
+                      __FUNCTION__,
+                      service_type_name,
+                      service);
     }
```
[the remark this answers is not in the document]
bmatherly: I think this message would indicate a programming error. So I think error level is good and the developer would hopefully notice it right away.

src/framework/mlt_repository.c
```diff
@@ -230,52 +230,51 @@ void mlt_repository_register(mlt_repository self,
...
     default:
         mlt_log_error(NULL, "%s: Unable to register \"%s\"\n", __FUNCTION__, service);
-        break;
+        return;
+    }
+
+    if (mlt_properties_get_data(service_list, service, NULL)) {
+        mlt_log_error(NULL,
+                      "%s: Duplicate %s service registration for \"%s\" (overwriting previous "
+                      "registration)\n",
```
ddennedy: I think this message is too long. It is nicer when it fits on a single line of common terminal window sizes. Typically, that meant 80 chars, but we often use wider now due to the prevalence of window systems now. Still, the "(overwriting previous...)" part is superfluous.

ddennedy:
Duplication is intentional;-see f50567ba0294cbf05fc4fab7d2872214644e5e64
Does this change break that? What problem does redundant registration cause? I agree to change mlt_repository to ignore a duplicate registration but not log it at error level, only perhaps debug level.

bmatherly:
> Duplication is intentional. Does this change break that?

The current behavior is that mltladspa and mltjackrack library files both register all of the LADSPA plugins. The second registration replaces the first.

This proposed change makes it so that mltladspa loads the plugins and mltjackrack does not.

Is that broken? I don't understand why mltjackrack would need to load the ladspa plugins if ladspa is already doing it.

My goal is to avoid unintentional double registration as is happening in openfx. So I like the error message as a way to alert a developer if they make a mistake. But if there are good reasons to duplicate registrations, then maybe this change is not helpful.

Another approach I had considered: keep the error message,  but allow a module to query to see if a service exists before registering it. I would be happy to add that for ladspa if duplicate registration is useful for that module.

ddennedy:
> Is that broken? I don't understand why mltjackrack would need to load the ladspa plugins if ladspa is already doing it.

I am fairly sure you are correct. You can build `mltjackrack` with JACK alone or with LADSPA, but if you build it with LADSPA you also get `mltladspa`.

src/modules/jackrack/factory.c
```diff
@@ -703,6 +703,7 @@ static mlt_properties vst2_metadata(mlt_service_type type, const char *id, char
 
 JACKRACK_MODULE_EXPORT MLT_REPOSITORY
 {
+#if defined(mltladspa_EXPORTS)
 #ifdef GPL
```
ddennedy: You can combine these `#if`s; easier to read.

I am trying to see why the msys2 build failed, and I guess it does not define `mltladspa_EXPORTS` for some reason. Another option is to add `target_compile_definitions(WITH_LADSPA)` in `CMakeLists.txt` for consistency.

src/modules/jackrack/factory.c
```diff
@@ -783,17 +784,26 @@ JACKRACK_MODULE_EXPORT MLT_REPOSITORY
...
-    MLT_REGISTER_METADATA(mlt_service_filter_type, "ladspa", metadata, "filter_ladspa.yml");
+#endif
 #endif
 #ifdef WITH_JACK
+#if defined(mltjackrack_EXPORTS)
+    // JACK consumer belongs to mltjackrack to avoid duplicate registration.
     MLT_REGISTER(mlt_service_consumer_type, "jack", consumer_jack_init);
     MLT_REGISTER_METADATA(mlt_service_consumer_type, "jack", metadata, "consumer_jack.yml");
 #endif
+#endif
```
ddennedy:
```suggestion
#ifdef WITH_JACK
#ifdef GPL
    // JACK filter services belong to mltjackrack to avoid duplicate registration.
MLT_REGISTER(mlt_service_filter_type, "jack", filter_jackrack_init);
MLT_REGISTER_METADATA(mlt_service_filter_type, "jack", metadata, "filter_jack.yml");
MLT_REGISTER(mlt_service_filter_type, "jackrack", filter_jackrack_init);
MLT_REGISTER_METADATA(mlt_service_filter_type, "jackrack", metadata, "filter_jackrack.yml");
#endif
    // JACK consumer belongs to mltjackrack to avoid duplicate registration.
MLT_REGISTER(mlt_service_consumer_type, "jack", consumer_jack_init);
MLT_REGISTER_METADATA(mlt_service_consumer_type, "jack", metadata, "consumer_jack.yml");
#endif
```

src/modules/jackrack/factory.c
bmatherly: I fixed this with some guards.

src/framework/mlt_repository.c
bmatherly: Done

src/modules/jackrack/factory.c
bmatherly: Good suggestion. I took it a step farther by consolidating all the conditionals for each set of registrations. I think it makes a lot more sense when reading it. Let me know what you think.

src/modules/jackrack/factory.c
```diff
@@ -105,10 +105,11 @@ plugin_mgr_t *g_jackrack_plugin_mgr = NULL;
 lv2_mgr_t *g_lv2_plugin_mgr = NULL;
 #endif
 
-#ifdef WITH_VST2
+#if defined(GPL) && defined(WITH_VST2)
 vst2_mgr_t *g_vst2_plugin_mgr = NULL;
 #endif
 
+#if defined(mltladspa_EXPORTS) || (defined(mltjackrack_EXPORTS) && defined(WITH_JACK))
```
ddennedy: Why do you think you need `defined(mltjackrack_EXPORTS)`? That is redundant with `WITH_JACK`

bmatherly: I thought that too, but the AI talked me into it due to the technicality of how CMakeLists is set up:


> No. They are not synonymous in CMakeLists.txt.
> 
> mltjackrack_EXPORTS comes from generate_export_header for the mltjackrack target, so it is part of building that module. WITH_JACK is only added inside the JACK::JACK conditional, so it means JACK support is present. That means WITH_JACK can be absent even while mltjackrack_EXPORTS is defined.
> 
> So the guard on metadata() should not be treated as redundant with WITH_JACK. If you want the helper available for both mltladspa and mltjackrack builds, the guard should follow the module exports that actually use it, not only the JACK feature flag.
> 
> The concise way to phrase it is:
> 
> "mltjackrack_EXPORTS means we are building the module; WITH_JACK means JACK support is available. Those are related but not equivalent."


That is a technicality. In practice, we would always want WITH_JACK to be defined when building mltjackrack.

Could that be why the windows build was failing? Because it is building mltjackrack but Jack is absent?

ddennedy: The `CMakeLists.txt` was not properly maintained to make the library targets (`add_library`) optional depending on the included dependencies. `mltjackrack` should only be built when `JACK::JACK`, which adds the compile definition `WITH_JACK`. Likewise, `mltladspa` should only be be added when `(GPL AND TARGET PkgConfig::xml AND TARGET PkgConfig::glib AND ladspa_h_FOUND)`. Instead, potentially no-op modules are built such as `mltjackrack` on Shotcut/Windows, which does not support JACK.

Can you clean that up as part of this? I would prefer to limit the usage of the EXPORT definitions to their respective header includes.

bmatherly: Yes. I am willing to clean that up. But, before I do, do we want to consider breaking this code into two directories?

Here is a quick proposal: make two directories: jackrack and ladspa. Some files in ladspa are "shared" and would be compiled by both modules (there is no precedence for this - we could also copy/paste it)

Legend: * = currently compiled into both module targets (mltjackrack and mltladspa)
filename -> Proposed directory (after split)

factory.c -> split into jackrack/factory.c and ladspa/factory.c

*filter_ladspa.c -> ladspa/
*producer_ladspa.c -> ladspa/
*jack_rack.c -> ladspa/
*jack_rack.h -> ladspa/
*lock_free_fifo.c -> ladspa/
*lock_free_fifo.h -> ladspa/
*plugin.c -> ladspa/
*plugin.h -> ladspa/
*plugin_desc.c -> ladspa/
*plugin_desc.h -> ladspa/
*plugin_mgr.c -> ladspa/
*plugin_mgr.h -> ladspa/
*plugin_settings.c -> ladspa/
*plugin_settings.h -> ladspa/
*process.c -> ladspa/
*process.h -> ladspa/

filter_jackrack.c -> jackrack/
consumer_jack.c -> jackrack/
filter_jack.yml -> jackrack/
filter_jackrack.yml -> jackrack/
consumer_jack.yml -> jackrack/

filter_lv2.c -> ladspa/
producer_lv2.c -> ladspa/
lv2_context.c -> ladspa/
lv2_context.h -> ladspa/
lv2_plugin.c -> ladspa/
lv2_plugin.h -> ladspa/
lv2_plugin_settings.c -> ladspa/
lv2_plugin_settings.h -> ladspa/
lv2_process.c -> ladspa/
lv2_process.h -> ladspa/
lv2_urid_helper.h -> ladspa/
lv2blacklist.txt -> ladspa/

filter_vst2.c -> ladspa/
producer_vst2.c -> ladspa/
vst2_context.c -> ladspa/
vst2_context.h -> ladspa/
vst2_plugin.c -> ladspa/
vst2_plugin.h -> ladspa/
vst2_plugin_settings.c -> ladspa/
vst2_plugin_settings.h -> ladspa/
vst2_process.c -> ladspa/
vst2_process.h -> ladspa/
vestige.h -> ladspa/

filter_ladspa.yml -> ladspa/
producer_ladspa.yml -> ladspa/
filter_lv2.yml -> ladspa/
producer_lv2.yml -> ladspa/
filter_vst2.yml -> ladspa/
producer_vst2.yml -> ladspa/
blacklist.txt -> ladspa/

Let me know your thoughts. If you are not interested, I'll just to a simple cleanup in CMakeLists.txt.

ddennedy: No, I do not want to do that.

bmatherly: CMakeLists.txt is cleaned up.

src/modules/jackrack/CMakeLists.txt
```diff
@@ -1,62 +1,58 @@
...
+  target_compile_options(mltjackrack PRIVATE ${MLT_COMPILE_OPTIONS})
+
+  target_link_libraries(mltjackrack PRIVATE mlt Threads::Threads JACK::JACK)
   if(MSVC)
-      target_link_libraries(mltjackrack PRIVATE msvccompat)
+      target_link_libraries(mltjackrack PRIVATE PThreads4W::PThreads4W msvccompat)
   endif()
+  target_include_directories(mltjackrack PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  target_sources(mltjackrack PRIVATE consumer_jack.c)
   target_compile_definitions(mltjackrack PRIVATE WITH_JACK)
```
ddennedy: Now, `WITH_JACK` is dependent upon the dependency and only defined when the optional module is built. But `factory.c` is now using the defines `mltjackrack_EXPORTS` instead of `WITH_JACK` that I asked for. I said I only wanted to use the EXPORTS defines for controlling the export header includes. I also, asked for the  `mltladspa` module to be optional and define a new `WITH_LADSPA` that can be used by itself in `factory.c` instead of `defined(mltladspa_EXPORTS) && defined(GPL)`. The new define depends on `(GPL AND TARGET PkgConfig::xml AND TARGET PkgConfig::glib AND ladspa_h_FOUND)` So, there would be no need to repeat the depenency on `GPL` in  `factory.c`. Instead, it would be clean, readable `#ifdef WITH_JACK` and `#ifdef WITH_LADSPA` blocks.

src/modules/jackrack/factory.c
```diff
@@ -43,7 +43,7 @@ extern mlt_consumer consumer_jack_init(mlt_profile profile,
 #include "plugin_mgr.h"
 #include <ladspa.h>
 
-#ifdef WITH_JACK
+#if defined(mltjackrack_EXPORTS)
```
ddennedy: Don't change this. In `CMakeLists.txt` `WITH_JACK` now is only defined when the `mltjackrack` module is built. The EXPORTS defines are for controlling export header includes.

src/modules/jackrack/factory.c
```diff
@@ -84,7 +84,7 @@ extern mlt_producer producer_lv2_init(mlt_profile profile,
                                       char *arg);
 #endif
 
-#ifdef WITH_VST2
+#if defined(GPL) && defined(WITH_VST2)
```
ddennedy: No need to change this. WITH_VST2 depends on GPL.

src/modules/jackrack/factory.c
```diff
@@ -703,96 +703,106 @@ static mlt_properties vst2_metadata(mlt_service_type type, const char *id, char
...
-
 #endif
 
-#ifdef WITH_JACK
+    // Registrations owned by mltjackrack.
+#if defined(mltjackrack_EXPORTS) && defined(GPL)
     MLT_REGISTER(mlt_service_filter_type, "jack", filter_jackrack_init);
     MLT_REGISTER_METADATA(mlt_service_filter_type, "jack", metadata, "filter_jack.yml");
     MLT_REGISTER(mlt_service_filter_type, "jackrack", filter_jackrack_init);
     MLT_REGISTER_METADATA(mlt_service_filter_type, "jackrack", metadata, "filter_jackrack.yml");
```
ddennedy: These depends on WITH_JACK and a new WITH_LADSPA. The latter is dependent upon GPL and does not need repeating here.

src/modules/jackrack/factory.c
```diff
@@ -703,96 +703,106 @@ static mlt_properties vst2_metadata(mlt_service_type type, const char *id, char
...
-#endif
     MLT_REGISTER(mlt_service_filter_type, "ladspa", filter_ladspa_init);
     MLT_REGISTER_METADATA(mlt_service_filter_type, "ladspa", metadata, "filter_ladspa.yml");
```
ddennedy: These depend upon LADSPA but by removing the `#endif` you made them also dependent upon JACK. Now, `mltladspa` would only be geting lv2 and vst2 plugins.

ddennedy (changes requested):

src/modules/jackrack/CMakeLists.txt
bmatherly: I was using the *_EXPORTS definition as a way to determine which library was being compiled. I see what you mean about WITH_LADSPA. I will make a commit to do that instead.

bmatherly: Can you explain more what you mean by:

> I also, asked for the mltladspa module to be optional 

Do you want the user to be able to enable/disable mltladspa as a cmake variable when compiling? Or do you just mean that if ladspa_h_FOUND=False, then do not build mltladspa?

We currently have a CMAKE variable to disable the entire jackrack module (directory) - which disables both mltjackrack and mltladspa. Maybe you are suggesting two CMAKE variables - one to disable jackrack and one to disable mltladspa?

ddennedy: I previously mentioned that it is possible to build a mltjackrack with only a JACK consumer only mltjackrack, meaning no mltladspa. 

I was not asking for a new CMake variable, but we can  There is already a line `if(GPL AND TARGET PkgConfig::xml AND TARGET PkgConfig::glib AND ladspa_h_FOUND)` that defines how to build a `mltladspa`. That can add the mltladspa library and define WITH_LADSPA. But now, I see it might be possible to build mltladspa without LADSPA but including LV2 and VST2. However, I am not sure that can actually work. So, there can be a new variable `USE_LADSPA` and to add mltladspa:
`if(GPL AND (USE_LADSPA OR USE_LV2 OR USE_VST2))`

I will do it and push a commit here.

CMakeLists.txt
```diff
@@ -267,23 +268,24 @@ endif()
...
-  check_include_file(ladspa.h ladspa_h_FOUND)
-  if(NOT ladspa_h_FOUND)
-    set(USE_LV2 OFF)
-    set(USE_VST2 OFF)
+  if(USE_LADSPA)
+    if(GPL)
+      pkg_check_modules(glib REQUIRED IMPORTED_TARGET glib-2.0)
+      pkg_check_modules(xml REQUIRED IMPORTED_TARGET libxml-2.0)
+      if(USE_LV2)
+        pkg_check_modules(lilv REQUIRED IMPORTED_TARGET lilv-0)
```
bmatherly, line 276: This changed a previous behavior. Before, if LV2 is not  found, it is automatically disabled. Now the build fails. Here is a suggestion to restore the previous behavior:

```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,7 +273,11 @@ if(MOD_JACKRACK)
       pkg_check_modules(glib REQUIRED IMPORTED_TARGET glib-2.0)
       pkg_check_modules(xml REQUIRED IMPORTED_TARGET libxml-2.0)
       if(USE_LV2)
-        pkg_check_modules(lilv REQUIRED IMPORTED_TARGET lilv-0)
+        pkg_check_modules(lilv IMPORTED_TARGET lilv-0)
+        if(NOT lilv_FOUND)
+          message(WARNING "USE_LV2=ON but lilv-0 was not found via pkg-config. Disabling LV2 support. Install lilv development files or set PKG_CONFIG_PATH to include lilv-0.pc.")
+          set(USE_LV2 OFF)
+        endif()
       endif()
       check_include_file(ladspa.h ladspa_h_FOUND)
       if(NOT ladspa_h_FOUND)

```

ddennedy: I know; it is intentional. Build dependencies are hard and intentional now.

PR #1273, opened 2026-07-27
Fix incorrect PAR in qtblend transition

j-b-m:
With the addition of `normalize_mlt_source_size` in https://github.com/mltframework/mlt/commit/86b18776f27f4ddbb5352a18bdaee8deba131134, the frames with a PAR != consumer PAR had their size adjusted twice, resulting in incorrectly sized images. A 1.3333 PAR 1440x1080 video (which should appear full screen in a 1920x1080 project) was scaled down and appeared with a large black border.

Kdenlive bug #522730


ddennedy:
@guiodic please review

guiodic:
> @guiodic please review

That looks like an excellent solution to me

ddennedy:
This change introduced a major regression reported here:
https://forum.shotcut.org/t/some-files-are-horribly-blurry-in-my-project/51785

I am going to revert it, but I wonder if I should revert 86b1877 as well.

I reproduce it with a command line like:
`melt -profile atsc_1080p_25 color:white -track /path/to.png -attach affine transition.rect=25%,25%,50%,50% -transition qtblend`
The PNG I am using is 512x512, less than 1920x1080.

guiodic:
> I am going to revert it, but I wonder if I should revert [86b1877](https://github.com/mltframework/mlt/commit/86b18776f27f4ddbb5352a18bdaee8deba131134) as well.

I presume the bug can be fixed without reverting [86b1877] that commit solves a worse bug. I will try tomorrow.

> I reproduce it with a command line like: `melt -profile atsc_1080p_25 color:white -track /path/to.png -attach affine transition.rect=25%,25%,50%,50% -transition qtblend` The PNG I am using is 512x512, less than 1920x1080.

Good, I will test a couple of ideas.

guiodic:
This https://github.com/mltframework/mlt/pull/1277 should fix the PAR problem without introducing the blur one. I tested with your melt command and the results seems as it was before this commit. Also I tested the project in KDE bug and the PAR problem is fixed.

PR #1275, opened 2026-07-30
Fix blur not drawing in shape for motion tracker

doruphin:
This was originally brought up in [Kdenlive](https://invent.kde.org/multimedia/kdenlive/-/work_items/1059#note_1558411).

Rather than always drawing the blur as a rectangle, it now holds it in a buffer and then applies it in the designated shape. Right now, this only really affects the eclipse shape since it's the only odd shape out.

Example:
Screenshot 2026-07-29 183017

### DIFF
--- a/src/modules/opencv/filter_opencv_tracker.cpp
+++ b/src/modules/opencv/filter_opencv_tracker.cpp
@@ -468,18 +468,19 @@ static int filter_get_image(mlt_frame frame,
     }
 
     if (blur > 0 && data->boundingBox.width > 1 && data->boundingBox.height > 1) {
+        cv::Mat roi = cvFrame(data->boundingBox);
+        cv::Mat blurredRoi;
+        roi.copyTo(blurredRoi);
+        bool do_blur = true;
+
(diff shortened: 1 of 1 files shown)


src/modules/opencv/filter_opencv_tracker.cpp
```diff
@@ -546,7 +566,7 @@ static int filter_get_image(mlt_frame frame,
                             bounding,
                             cv::Scalar(shape_color.r, shape_color.g, shape_color.b),
                             shape_width,
-                            1);
+                            cv::LINE_4);
```
doruphin, line 569: From the cv library, the 1 before defaulted to a 4, so this just uses the nicer enum

From `opencv-4.5.5/modules/imgproc/src/drawing.cpp:263`
```
if( connectivity == 0 )
        connectivity = 8;
    else if( connectivity == 1 )
        connectivity = 4;
```

luzpaz:
Thank you @doruphin for working on this :pray:

ddennedy:
Do you want to close this in favor of #1279, which also includes these commits?

PR #1278, opened 2026-08-03
Add track audio-level metadata propagation to output frames.

bmatherly:
This updates MLT so per-track audio level metadata survives normal audio processing and reaches the final frame consumed by the aplication. The audiolevel filter now supports a prefix and appends the channel index, and mix/tractor copy the relevant metadata after audio is realized.

The change was made because track-level meter values were being generated but not visible on the output. By propagating metadata in the post-audio path, we makw track meter values appear on the output frame.

### DIFF
--- a/src/framework/mlt_tractor.c
+++ b/src/framework/mlt_tractor.c
@@ -440,6 +440,10 @@ static int producer_get_audio(mlt_frame self,
                        "producer_consumer_fps",
                        mlt_properties_get(properties, "producer_consumer_fps"));
     mlt_frame_get_audio(frame, buffer, format, frequency, channels, samples);
+
+    // Propagate audio-level metadata onto the tractor output frame.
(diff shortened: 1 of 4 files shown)


src/framework/mlt_tractor.c
```diff
@@ -440,6 +440,10 @@ static int producer_get_audio(mlt_frame self,
                        "producer_consumer_fps",
                        mlt_properties_get(properties, "producer_consumer_fps"));
     mlt_frame_get_audio(frame, buffer, format, frequency, channels, samples);
+
+    // Propagate audio-level metadata onto the tractor output frame.
+    mlt_properties_copy(properties, frame_properties, "meta.");
```
bmatherly, line 445: This is a wider net than I wanted,  but I didn't want something application specific. Shotcut sets the audio level metadata prefix to meta.track.<n>.audiolevel.<n>.

PR #1291, opened 2026-09-06
feat(effects): Add keyword in dynamictext

camillem:
Hi,

Having a frame number starting at 1 is useful in some animation workflows to have a more natural numbering of frames.
I basically just duplicated the code for the keyword "frame" and called it "frameo" ("o" for "offset").

Cheers, 
Camille

### DIFF
--- a/src/modules/plus/filter_dynamictext.c
+++ b/src/modules/plus/filter_dynamictext.c
@@ -88,11 +88,11 @@ static void get_timecode_str(mlt_filter filter,
         strncat(text, s, MAX_TEXT_LEN - strlen(text) - 1);
 }
 
-static void get_frame_str(mlt_filter filter, mlt_frame frame, char *text)
+static void get_frame_str(mlt_filter filter, mlt_frame frame, char *text, int offset)
 {
-    int pos = mlt_frame_get_position(frame);
+    int pos = mlt_frame_get_position(frame) + offset;
     char s[12];
-    snprintf(s, sizeof(s) - 1, "%d", pos);
+    snprintf(s, sizeof(s), "%d", pos);
     strncat(text, s, MAX_TEXT_LEN - strlen(text) - 1);
 }
 
@@ -221,7 +221,9 @@ static void substitute_keywords(mlt_filter filter, char *result, char *value, ml
         } else if (!strcmp(keyword, "smpte_ndf")) {
             get_timecode_str(filter, frame, result, mlt_time_smpte_ndf);
         } else if (!strcmp(keyword, "frame")) {
-            get_frame_str(filter, frame, result);
+            get_frame_str(filter, frame, result, 0);
+        } else if (!strcmp(keyword, "frame+1")) {
+            get_frame_str(filter, frame, result, 1);
         } else if (!strncmp(keyword, "filedate", 8)) {
             get_filedate_str(keyword, filter, frame, result);
         } else if (!strncmp(keyword, "localfiledate", 13)) {
--- a/src/modules/plus/filter_dynamictext.yml
+++ b/src/modules/plus/filter_dynamictext.yml
@@ -2,7 +2,7 @@ schema_version: 7.2
 type: filter
 identifier: dynamictext
 title: Dynamic text
-version: 5
+version: 6
 copyright: Meltytech, LLC
 license: LGPLv2.1
 language: en
@@ -27,6 +27,7 @@ parameters:
         * #smpte_ndf#     - SMPTE non-drop-frame timecode of the frame
         * #timecode#      - same as #smpte_df#
         * #frame#         - frame number of the frame
+        * #frame+1#       - frame number of the frame, starting at 1, not 0
         * #filedate#      - modification date of the file (GMT)
         * #localfiledate# - modification date of the file (local)
         * #localtime#     - current system date and time


src/modules/plus/filter_dynamictext.yml
```diff
@@ -20,13 +20,14 @@ parameters:
...
     description: |
-      The text to overlay. May include keywords enclosed in "#". 
+      The text to overlay. May include keywords enclosed in "#".
       Keywords include:
         * #createdate#    - Best guess of file creation date
         * #smpte_df#      - SMPTE drop-frame timecode of the frame
         * #smpte_ndf#     - SMPTE non-drop-frame timecode of the frame
         * #timecode#      - same as #smpte_df#
         * #frame#         - frame number of the frame
+        * #frameo#        - frame number of the frame, starting at 1,not 0
```
bmatherly: In the count producer, we have "seconds" and "seconds+1". 
https://www.mltframework.org/plugins/ProducerCount/

So, for this, I think we should have "frame" and "frame+1"

src/modules/plus/filter_dynamictext.c
```diff
@@ -96,6 +96,14 @@ static void get_frame_str(mlt_filter filter, mlt_frame frame, char *text)
     strncat(text, s, MAX_TEXT_LEN - strlen(text) - 1);
 }
 
+static void get_frameo_str(mlt_filter filter, mlt_frame frame, char *text)
```
bmatherly: Instead of adding a whole new function, please add an offset parameter to get_Frame_str().

src/modules/plus/filter_dynamictext.yml
```diff
@@ -64,7 +65,7 @@ parameters:
     title: Font size
     type: integer
     description: >
-      The size in pixels of the font. 
+      The size in pixels of the font.
```
bmatherly: Please remove all these whitespace changes.

src/modules/plus/filter_dynamictext.yml
bmatherly: The filter  version should be incremented for this change.

bmatherly (changes requested):

shadowfax65:
How much more difficult would it be to create a variable offset that the user can define? Instead of a new keyword for a fixed offset of one frame it would allow for more flexibility if there was a parameter `offset` that could be set by the user with an arbitrary value.

bmatherly:
> How much more difficult would it be to create a variable offset that the user can define? 

It's not really about difficulty. Its more about intuition and purpose. The dynamictext filter isn't exclusively about frame position. Would the offset parameter apply to "createdate" or "meta.media.0.codec.frame_rate"?

If we want to make advanced frame/position counting features, I would rather put that in the count producer which is intended for that kind of thing.

But i don't think a simple "+1" keyword confuses anything and it works with existing user interfaces (Shotcut, etc).

RonOnGithub:
> It's not really about difficulty. Its more about intuition and purpose.

The original request came from a user doing animation - so the core of what they really want is a genuine frame count to know which animation cel is being shown rather than a timecode.

Which may also be interesting if we ever directly support variable frame rate video, where each frame can have its own variable duration?

src/modules/plus/filter_dynamictext.yml
camillem: Done, thanks.

src/modules/plus/filter_dynamictext.yml
camillem: Done, thanks.

src/modules/plus/filter_dynamictext.c
camillem: I gave it a try: hope it was the right way.

src/modules/plus/filter_dynamictext.yml
camillem: Done, thanks.

bmatherly:
> The original request came from a user doing animation - so the core of what they really want is a genuine frame count to know which animation cel is being shown rather than a timecode.

I'm probably missing some context. I don't know your animation workflow or why you are burning a frame number on your image. If there is a more complex workflow that this filter wasn't intended for, maybe we need a new filter.

frdbr:
> > The original request came from a user doing animation - so the core of what they really want is a genuine frame count to know which animation cel is being shown rather than a timecode.
> 
> I'm probably missing some context. I don't know your animation workflow or why you are burning a frame number on your image. If there is a more complex workflow that this filter wasn't intended for, maybe we need a new filter.

This is very handy in the pre-production phase. Usually animators use an animatic to help them work on their scenes, which is like an animated storyboard. It also helps productions with organizing audio stuff like speech, music, fx...

camillem:
The original request is very simple, and I think this PR solves the problem.

>  Is there a way to make the on-screen frame counter start at 1 instead of 0?
> Timer has an offset, but lacks a pure absolute "Frames" format in the dropdown.
> Are we missing a hidden setting? If it's not possible right now, could this be considered for a future update (like adding an offset to Dynamic Text or a "Frames" format to the Timer)? It would be amazing for animation workflows. Thanks

> Just to give some context on why starting at 1 is so important: In animation pipelines, we need to know the exact number of frames per shot. This allows us to measure the scope of the work and accurately track the workload for the animators during the production phase. Starting the count at frame 1 makes this tracking intuitive and aligns with standard animation workflows.

RonOnGithub:
> In animation pipelines, we need to know the exact number of frames per shot.

I think this is the key point of the original request in terms of what is needed from MLT for this use.   They need to be able to perfectly track the actual frame/cel number being displayed, not the timecode divided by some arbitrary number of frames per second.

bmatherly (approved):

